PR #8, opened 2017-03-05
Update docs and add new contrib instructions

dcwalk:


dcwalk:
hey @danielballan -- your thoughts?

PR #20, opened 2017-03-06
Cleanup with docs removal

dcwalk:
I moved the proposals into the web-monitoring wiki (shudder): https://github.com/edgi-govdata-archiving/web-monitoring/wiki


dcwalk:
I'd like to move the following to the main repo to reduce duplication:

- `archives/`
- `pagefreezer-api.md`
- `archive-layout.md`

We could link to them from the README as needed?

(I know you are using the archives for your `interactive.py` example @danielballan would it be difficult to change the path or `curl` the example data in?

dcwalk:
hey @danielballan -- your thoughts?

danielballan:
Sorry being AWOL, @dcwalk. I agree the archive/ stuff doesn't need to live here. Soon we'll want a much larger test set that definitely shouldn't live in the git repo.

PR #22, opened 2017-03-09
Update database schema to new standard

danielballan:
The database schema here was only roughly aligned with the one in the Rails app. Today, @Mr0grog and I came up with a database schema that works for both of us. See web_monitoring/db.py. I also did some cleanup and updated the notebook to work with the new schema.


danielballan:
Self-merging for now.

PR #23, opened 2017-03-12
rm out of date docs

dcwalk:
Per #20, further removal of out of date docs, these are transferred to my new favourite wiki: 
https://github.com/edgi-govdata-archiving/web-monitoring/wiki


dcwalk:
My understanding is we will have a new custom api within a week, the other doc we might want to move to the web-monitoring main repo?

danielballan:
Which other doc?

dcwalk:
the description of a pagefreezer snapshot layout. I don't think it is essential but maybe in that sampel dataset we have an excellent doc :)

PR #45, opened 2017-05-27
ENH: Submit IA and PF versions to web-monitoring-db

danielballan:
Example, tested against a local deployment of the web-monitoring-db app:

```python
>>> import web_monitoring.internetarchive as wi
>>> versions = list(wi.list_versions('nasa.gov'))
>>> formatted_versions = [wi.format_version(url='https://nasa.gov', dt=v[0], uri=v[1], version_hash='placeholder', title='placeholder', site='placeholder', agency='placeholder') for v in versions]
>>> formatted_versions[0]
{'capture_time': '1996-12-31T23:58:47',
 'page_title': 'placeholder',
 'page_url': 'https://nasa.gov',
 'site_agency': 'placeholder',
 'site_name': 'placeholder',
 'source_metadata': {},
 'source_type': 'internet_archive',
 'uri': 'http://web.archive.org/web/19961231235847/http://www.nasa.gov:80/',
 'version_hash': 'placeholder'}
# Looks good -- ready to submit to the web-monitoring-db app
>>> res = wi.import_into_db(formatted_versions[:2])
>>> res.content
b'{"data":{"id":4,"user_id":1,"status":"pending","file":"619b3f82-b61a-44ce-bb91-eb6750134e40","processing_errors":[],"created_at":"2017-05-27T21:15:18.046Z","updated_at":"2017-05-27T21:15:18.046Z"}}'
```


Mr0grog:
Is it possible to amend this to provide the `page_title` field, or should we add the ability for -db to download the raw content and extract a title from it?

(That said, this only matters for pages that -db does not already have a record for.)

danielballan:
Doesn't it already provide a `page_title` field? I think I'm missing your meaning.

Mr0grog:
I guess I was reading this in the example:

```py
{'capture_time': '1996-12-31T23:58:47',
 'page_title': 'placeholder',
```

…and presuming you aren’t actually sending anything meaningful for `page_title`—but maybe I was understanding `placeholder` to be what you are actually sending, when you just meant “something meaningful would normally be here?”

danielballan:
Yeah, that's what I was going for. But, I agree that extracting the title makes sense. We already need to download the version HTML in order to get `version_hash`, so we might as well automate extracting the `title` as well.

danielballan:
P.S. If you're still working on this for awhile, I have some semi-sensitive questions on Slack.

danielballan:
Added CLI utility for ingesting IA Versions that incorporates auto-extracting the `<title>` and taking in a site and agency.

web_monitoring/cli.py
```diff
@@ -0,0 +1,55 @@
...
+def import_ia(url, agency, site):
+    print('obtaining versions list from Internet Archive...')
+    versions = ia.list_versions(url)
+    # Collect all the results and POST them as a unit (rather than streaming).
+    formatted_versions = []
+    for dt, uri in tqdm(versions, desc='formatting versions'):
+        res = requests.get(uri)
+        assert res.ok
+        version_hash = hashlib.sha256(res.content)
+        content_as_file = io.StringIO(res.content.decode(errors='ignore'))
```
danielballan: @b5 Notice `errors='ignore'` here. I have seen some string decoding errors such as:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x85 in position 15924: invalid start byte
```

Do you know if there is a better string encoding to choose here?

b5: pardon my not taking time to understand the purpose of the `content_as_file` var, but you'd probably want to avoid any sort of coercion during string conversion. the `decode` func looks suspicious to me, as it's clearly trying to intepret the underlying response bytes, which aren't necessarily UTF-8 (character encoding is supposed to be specified setting the Content-Type header's charset property). Any sort of coercion here runs the risk of beaking the hash on line 27. If I'm understanding the purpose of `content_as_file` correctly, you'd just want to do a straight string representation of the underlying byte array.

danielballan: Don't worry, the `version_hash` is performed on raw undecoded bytes. This `content_as_file` is just a temporary var from which I am parsing the 'title' tag to auto-extract the Page title (to be used by humans, not machines). I have decode from bytes to string before I can parse XML. I think ignoring errors is fine here because (1) fidelity is not critical here and (2) errors seem rare.

b5: of course, b/c you're super smart. Now that I understand it, yep I'd say you're totally fine to ignore errors.

If you have the bandwidth, I'd do a check on the `Content-Type` header for the `encoding` prop, switch on that for decoding, with the default decoding being `ASCII` / `ISO-8859-1`. Doing this groundwork to determine encoding is well worth it IMHO, as _many_ places now serve `UTF-8`, but `ISO-8859-1` is still the sensible default given the history of the web. This approach would keep strange unrecognized utf-8 characters out of your extracted titles, and set you up for future parsing stuuuuuuff

danielballan:
Example of successfully importing from Internet Archive into [a local deployment of] the web-monitoring-db app:

```sh
$ wm import ia envirodatagov.org --site foo --agency bar
obtaining versions list from Internet Archive...
formatting: 34versions [00:04,  7.31versions/s]
posting to db....
{'data': {'id': 10, 'user_id': 1, 'status': 'pending', 'file': '83d53fe1-a28e-4ad7-a7bb-8b72ff064513', 'processing_errors': [], 'created_at': '2017-05-30T20:02:42.689Z', 'updated_at': '2017-05-30T20:02:42.689Z'}}
```

I confirmed that the new Versions appear in the app's views.

danielballan:
And now an example of successfully importing from PageFreezer's Google Cloud deployment for EDGI into web-monitoring-db:

```bash
$ wm import pf J_004232364409_3994fe0dd66ddd779 1494957042 --site foo --agency bar
formatting: 35versions [00:20,  1.73versions/s]
posting to db....
{'data': {'id': 11, 'user_id': 1, 'status': 'pending', 'file': 'ea435a12-7b3a-42a6-a12c-18df508ff846', 'processing_errors': [], 'created_at': '2017-05-30T20:10:53.292Z', 'updated_at': '2017-05-30T20:10:53.292Z'}}
```

web_monitoring/pf_edgi.py
```diff
@@ -0,0 +1,172 @@
...
+    unload_archive(cabinet_id, archive_id)
+
+
+def unique_urls(cabinets):
+    return set([entry['url']
+                for cabinet in cabinets.values()
+                for entry in cabinet])
+
+
+class Context:
```
Mr0grog: Is this vestigial or maybe future-focused? Doesn’t seem like it’s getting used. (Likewise `unique_urls()`, whose only usage is from this class.)

web_monitoring/pf_edgi.py
```diff
@@ -0,0 +1,172 @@
...
+                    agency, site):
+    """
+    Obtain URI, timestamp, metadata, hash, and title and return a Version.
+    """
+    uri = file_command_uri(cabinet_id, archive_id, page_key, 'file')
+    dt = datetime.fromtimestamp(int(archive_id))
+    metadata = get_file_metadata(cabinet_id, archive_id, page_key)
+    content = get_file(cabinet_id, archive_id, page_key)
+    version_hash = hashlib.sha256(content).hexdigest()
+    title = utils.extract_title(content)
```
Mr0grog: Would it make sense or be worthwhile to check the content-type from the metadata before trying to extract the title? (To skip this step on non-HTML files)

web_monitoring/cli.py
```diff
@@ -0,0 +1,67 @@
...
+# CLI. They also print. To access this functionality programmatically, it is
+# better to use the underlying library code.
+
+
+def import_ia(url, agency, site):
+    print('obtaining versions list from Internet Archive...')
+    versions = ia.list_versions(url)
+    # Collect all the results and POST them as a unit (rather than streaming).
+    formatted_versions = []
+    for dt, uri in tqdm(versions, desc='formatting', unit='versions'):
```
Mr0grog: formatting: 34versions [00:02, 10.49versions/s]

The lack of space between the number and “versions” is really bugging me here :P

danielballan: Me too! I guess we could do `unit=' versions'`. I assume tqdm is being aggressive about conserving horizontal space.

web_monitoring/cli.py
```diff
@@ -0,0 +1,67 @@
...
+                                                site=site,
+                                                agency=agency)
+        formatted_versions.append(version)
+    print('posting to db....')
+    res = db.post_versions(formatted_versions)
+    assert res.ok
+    print(res.json())
+
+
+def import_pf_archive(cabinet_id, archive_id, *, agency, site):
```
danielballan: Fortunately, it doesn't soak up anything. If you want it to soak up extra position args, you have to put a variable beside it like `*args`. What is does is force all args after it to be called with the keywords: it makes `agency` and `site` "keyword-only arguments." For functions with many arguments and/or arguments with no obvious ordering, it's a great way to avoid bugs. It also means that you can re-order these arguments in the future without breaking the API.

web_monitoring/pf_edgi.py
danielballan: Future focused. Querying all the archives is expensive (~30 seconds) and I think we might want a special data structure to juggle it all. I'm picturing adding methods that let us filter archives by timestamp, in RAM. That said, I generally try to get as far as possible using built-in data structures and I'm not sure this is worth keeping.

setup.py
```diff
@@ -1,6 +1,27 @@
...
+import os
 import setuptools
+import sys
+
+
+python_version = sys.version_info
+if python_version[0] < 3:
+    raise RuntimeError("Python version is {}. Requires 3.6 or greater."
+                       "".format(sys.version_info))
+elif python_version[0] == 3 and python_version[1] < 6:
```
danielballan: So it is. I remembered using `StrictVersion` once but then couldn't remember why it was useful and ended up here. Comparing tuples seems like the way to go.

web_monitoring/db.py
```diff
@@ -0,0 +1,32 @@
...
+    versions : iterable
+        iterable of dicts from :func:`format_version`
+    """
+    # Existing documentation of import API is in this PR:
+    # https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/32
+    url = WEB_MONITORING_IMPORT_API.format(db_url=settings['db_url'])
+    return requests.post(url,
+                         auth=(settings['db_email'], settings['db_password']),
+                         headers={'Content-Type': 'application/x-json-stream'},
+                         data='\n'.join(map(json.dumps, versions)))
```
danielballan: I haven't, but chunking seems worthwhile, as does polling. I'll add that before we merge.

scripts/wm
```diff
@@ -0,0 +1,6 @@
+#!/usr/env/python
```
danielballan: Must be an OSX-ism. Fixed.

web_monitoring/pf_edgi.py
danielballan: What's the canonical way to do that? Extract something like `<meta charset="UTF-8">`?

danielballan:
> I’m assuming some of the extra stuff, like list_cabinets, is meant for future use in related commands or as these commands evolve, since this currently requires explicit knowledge of which cabinet and archive to import, even though new ones should always be being created.

Yes. I intentionally resisted the urge to impose a more "user-friendly" structure. Starting from a stateless, low-level API, the right way to factor more helpful tools should be more obvious later.

scripts/wm
Mr0grog: Not apologizing for my unix-but-not computer :P

web_monitoring/cli.py
Mr0grog: Oooooh, neat, I did not know you could force kwargs this way.

web_monitoring/pf_edgi.py
Mr0grog: Oh, I think I meant more like:

```py
content_type = metadata['file']['ContentType'] or '' # Not certain how reliable the presence of these keys are
title = content_type.startswith('text/html') and utils.extract_title(content) or ''
```

web_monitoring/pf_edgi.py
Mr0grog: I usually try to avoid including bits that aren’t yet used in PRs, but I don’t really think it’s a big deal. No major reason to revise if you’re likely to start building on this pretty quickly.

danielballan:
Should be good to go. Tested with `BATCH_SIZE=10` and `BATCH_SIZE=1000`. (Later I'll make this configurable -- it's easy enough to do -- but I'd like to punt making the CLI nicer to a separate PR.)

Import from IA:

```sh
$ wm import ia envirodatagov.org --site foo --agency bar
importing: 34 versions [00:04,  7.08 versions/s]
Done. Now polling server, monitoring for job completition...
Completed 34 versions succcessfully. There were 0 errors.
```

Import from PF:

```sh
$ wm import pf J_004232364409_3994fe0dd66ddd779 1494957042 --site foo --agency bar
importing: 35 versions [00:18,  1.94 versions/s]
Done. Now polling server, monitoring for job completition...
Completed 35 versions succcessfully. There were 0 errors.
```

I also saw some transient server-side errors which we should investigate later. The good side-effect of that is that the error-message-printing code path in this PR was tested! :- )

One last pass, @Mr0grog?

web_monitoring/db.py
danielballan: This is done.

web_monitoring/pf_edgi.py
```diff
@@ -0,0 +1,177 @@
...
+    # Sniff whether this is text and, if so, what the encoding is.
+    # PF provides its own 'ContentType' key, mapped to a string,
+    # not to be confused with 'Content-Type' in the Header, mapped to a list.
+    content_type = metadata['file']['ContentType']
+    is_text = content_type.startswith('text/html')
+    if is_text:
+        if 'charset=' in content_type:
+            _, encoding = content_type.split('charset=')
+        else:
+            enconding = 'utf-8'  # best effort
```
Mr0grog: Typo here I think means this fallback isn’t going to work.

danielballan: Sharp eyes! Fixed.

janakrajchadha:
@danielballan I noticed a small issue while using a few of these scripts. In `pf_edgi.py`, I think `utils` should be directly imported instead of `from web_monitoring` as both of them are in the same folder.

danielballan:
Did you encounter an error?

Its from that, for example, `from . import utils` would have been equivalent to than `from web_monitoring import utils` and somewhat better style.

danielballan:
[For those watching by email, I edited in place the confusing typo above.]

janakrajchadha:
Yes, I did encounter an error (Module not found). When I changed it to `from . import utils`, it returned an import error - "attempted relative import with no known parent package".

danielballan:
Hmm. Were you sitting in a directory inside the package when you tried that? Sometimes that can confuse things. Are you sure that the package is installed? What happens when you try:

```python
>>> import web_monitoring
>>> web_monitoring.__file__
```

janakrajchadha:
Okay, I figured it out. I was trying to import `pf_edgi` in an interpreter to use some of the functions. It threw an error then. It works fine if you run the tests. Sorry about this.

danielballan:
Cool, no worries. The subtleties of the import system still bite me from time to time as well. :-)

PR #47, opened 2017-05-28
ENH: API for PF EDGI deployment.

danielballan:
A first step toward importing PF content into web-monitoring-db.


danielballan:
I folded these commits into #45 because there is some common code between post Versions from IA and PF.

Issue #36, opened 2017-04-15
OCI Runtime error - gunicorn executable file not found in PATH


ChaiBapchya:
# Command
`docker run -t yay`

# Output
```
docker: Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "exec: \"gunicorn\": executable file not found in $PATH".
```

ChaiBapchya:
When I tried to look at the Dockerfile
Following two lines were commented
```
RUN apt-get install -y gunicorn
RUN apt-get install -y python-gevent
```

Looked like that was the issue. The executable file of gunicorn didn't exist because it wasn't installed in the first place.
However after docker build and subsequent docker run
Following error -
```
Traceback (most recent call last):
  File "/usr/bin/gunicorn", line 9, in <module>
    load_entry_point('gunicorn==19.0.0', 'console_scripts', 'gunicorn')()
  File "/usr/lib/python2.7/dist-packages/gunicorn/app/wsgiapp.py", line 74, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/usr/lib/python2.7/dist-packages/gunicorn/app/base.py", line 166, in run
    super(Application, self).run()
  File "/usr/lib/python2.7/dist-packages/gunicorn/app/base.py", line 71, in run
    Arbiter(self).run()
  File "/usr/lib/python2.7/dist-packages/gunicorn/arbiter.py", line 169, in run
    self.manage_workers()
  File "/usr/lib/python2.7/dist-packages/gunicorn/arbiter.py", line 477, in manage_workers
    self.spawn_workers()
  File "/usr/lib/python2.7/dist-packages/gunicorn/arbiter.py", line 537, in spawn_workers
    time.sleep(0.1 * random.random())
  File "/usr/lib/python2.7/dist-packages/gunicorn/arbiter.py", line 209, in handle_chld
    self.reap_workers()
  File "/usr/lib/python2.7/dist-packages/gunicorn/arbiter.py", line 459, in reap_workers
    raise HaltServer(reason, self.WORKER_BOOT_ERROR)
gunicorn.errors.HaltServer: <HaltServer 'Worker failed to boot.' 3>
```

danielballan:
@lh00000000, do you have the bandwidth to take a look at this?

lh00000000:
hi, 
sorry about that. It was a hastily pasted-from-my-snippets dockerfile. 
[...]

PR #50, opened 2017-05-29
gunicorn rises again

lh00000000:
changes so docker image will be able to find gunicorn


danielballan:
Thanks!

(For anyone following along, see #36 for discussion.)

PR #59, opened 2017-06-22
ENH: Minimal plugable diffing server.

danielballan:
**This description is now out of date -- see recap in my last comment.**

Short Version: You write a Python function `f(a, b)` where `a` and `b` are HTML. (To be precise, they are generic bytestrings of response body content.) You register your function with the server by adding a line to a config file. Boom! Your function is now a diffing service.

Long Version:

This PR provides a generic, plugable "diffing server" that makes it easy to deploy diffing utilities without re-implementing network code. Specifically, the server handles incoming requests, fetches Version content, and dispatches to a registry of differs. (In the future, content validation and caching can be handled as well.) All a differ needs to do is compare the bytes of the response bodies.

A differ can be expressed as a Python function that takes two positional arguments -- the bytestrings to be diffed. It might also accept additional options specific to the differ, given as keyword arguments that may be required or optional. Here's a differ that returns the difference in the lengths of the two bytestrings.

```python
def compare_length(a, b):
    return len(b) - len(a)
```

The actual processing code in the diffing fucntion may in Python directly, as in this example, or it may just be called from Python by executing a shell command in a subprocess or making a network request.

To plug this a differ into the server, add it to the config file.

This example ,`compare_length` happens to be defined in the module `web_monitoring.differs`. (It could just as well be defined is some other user module, `my_favorite_differs`.) The example config file, `diff_config.json.example`, specifies differ names, each mapped to a list with two elements: the module where the function can be imported from and the name of the function. The differ name may be the same as the function name, but it doesn't have to be.

```json
# diff_config.json.example
{"dummy": ["web_monitoring.differs", "dummy"],
"length": ["web_monitoring.differs", "compare_length"]}
```

Start the diffing server and provide the path to the config file as an argument.

```bash
$ wm-diffing-server diff_config.json.example
```

Example usage with curl:

```bash
$ curl "localhost:8888/length?a=https://google.com&b=https://yahoo.com"
{"diff": 721891}
```

Query parameters other than a and b are passed through to the function as keyword arguments. The `dummy` differ illustrates this:

```python
def dummy(a, b, **options):
    # dummy example illustrating how query parameters pass through
    return options
```

```bash
$ curl "localhost:8888/dummy?a=https://google.com&b=https://yahoo.com&c=cat&d=dog"
{"diff": {"c": "cat", "d": "dog"}}
```

TODO:

- [x] Documentation
- [ ] Basic caching of fetched Version content
- [x] Use optional `a_hash` and `b_hash` parameters to validate downloaded Version content.
- [x] Integrate more differs: PageFreezer and plain text to start

### DIFF
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -5,3 +5,7 @@ def dummy(a, b, **options):
 
 def compare_length(a, b):
     return len(b) - len(a)
+
+
+def identical_bytes(a, b):
+    return a == b
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -1,4 +1,5 @@
 from bs4 import BeautifulSoup
+import diff_match_patch
 import re
 import web_monitoring.pagefreezer
 import sys
@@ -57,3 +58,35 @@ def pagefreezer_direct(a_text, b_text):
     "Dispatch to PageFreezer, the slow way."
     # Send PF the full body.
     return web_monitoring.pagefreezer.compare(a_text, b_text)
+
+
+d = diff_match_patch.diff_match_patch()
+
+def _diff(a_text, b_text):
+    return d.diff_compute(a_text, b_text, False, DEADLINE)
+
+
+def html_text_diff(a_text, b_text):
+    """
+    Example
+    ------
+    >>> text_diff('<p>Deleted</p><p>Unchanged</p>',
+    ...           '<p>Added</p><p>Unchanged</p>')
+    [(0, '<p>'), (-1, 'Delet'), (1, 'Add'), (0, 'ed</p><p>Unchanged</p>')]
+    """
+    t1 = ' '.join(_get_visible_text(a_text))
+    t2 = ' '.join(_get_visible_text(b_text))
+    DEADLINE = 2  # seconds
+    return d.diff_compute(t1, t2, checklines=False, deadline=DEADLINE)
+
+
+def html_source_diff(a_text, b_text):
+    """
+    Example
+    ------
+    >>> text_diff('<p>Deleted</p><p>Unchanged</p>',
+    ...           '<p>Added</p><p>Unchanged</p>')
+    [(-1, '<p>Deleted</p>'), (1, '<p>Added</p>'), (0, '<p>Unchanged</p>')]
+    """
+    DEADLINE = 2  # seconds
+    return d.diff_compute(a_text, b_text, checklines=False, deadline=DEADLINE)
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -59,10 +59,12 @@ d = diff_match_patch.diff_match_patch()
 
 def html_text_diff(a_text, b_text):
     """
+    Diff the visible textual content of an HTML document.
+
     Example
     ------
-    >>> text_diff('<p>Deleted</p><p>Unchanged</p>',
-    ...           '<p>Added</p><p>Unchanged</p>')
+    >>> html_text_diff('<p>Deleted</p><p>Unchanged</p>',
+    ...                '<p>Added</p><p>Unchanged</p>')
     [(0, '<p>'), (-1, 'Delet'), (1, 'Add'), (0, 'ed</p><p>Unchanged</p>')]
     """
     t1 = ' '.join(_get_visible_text(a_text))
@@ -73,11 +75,13 @@ def html_text_diff(a_text, b_text):
 
 def html_source_diff(a_text, b_text):
     """
+    Diff the full source code of an HTML document.
+
     Example
     ------
-    >>> text_diff('<p>Deleted</p><p>Unchanged</p>',
-    ...           '<p>Added</p><p>Unchanged</p>')
-    [(-1, '<p>Deleted</p>'), (1, '<p>Added</p>'), (0, '<p>Unchanged</p>')]
+    >>> html_source_diff('<p>Deleted</p><p>Unchanged</p>',
+    ...                  '<p>Added</p><p>Unchanged</p>')
+    [(0, '<p>'), (-1, 'Delet'), (1, 'Add'), (0, 'ed</p><p>Unchanged</p>')]
     """
     DEADLINE = 2  # seconds
     return d.diff_compute(a_text, b_text, checklines=False, deadline=DEADLINE)
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -54,12 +54,6 @@ def pagefreezer(a_url, b_url):
     return web_monitoring.pagefreezer.compare(a_url, b_url)
 
 
-def pagefreezer_direct(a_text, b_text):
-    "Dispatch to PageFreezer, the slow way."
-    # Send PF the full body.
-    return web_monitoring.pagefreezer.compare(a_text, b_text)
-
-
 d = diff_match_patch.diff_match_patch()
 
 def _diff(a_text, b_text):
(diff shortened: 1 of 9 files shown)


web_monitoring/diffing_server.py
```diff
@@ -0,0 +1,92 @@
...
+        # Fetch server response for URLs a and b.
+        res_a, res_b = yield [client.fetch(a), client.fetch(b)]
+        # TODO Add caching of fetched URIs.
+
+        # TODO Validate the bytes against the hash, if provided.
+
+        # Pass the bytes and any remaining args to the diffing function.
+        res = yield executor.submit(func, res_a.body, res_b.body,
+                                    **query_params)
+        self.write(json.dumps({'diff': res}))
```
Mr0grog: Just wondering: was there particular thinking behind this, as opposed to just returning the raw result from the differ function?

web_monitoring/diffing_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+def cli():
+    doc = """Start a diffing server.
+
+Usage:
+wm-diffing-server <config_file> [--port <port>]
+
+Options:
+-h --help     Show this screen.
+--version     Show version.
```
Mr0grog: Probably should list `--port` here (I think you can set—and document—a default value this way, too).

web_monitoring/diffing_server.py
danielballan: Will we want the differ to report back other information about itself, like it's version? With that kind of thing in mind, I figure it was safer to scope the content inside a key.

Mr0grog: Oh, good point, I guess. Would we expect a differ that returns images or PDFs (e.g. https://github.com/edgi-govdata-archiving/web-monitoring/issues/36 or Versionista’s rendered images diff) to returned a base64-encoded string (or data URL, so it has MIME type info)? The pain of dealing with binary response data is the reason I was thinking raw responses.

(Of course I guess we have to handle it in `web-monitoring-db`, too… *[smacks forehead]*)

danielballan:
> Big picture, are you thinking people should fork/copy this to create a new differ project, or should they be submitting PRs back to this to add new diffing routines? If the latter, maybe the config shouldn’t be an example, but the actual config config file we want to deploy.

The latter. Hahah, I had the same thought. Will change.

> Maybe the first two args should be objects that represent the full HTTP response or that otherwise encapsulate a bunch of metadata along with a body field?

That would be a reasonable way to do it. But I think I'd prefer each function take in only the information that it needs. That makes it clear what it depends on, and it's also easier to test and experiment with. The determination of format probably has to be made higher up: send PDFs to the PDF differ and HTML to an HTML differ. A differ like `compare_length` doesn't need to know byte encoding, but a plain text differ would probably require that as one of its keyword arguments. We will probably need a middle layer that can make it a generic Change and dispatch to all the available differs that suit its Content-Type. Does that sound workable?

web_monitoring/diffing_server.py
danielballan: Oh, definitely.

web_monitoring/diffing_server.py
danielballan: Yeah, I was imagining base64-encoded strings. But as long as the UI knows what to expected back from a given differ we don't necessary need to stick to exactly one return value format.

web_monitoring/differs.py
```diff
@@ -5,3 +5,7 @@ def dummy(a, b, **options):
 
 def compare_length(a, b):
```
janakrajchadha: This may lead to false positives in the situation when two byte-strings have the same length but they are, in fact, different.

danielballan: Definitely. All this checks is the length. If the length is *different*, then we know that the byte-strings are different. This may or may not be useful; if nothing else it makes for a simple instructive example.

danielballan:
Had a chat with @jeffreyliu and @b5 and settled on trying the following scheme:

Differs are simple functions with varied signatures. Some will need byte encodings, some may need other info that is knowable from the header. As originally planned, some will be implemented in this package and some maybe maintained separately in external packages.

A new layer will be added to the server that takes in the header, determines which differs are applicable to this content, extract the header parameters that they need, and call them.

To support this, the JSON file that currently specifies the set of available of differs will have to be replaced by something more complex.

danielballan:
... I'll illustrate that previous comment with actual code shortly. :- )

But if that make sense early feedback in welcome!

danielballan:
OK, I think this is a good place to declare victory on this PR. To recap the features:

* There is a plain text 'differ', a dispatch to PageFreezer, and a couple other very simple differs.
* If incoming requests provide `a_hash` and/or `b_hash`, downloaded Versions are validated.
* There is no caching of Version content, but there is a clear way forward to add it.
* Incoming requests are targeted to specific differs; there is no way to make a blanket request like, "Diff this HTML will all applicable differs," or "Diff this PDF...."

To recap the implementation:

I'm putting forward a scheme where diffing functions specify their dependencies (e.g., raw bytes, decoded bytes, original request URLs) via their signatures [argument names]. I like that this makes the inputs very clear and makes the functions easy to test. To the maximum extent, it separates the diffing logic from any dependency on the server implementation and its object types. An alternative approach would be diffing functions that simply take in objects `a` and `b` which encapsulate all this information. Of course, this just moves the complexity to a different place. Nonetheless, it may well be that encapsulation turns out to be a better approach. Having tried this a couple different ways I'd like to try this dependency injection scheme first.

When we want to support the "blanket" requests mentioned above ("Diff this HTML will all applicable differs"), we will insert a layer into the server and add additional fields to the configuration. This can be done in a smooth, non-breaking way.

jeffreyliu:
I like it. Separating dispatch logic from diffing logic feels natural. For
the dispatcher, perhaps you want something more rich than just the function
signature to identify its inputs? I'm imagining something like a
registration function that allows differs to specify things like encodings
in addition to data types.

My use case example is like two image differs, one might compare some
aspect of the byte streams, which might only support a given encoding, and
one might render the images and compare pixels, which might not be as
sensitive to the specific encoding. I'm not sure that the function
signature alone is rich enough to distinguish between the two?

danielballan:
Agreed. The function signature is good to specifying what the differ _needs_ but probably not whether is it _applicable_. Deciding which differs are applicable may be too difficult to express "descriptively" (i.e., in a JSON file) and might just needed to be hard-coded into our dispatcher. Either way, the correct way to do this will probably take some exploring to discover, and we don't need it for "Version 0" / feature parity with the current workflow.

jeffreyliu:
Yup, agreed, and it's something that having separate dispatcher logic
facilitates anyway.

web_monitoring/diffing_server.py
```diff
@@ -0,0 +1,165 @@
...
+        for query_param, res in zip(('a_hash', 'b_hash'), (res_a, res_b)):
+            try:
+                expected_hash = query_params.pop('a_hash')
+            except KeyError:
+                # No hash provided in the request. Skip validation.
+                pass
+            else:
+                actual_hash = hashlib.sha256(res.body).hexdigest()
+                if actual_hash != expected_hash:
+                    self.send_error(500)
```
Mr0grog: Seems like it might be nice to send back a message indicating more clearly that the problem was in the hash not matching the retrieved content.

Mr0grog:
This looks awesome to me. I made one comment inline, and have one other thought:

I think it might be good to put each differ implementation in a separate module (so we have `/web_monitoring/differs/compare_length.py`, `/web_monitoring/differs/side_by_side_text.py`, etc). Since the intention is to expand this quite a bit, I suspect having all differs in that one file will get unwieldy soon. I was already spending a little effort figuring out which functions in there were just support for one of the diff implementations.

danielballan:
Just noting that @Mr0grog and I agreed on Slack to punt the question of splitting this up into one public function per module to a different PR.

PR #60, opened 2017-06-23
Pagefreezer update and diff service demo

janakrajchadha:
I've added a function to get the cabinetID of a given url. 
This was required as a developer may need to view a certain archive which is stored in a specific cabinet with an ID.
This allows us to get the cabinet ID of the archive we want and thus, eliminate the need to go through all cabinets to find the archive.
I've added the test for it as well.

I've also added a Jupyter notebook which allows us to test the different diff service functions separately in case we do not wish to run all tests.

Also, fixed a variable typo error in  pagefreezer.py


janakrajchadha:
https://github.com/edgi-govdata-archiving/web-monitoring/issues/53

janakrajchadha:
@danielballan Review please, I don't seem to have the permission to add the reviewers in the section on the right.

web_monitoring/pagefreezer.py
```diff
@@ -102,7 +102,7 @@ def report(self):
         print(counts)
 
     def run_query(self):
-        self.query_result = compare(self.url1, self.url2)['result']
+        self.query_result = compare(self.url_1, self.url_2)['result']
```
danielballan: Good catch.

web_monitoring/pf_edgi.py
```diff
@@ -20,6 +20,17 @@ def list_cabinets():
     assert content['status'] == 'ok'  # business logic is OK
     return content['cabinets']
 
+def get_cabinetID(url):
```
danielballan: Minor quibble: can it be `cabinet_id` for consistency with the existing function signatures (such as `list_archives(cabinet_id)`?

web_monitoring/pf_edgi.py
```diff
@@ -20,6 +20,17 @@ def list_cabinets():
...
+    cabinets = list_cabinets()
+    cabinetID = ''
+    for key,value in cabinets.items():
+        for dictionary in value:
+            if(str(urlparse(dictionary['url']).scheme + '://' + urlparse(dictionary['url']).netloc) == url):
+                cabinetID = dictionary['name']                
+
+    assert cabinetID != ''
```
danielballan: Could we raise a `ValueError` with a message (something like "no such url") instead of a bare `AssertionError`?

danielballan:
> I've also added a Jupyter notebook which allows us to test the different diff service functions separately in case we do not wish to run all tests.

The notebook is nice! Also, in case you haven't seen it yet, you can also or run specific tests using:

```sh
python run_tests.py web_monitoring/tests/test_pf.py
```

and other py.test filter like `-k`.

web_monitoring/pf_edgi.py
janakrajchadha: Absolutely! It should be that, I should've noticed this earlier.

janakrajchadha:
@danielballan I have seen the specific tests. The current test file does not use all of the pagefreezer methods. I did think about adding more test cases to `test_pf.py`, but a notebook seemed like a better way to go to as there are functions which make use of the IPython module and things are displayed in a clean and crisp way in a notebook. Also, it gives one a chance to experiment with the results of the different function. I'll probably add more to it soon.

danielballan:
This looks great. Clearly we have remaining questions about the meanings of some fields, but _some_ documentation is generally better than _no_ documentation, so I'm going to merge this as is and iterate from here.

PR #64, opened 2017-07-06
FIX: Repond with the correct Content-Type

danielballan:
passing a dict to the 'write' method automatically sets the content-type


danielballan:
This is a minor fix -- self-merging

PR #65, opened 2017-07-08
Add a special case for 'file:///' requests for local development.

danielballan:


Mr0grog:
@janakrajchadha Have you looked at this at all? Is there any reason not to merge it?

janakrajchadha:
@Mr0grog I can't think of a reason not to merge this. I think Dan must've already tested it.

Mr0grog:
Thinking about this slightly more; this is *really* insecure. I’m thinking we should add an env var. Maybe `APP_ENV=production|development|test`? If not set to `production`, then enable file URLs. What do you think?

janakrajchadha:
That would also require us to make sure the var is set to production when we push this into production and cannot be altered by the users then.
Another way, which is perhaps rudimentary, would be to create a separate file with the classes for local deployment and not include that file in the production copy. 
Maybe I'm wrong, I have minimal knowledge of making secure software packages and I think you'd be the best judge of the better alternative.

Mr0grog:
> That would also require us to make sure the var is set to production

Yeah, it would become configuration. I don’t think that’s really meaningful overhead; we already have to set other vars in production, e.g. the PageFreezer key.

> and cannot be altered by the users then.

I think that’s kind of the point, though, right? Have a server able to read and output any local file it can access (perhaps including its own configuration) is not something you’d ever want to have running live.

janakrajchadha:
@Mr0grog Okay, I'm a little confused. Just to make sure we are on the same page, are you suggesting we put it in the config file or some other way?

Mr0grog:
> are you suggesting we put it in the config file or some other way?

I’m suggesting we make it an environment variable.

Sorry if that was confusing; I wasn’t thinking about the `diff_config.json` file, which doesn’t really have a way to store any information besides a mapping of differ names to functions. I think it would be weird to modify that, especially since we already store most config data in the environment (e.g. `PAGE_FREEZER_API_KEY`, `WEB_MONITORING_DB_URL`, etc).

We could make it more specific, like `ACCEPT_FILE_URLS` instead of having it be more of a production/dev switch. I was thinking about prod/dev because it would also give us a nice way to turn on/off Tornado’s `debug` setting at the same time.

The environment is usually considered part of the app’s configuration, and in many circles, is considered *the* way to do config. https://12factor.net/config has a good write-up on why. There’s plenty of room for debate there, though.

janakrajchadha:
I'm sorry, I think I didn't interpret it correctly.
I think making it an env variable seems like a good way to go.
If we don't have any other suggestions, we should go ahead with it.

Also, thanks for the article, it's really helpful for me!

Mr0grog:
@danielballan what are you feeling on this since you aren’t gone anymore? Do you want to take this and clean it up? Should I take it? Just close it out?

danielballan:
I cleaned it up. Good catch about the obvious security hole. Of course this will go away once we switch from a server to queues, but this is useful in the meantime for testing differs with web-monitoring-ui.

web_monitoring/diffing_server.py
```diff
@@ -48,8 +60,23 @@ def get(self, differ):
...
+                    body = f.read()
+                    res_a = MockResponse(a, body, headers)
+                with open(b[7:], 'rb') as f:
+                    body = f.read()
+                    res_b = MockResponse(b, body, headers)
+            else:
+                self.send_error(
+                    500, reason=("Local files can only be used in development "
+                                 "or testing environment."))
+        else:
```
janakrajchadha: What happens when only one of the arguments is a local file? Although I don't see that happening frequently, I think the code should cover all cases.

danielballan: That's a valid point. Rewriting to deal with that is probably a ~30-line change, and we're planning to delete all this code (in favor of a queue-based system) in ~weeks, so I think I'll punt if unless anyone feels strongly about it.

danielballan:
(Closed and re-opened to trigger Circle-CI webhooks. Not sure it worked....)

web_monitoring/diffing_server.py
janakrajchadha: I think we can skip it if the queue-based system isn't far.

janakrajchadha:
@danielballan I think it didn't trigger because you've added commits to your fork and not a branch of the main repo. I'm not sure about this, could be some other issue.

danielballan:
I see. I have much more experience with the idiosyncrasies of Travis-CI than Circle-CI (but Circle does seem to be better these days). Let me see if we turn on testing for PR branches...

In the meantime, closing in favor of #88.

web_monitoring/diffing_server.py
Mr0grog: > and we're planning to delete all this code (in favor of a queue-based system) in ~weeks

Wait, are we? I thought queues were only meant to facilitate knowing when new versions have been added in order to do pre-analysis. We would still use requests for generalized diffing (e.g. when someone wants to compare version 1 and version 5 of a page), wouldn’t we?

Issue #67, opened 2017-07-20
Diff-match-patch seems to be working poorly


Mr0grog:
This might just be that we need to twiddle the settings a bit, but we are getting pretty poor diff results from the current diff-match-patch library. In this example from the UI, we should *not* just have two big blocks. Also odd is that the defaults for this library and `go-calc-diff` are the same, but their results are very different:

![dmp-side-by-side](https://user-images.githubusercontent.com/74178/28443491-3191c470-6d6b-11e7-8200-6b23f3be358f.png)

Implementation in this repo on left, `go-calc-diff` on right.

I briefly tried switching to [`diff-match-patch-python`](https://pypi.org/project/diff_match_patch_python/), and it seems to get nice results. Check out the [`switch-to-dmp-python` branch](https://github.com/edgi-govdata-archiving/web-monitoring-processing/compare/switch-to-dmp-python).

b5:
wow, that is odd. If switching out the package yields usable results I'd vote for just dropping that in & moving on. We're going to have to spend a bunch of time tuning & fiddling anyway

Mr0grog:
Thanks. I think I’ll leave this over the weekend in case @danielballan is able to take a look and knows anything I don’t. If not, I’ll just merge that branch in on Monday.

janakrajchadha:
@Mr0grog I had discussed this with @danielballan a few days back. I had found python libraries for it but I had never tried them together and tested the results. From what I have seen and read, `diff-match-patch-python` seems to be the better option. I'll take up this issue and get it done by Monday.

janakrajchadha:
@Mr0grog I've added a few things on top of your changes. You can find them [here](https://github.com/janakrajchadha/web-monitoring-processing/blob/master/web_monitoring/differs.py). Please take a look and I'll create a PR if you think it's fine.

janakrajchadha:
@Mr0grog Any comments on this?

Mr0grog:
So sorry, I didn't see your first request for review. I think this makes sense to PR. Quick notes:

- I don’t think renaming the DMP functions to `d` and `d_b` makes sense here. It did before because we were creating an instance of an object. These are now just static functions, so this isn’t doing much besides obscuring the function call. (It also might make sense to just directly import the two functions we are using instead of asking for `diff_match_patch.diff`, for example)

[...]

PR #68, opened 2017-07-21
Filtering work

janakrajchadha:
~~WIP - DO NOT MERGE~~

This PR has been created for review and feedback from other developers as new filtering functionalities are added and caching is implemented and improved in the next few days.


Mr0grog:
Do you want review and feedback on this now, or are you still working?

janakrajchadha:
I'm still working on this, I'll request a review as soon I think I've completed a sufficient part of the work.

janakrajchadha:
@suchthis I've pushed some of the filters which I've tested and they're working fine. I've also worked on the `diff-match-patch` issue and pushed those changes as well.
I was thinking that we can merge this after a review and create new PRs later when I add more filtering.

web_monitoring/differs.py
```diff
@@ -51,11 +51,20 @@ def pagefreezer(a_url, b_url):
...
     # validate it against the expected hash.
-    return web_monitoring.pagefreezer.compare(a_url, b_url)
+    return web_monitoring.pagefreezer.PageFreezer(a_url, b_url)
 
+def compute_dmp_diff(a_text, b_text, timelimit=4):
+    diff_codes = {'=': 0, '-': -1, '+': 1}
 
-d = diff_match_patch.diff_match_patch()
-
+    if(isinstance(a_text, str) and isinstance(b_text, str)):
```
mhucka: Minor point: as a matter of style, I think we're following guidelines that recommend all `if` and `elif` should use a space before the parentheses; i.e., `if (...)` not `if(...)`. (This applies not just to this line but to all of the code below.)

web_monitoring/filtering.py
```diff
@@ -0,0 +1,58 @@
...
+def df_filter(df):
+
+    df['review'] = 'yes'
+    df['id'] = 'none'
+    df['priority'] = 1.0
+    df['annotation'] = 'no annotation'
+    
+    for index,row in df.iterrows():
+        if((str(row['new']).lower() in month_list) and (str(row['old']).lower() in month_list)):
+            df.loc[index]=df.loc[index].replace(df.loc[index]['review'],'no')
```
mhucka: Minor style point: it helps readability to put spaces on both sides of `=` in assignments, and also after commas' in argument lists (here, before the `'no'`).

web_monitoring/differs.py
```diff
@@ -1,5 +1,5 @@
-from bs4 import BeautifulSoup
-import diff_match_patch
+from bs4 import BeautifulSoup   
+from diff_match_patch import diff,diff_bytes
```
Mr0grog: As noted in a few other places, space after the comma, please!

web_monitoring/differs.py
```diff
@@ -51,11 +51,20 @@ def pagefreezer(a_url, b_url):
     # Just send PF the urls, not the whole body.
     # It is still useful that we downloaded the body because we are able to
     # validate it against the expected hash.
-    return web_monitoring.pagefreezer.compare(a_url, b_url)
+    return web_monitoring.pagefreezer.PageFreezer(a_url, b_url)
 
+def compute_dmp_diff(a_text, b_text, timelimit=4):
+    diff_codes = {'=': 0, '-': -1, '+': 1}
```
Mr0grog: This is a really minor optimization, but it might make sense to pull this definition out of the function so you only allocate the dictionary once.

web_monitoring/differs.py
janakrajchadha: Noted 👍

web_monitoring/filtering.py
janakrajchadha: @mhucka As you may have noticed, I have followed these rules for some other parts of the code. I'll make sure I do a style and readability check before pushing code.

web_monitoring/differs.py
janakrajchadha: Thanks for the catch @Mr0grog!

janakrajchadha:
@Mr0grog As we still don't have concrete plans for caching, I'm removing it from this PR. Now that I think of it, I should have created separate PRs in the first place itself as the two issues are not connected to each other.

Data from different sources.md
```diff
@@ -63,14 +63,15 @@ Versionista returns a JSON blob which contains the following fields:
 - `versionId` : Id of a version in Versionista.
 - `url` : The full URL to view this version in Versionista. You’ll need to be logged into the appropriate Versionista account to make use of it.
 - `date` : The date and time when the version was captured.
-- `hasContent` : Indicates if the version has any content or not. `True` or `False`
+- `hasContent` : Indicates if Versionista has stored any conetent of the page or not. There is a limit on the size of the versions Versionista can store. `True` or `False`
```
suchthis: Typo here in "conetent"

web_monitoring/differs.py
```diff
@@ -65,13 +74,14 @@ def html_text_diff(a_text, b_text):
     ------
     >>> html_text_diff('<p>Deleted</p><p>Unchanged</p>',
     ...                '<p>Added</p><p>Unchanged</p>')
-    [(0, '<p>'), (-1, 'Delet'), (1, 'Add'), (0, 'ed</p><p>Unchanged</p>')]
+    [[-1, 'Delet'], [1, 'Add'], [0, 'ed Unchanged']]
```
suchthis: Assuming "Delet" here is a typo?

Data from different sources.md
janakrajchadha: Oh, I can't believe I missed that.

web_monitoring/differs.py
janakrajchadha: It isn't a typo @suchthis, this is the result from the algorithm. If you see the two strings, `Deleted Unchanged` and `Added Unchanged`, you'll see that only the part before `ed` in `Deleted` has changed, the remainder of the string doesn't change. The function returns what part of the string has been removed with the tag -1, addition with +1, and no change with 0.

Mr0grog:
Ok, aside from the typo fix above, does anything else need to be done to merge this?

@janakrajchadha are you able to deploy these updates, or do we need to work together to understand the current deployment (and then document it)?

Mr0grog:
Hmmm, just realized this patch does not update `requirements.txt`, so the doffs will break here. That needs to be fixed.

janakrajchadha:
@Mr0grog I'm not really sure what deploying would mean in this context. I think we need to create an script which takes versions from the db, passes it through the filtering and then posts the updated annotations back. I'm almost done with the functions required to do this but we haven't had any discussion around automating this process to run at regular intervals.
Also, I've updated `requirements.txt`, will be pushing commits today.

Mr0grog:
> I'm not really sure what deploying would mean in this context.

Sorry, I mean deploying the doffing server, which has major critical changes in this PR. I'd like to not have to keep telling people "ignore the fact that this diff is totally wrong" in demos :p 

Filtering and analysis stuff needs more work before we even have a meaningful concept of deployment for them, exactly as you are saying. That doesn't seem relevant to this PR; it should be proposed and merged separately.

janakrajchadha:
@Mr0grog Oh, okay, makes sense.

I'm still reading a little before I start making changes to the diffing functions.
You are talking about the issues mentioned [here](https://github.com/edgi-govdata-archiving/web-monitoring/issues/48), right?

Please take a look at the latest commit and tell me if you see this PR fit for merging.

Mr0grog:
> You are talking about the issues mentioned [here](https://github.com/edgi-govdata-archiving/web-monitoring/issues/48), right?

Oh, no! We need to work on those, too, but I’m talking about #67, the code for which is already in this PR.

janakrajchadha:
@Mr0grog I think I pushed the `db.py` to master in a hurry. I've moved it to another branch. I think we can merge this now.

web_monitoring/db.py
```diff
@@ -44,4 +44,4 @@ def query_import_status(import_id):
     url = WEB_MONITORING_SHOW_IMPORT_API.format(db_url=settings['db_url'],
                                                 import_id=import_id)
     return requests.get(url,
-                        auth=(settings['db_email'], settings['db_password']))
+                        auth=(settings['db_email'], settings['db_password']))
```
Mr0grog: This is a minor nit, but can you make sure to end files with a blank line to make Unix tools more happy?

web_monitoring/differs.py
```diff
@@ -1,13 +1,14 @@
-from bs4 import BeautifulSoup
-import diff_match_patch
+from bs4 import BeautifulSoup
```
Mr0grog: Also really minor, but we should clean up the spaces at the end of the line here. Do you use `editorconfig` at all? It can really help if your editor supports it.

Mr0grog:
Two suuuuper minor formatting issues, but otherwise, I think this is good to go. You should fix those and then just merge this.

web_monitoring/db.py
janakrajchadha: Absolutely! 😅

web_monitoring/differs.py
janakrajchadha: Yep, just found the plugin for sublime text.
Thanks!

janakrajchadha:
@Mr0grog I don't think I have the write access to merge PRs. So, I think you'll have to merge it 🙂

web_monitoring/differs.py
janakrajchadha: `editorconfig` works like a charm, thanks again!

Mr0grog:
@janakrajchadha how have you been testing this? I had to add https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/68/commits/7c2e792f2015c5e544d77d50f11b1d4463540cba in order to get the server to run.

In doing so, I also realized that your changes have broken the `pagefreezer` diff endpoint. This URL does not work, for example:

```
http://localhost:8000/pagefreezer?a=https://edgi-versionista-archive.s3.amazonaws.com/versionista2/74279-6212728/version-12308483.html&b=https://edgi-versionista-archive.s3.amazonaws.com/versionista2/74279-6212728/version-12334252.html
```

So I thought this was ready to merge, but it’s clearly not.

Mr0grog:
Specifically, it is erroring with `TypeError: Object of type 'PageFreezer' is not JSON serializable`, which is happening because the diff function is returning a `pagefreezer.PageFreezer` instance where it used to return the result of `pagefreezer.compare()`: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/68/files#diff-76dcaf15de70c0bc3fc6920ed03fd1d2R55

janakrajchadha:
I had to separately test with a file on my local because of the whole Versionista/PF discussion. I had to change it earlier myself to `from filtering`.
As for the pagefreezer issue, I had replaced it with `pagefreezer.PageFreezer().query_result` but clearly I did not save it and pushed the wrong thing. This will return only the result part of the response from the compare API.
There seems to be some issue with my VirtualBox VM, reinstalling it and fixing that right away.

web_monitoring/differs.py
suchthis: Gotcha, thanks!

janakrajchadha:
@Mr0grog After a few hours of figuring out what was wrong with my VirtualBox installation and fixing it, I tested the changes after a complete deployment of the diffing server. It is working now.
However, I'd like to point something out, the pagefreezer API does not seem to work in some cases (Not a JSON issue, the result from the API gives an error, I checked it by directly accessing the API). This is something @danielballan had noticed too and notified me in one of our discussions. Seems to be an issue which we need to bring up with the PF folks.

janakrajchadha:
@Mr0grog I tried adding a blank line manually as well as with the use of editorconfig but they seem to disappear in the commits. Take a look at `filterting.py` and `pagefreezer.py` specifically. Other changes are done.

Mr0grog:
> I tried adding a blank line manually as well as with the use of editorconfig but they seem to disappear in the commits

It looks like the blank line is there in both `filtering.py` and `pagefreezer.py`. GitHub doesn’t show it; instead it shows this red icon if you *don’t* have a blank line at the end:

screen shot 2017-08-03 at 1 20 33 pm

janakrajchadha:
Oh, that's interesting. Well, looks like no issues then.

janakrajchadha:
@Mr0grog The tests passed after I fixed the missing `coverage` and other issues. However, we need to write better tests such that they act as conclusive evidence that the code works on practical test cases and not just examples.

Mr0grog:
All the differs run fine for me and the tests are now passing, so I’m going to go ahead and merge now.

web_monitoring/differs.py
```diff
@@ -51,11 +52,20 @@ def pagefreezer(a_url, b_url):
     # Just send PF the urls, not the whole body.
     # It is still useful that we downloaded the body because we are able to
     # validate it against the expected hash.
-    return web_monitoring.pagefreezer.compare(a_url, b_url)
+    obj = web_monitoring.pagefreezer.PageFreezer(a_url, b_url)
```
danielballan: @janakrajchadha What was the motivation for this change?

janakrajchadha: There were a couple of reasons:
1. I think it's better if we only return the result part of what we get from the API which was already implemented using `query_result` (which, I think, was also the motivation behind the function itself).
2. In case we add more diff functions which utilise the results from the PF diff, it would make sense to make use of the class methods in `pagefreezer.py`.

danielballan: The class `PageFreezer` does some expensive operations on `__init__` (parse_to_df, use_filter) whose results don't matter to us here and just get thrown away. I think it was more intended for interactive use, which is why I factored out the underlying operations into functions that can be called individually.

The question of whether to result the whole response or just the 'result' part is debatable. Some of the metadata in the response (like PageFreezer's claim of how long it took to compute) could be interesting, so I judged that it's probably better not to throw it away in this layer. But reasonable people could disagree on that. :- )

janakrajchadha: Yeah, you're right. The `__init__` operations could lead to extremely slow responses when we're handling a large number of diffs. 
This reminds me of the question I had about our go-to diff service.
Given the uncertainty around the external services, this diff could probably be non-existent in the future. Do we have any updates on that yet?

danielballan: I think the plan is to retire go-to-diff, long term, but in general we do want a path to integrate with differs written in other languages, and go-to-diff might be a nice proof of concept for that. But I don't think that's a "version 0" priority.

janakrajchadha: I think our priority was setting up our own diff service which is already up (thanks to you).
What I had in mind, as a long term plan, is addition of more custom diffs written by our developers and no dependency on PF, Versionista etc which I think is pretty much along the lines of the ideas you have shared.
Also, to clarify, what I meant by our go-to diff service, was the service(s) we are currently using or relying on (PF and Versionista). So, what I meant to ask was, given that we're almost certain about not using them in the future, does it make sense to spend time on figuring out how to best use their API?

PR #72, opened 2017-08-01
Add documentation for current deployment process

### LINKED ISSUE #71: Understand and document current deployment process
The diff server that is part of this repo is currently deployed on Google Cloud. I believe @danielballan did that deployment entirely by hand. We need to get a better understanding of how that deployment works and document it so that nobody is stymied if one person is out. See, for example, web-monitoring-versionista-scraper’s docs: https://github.com/edgi-govdata-archiving/web-monitoring-versionista-scraper/blob/master/deployment.md

**This is not about rethinking deployment or coming up with a better process.** We should do that, too, but we first need to understand and clarify what we’ve already got.

Mr0grog:
Document the current, basic deployment setup. This almost the same as our actual deployment, with some naming differences and better permissions for Conda.

@mhucka @janakrajchadha does this read clearly and make sense to you? If you can test it, does it work for you?

Fixes #71.

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -63,10 +63,8 @@ Run the actual installer:
 ```sh
 $ cd web-monitoring-processing
 
-# Install these packages via conda first -- you get pre-built versions this way
-# that install faster and more reliably
-$ conda install lxml
-$ conda install pandas
+# Install any packages available through conda as pre-built binaries
+$ while read requirement; do conda install --yes $requirement; done < requirements.txt
 
 # Actually install the project
 $ python setup.py install
--- a/deployment.md
+++ b/deployment.md
@@ -1,0 +1,225 @@
+# Deployment
+
+There are a number of tools and libraries included in this repository. One is a web server `wm-diffing-server`, which can be deployed to a remote machine to be used in conjuction with the other Web Monitoring projects.
+
+
+## Set up your machine (first time only)
+
+### Set up Python environment
+
+First, SSH into your server, update the package manager, and install `nginx`, `git`, `build-essential`, and `libxml2-dev`. On Debian or Ubuntu Linux, you should run:
+
+```sh
+$ sudo apt-get update
+$ sudo apt-get install git nginx build-essential libxml2-dev
+```
+
+Next, install [`conda`](https://conda.io/), which we’ll use to manage Python versions and environments. You can install either [*Anaconda*](https://www.continuum.io/downloads) (the full-featured version with extra packages and tools) or [*Miniconda*](https://conda.io/miniconda.html) (the minimal, light-weight version). Minconda is recommended to keep the server as simple as possible.
+
+```sh
+# Find the URL of the installer you want to use.
+# For Anaconda, see: https://www.continuum.io/downloads for download URLs
+# For Miniconda, see: https://conda.io/miniconda.html for download URLs
+$ curl <conda_url> > conda_installer.sh 
+# e.g: curl https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > conda_installer.sh
+```
+
+Then run the installer. Follow the prompts, but instead of the default install location, choose `/opt/conda`. (This gets us a system-wide install so the web server can use it.)
+
+```sh
+$ sudo bash conda_installer.sh
+```
+
+Finally, ensure that you can access conda by creating a new conda group and adding yourself to it.
+
+```sh
+# Create a group for conda and add yourself to it
+$ sudo groupadd conda
+$ sudo usermod -a -G conda <your_username>
+
+# Ensure users in that group have access to conda
+$ sudo chgrp -R conda /opt/conda
+$ sudo chmod 770 -R /opt/conda
+```
+
+
+### Install the application
+
+Next, we need to actually install `web-monitoring-processing` in the `/var/www/web-monitoring-processing` directory:
+
+```sh
+$ cd /var/www
+
+# Make the directory first to ensure it does not get installed for the root user
+$ sudo mkdir web-monitoring-processing
+$ sudo chown <your_user> web-monitoring-processing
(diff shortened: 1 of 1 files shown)


Mr0grog:
I couldn’t figure out how to get `numpy` and `lxml` to install cleanly without manually installing them via conda first (also `apt-get libxml2-dev`). If you know how to get that working, fill it in! (If you install Anaconda instead of Miniconda, it also solves this problem, since they are built into Anaconda.)

Some *future* improvements I’m thinking we might want to make here:

- Run tornado servers as users with reduced privileges
- Use a process group for tornado servers in Supervisor (http://supervisord.org/configuration.html#program-x-section-settings)

Mr0grog:
Also note: you can preview this on Github here: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/71-document-deployment/deployment.md

janakrajchadha:
@Mr0grog Are you looking for a way to use a single command to install them all instead of manually installing each package?
If that is the case, we can use `while read requirement; do conda install --yes $requirement; done < requirements.txt`. This is something I had come across some time back.

janakrajchadha:
Also, assuming that someone trying to deploy it themselves will probably work on updates and other issues in the future, it is always safer to install `Anaconda` than go for the simplicity of `Miniconda`.

deployment.md
```diff
@@ -0,0 +1,225 @@
...
+Finally, ensure that you can access conda by creating a new conda group and adding yourself to it.
+
+```sh
+# Create a group for conda and add yourself to it
+$ sudo groupadd conda
+$ sudo usermod -a -G conda <your_username>
+
+# Ensure users in that group have access to conda
+$ sudo chgrp -R conda /opt/conda
+$ sudo chmod 770 -R /opt/conda
```
janakrajchadha: As mentioned in the comments, we can use the `requirements.txt` file trick and put it here instead of individually installing packages later.

deployment.md
```diff
@@ -0,0 +1,225 @@
...
+```sh
+$ cd /var/www
+
+# Make the directory first to ensure it does not get installed for the root user
+$ sudo mkdir web-monitoring-processing
+$ sudo chown <your_user> web-monitoring-processing
+
+# Clone the git repo for the project
+$ git clone https://github.com/edgi-govdata-archiving/web-monitoring-processing.git
+```
```
janakrajchadha: I think this has been wonderfully documented.
Do you think it would be helpful if we put the various commands into a single shell script and run that?
I think that'll make deployment a little easier for new users and seems to be possible to implement here. It's just a suggestion based on my little experience, I'm no expert.

Mr0grog:
> Are you looking for a way to use a single command to install them all instead of manually installing each package?

I was trying to understand whether I had something misconfigured. Ideally, `python setup.py install` should solve that and there wouldn’t be any commands that touch `requirements.txt` at all. (I’m not too surprised or concerned about needing to `apt-get libxml2-dev`, though)

> Also, assuming that someone trying to deploy it themselves will probably work on updates and other issues in the future, it is always safer to install Anaconda than go for the simplicity of Miniconda.

I think the scope of this document is about production server deployment, not local development (if we think people should be using Anaconda in development, we should say so in the installation section of the README).

I don’t know that Anaconda is necessarily a problem in production, but in general I think you’d want to keep the installed packages on a server minimal in order to reduce the number of possible conflicts, side-effects, and vulnerabilities they might introduce. Also, it looks to me like the deployment @danielballan already has running is Miniconda.

deployment.md
Mr0grog: > Do you think it would be helpful if we put the various commands into a single shell script and run that?

It would! (There are *lots* of things we could do to improve deployment). That’s out of scope for this PR, though. You should file an issue for it.

deployment.md
Mr0grog: Is this comment on the right line? (Maybe you were looking for line 68?) We still need to make sure people can use the copy of conda they just installed.

janakrajchadha: I don't see the point of installing them later. We are not changing environments anywhere, so it doesn't matter if we install them here. It just seems more _natural_ to install it right after install `Anaconda` or `Miniconda`.

deployment.md
janakrajchadha: Alright!

janakrajchadha:
What was the error you were getting when you tried installing it through the `setup.py` process?

Well, given the scope of the document, Miniconda seems the right way to go then! 👍

deployment.md
Mr0grog: I was trying to keep “installing required Python packages” together (and separate from installing system software), whether it be done via `conda install`, `pip install`, or `python setup.py install`.

Mr0grog:
> What was the error you were getting when you tried installing it through the setup.py process?

IIRC, I think I was encountering C compiler errors from `numpy`. If I manually installed `lxml` via `pip` first, then it seemed to work OK, but was still a really slow build (with a long list of compiler warnings), so it seemed like `conda install` was better all around because it used pre-built packages.

The more I think about it, your technique of iterating over the lines of `requirements.txt` and running `conda install` for each works pretty well. Will test and add to the doc.

Mr0grog:
Note tests are already fundamentally broken; there’s no new issue with this PR.

Mr0grog:
Ok, if no other comments after the changes that @janakrajchadha suggested, I’ll merge this tomorrow (Friday).

janakrajchadha:
@Mr0grog I have deployed this myself and it works without any issues.
The test issues(`coverage.py`) have been solved separately. If you think it's helpful, you can either pull those fixes and let another circle CI test run or you can merge directly, as the PR doesn't have an issue in itself.

danielballan:
:+1:

All looks good. Sorry that you all had to reverse-engineer this without better guidance.

PR #76, opened 2017-08-06
Text diff fixes

### LINKED ISSUE #70: Text diff includes some HTML code
I think this a conditional comment intended to enable/disable code in certain versions of Internet Explorer, but the text diff here starts with `<![endif]`

https://api-staging.monitoring.envirodatagov.org/api/v0/pages/c43413cb-2939-4d34-bb0f-e8ddac22e9e2/changes/e7d1062b-e5d5-42b8-8a2d-e3b7c015f7f4..8693b164-3814-494c-b3c2-a86164b42ffe/diff/html_text

That same diff also has some `<p>`, `<h1>`, and `<br />` markup (snapshot from web-monitoring-ui):

screen shot 2017-07-26 at 6 06 25 pm

janakrajchadha:
Fixes #70

### DIFF
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -1,4 +1,4 @@
-from bs4 import BeautifulSoup
+from bs4 import BeautifulSoup, Comment
 from diff_match_patch import diff, diff_bytes
 import re
 import web_monitoring.pagefreezer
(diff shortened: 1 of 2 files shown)


janakrajchadha:
@Mr0grog I've only added a simple test in `test_differs.py`. However, I've tested it on the html page you've used for the diff in the issue conversation.

danielballan:
This looks good. (I also just saw a live demo of it via video chat.)

We already depend on lxml in our requirements.txt, so if it works better here there's no cost to switching.

Issue #69, opened 2017-07-27
Add HTML diff for rendering


Mr0grog:
The most common diff view that analysts use in Versionista is the side-by-side highlighted diff:

![screen shot 2017-07-26 at 5 44 00 pm](https://user-images.githubusercontent.com/74178/28649532-6d530a1e-722a-11e7-93e8-6d0981d2f7d9.png)

It’s often hard to tell exactly what is meaningfully changing from a content perspective when looking at a source code diff—especially if, like many of the analysts, you are not a web developer who knows HTML really well. This view makes it very clear what is changing and how it relates to the page as a user sees it.

The main idea here is to be able to generate two versions of the HTML page—one with removed content highlighted (probably wrapped in `<del>` tags) and one with added content highlighted (probably wrapped in `<ins>` tags). This means the diff needs to be structurally aware and not start an insertion or deletion in the middle of a markup for a tag.

For example, it should handle scenarios like:

```
<a href="/careers/interns-post-docs.html">Interns & Postdocstoral<br/>Opportunities</a>
                  --------                ----------       -+++++++++++++++++++++++
```

Which can’t really be rendered as a web page because the diff intersects with HTML code. Instead we should get something like:

```
<a href="/careers/interns-post-docs.html">Interns & Postdocs</a><a href="/careers/post-docs.html">Postdoctoral<br/>Opportunities</a>
----------------------------------------------------------------++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

Or, even fancier, data about non-visual changes included (Versionista isn’t this fancy, but this would really help analysts):

```
<a href="/careers/interns-post-docs.html" data-wm-change="Link URL changed from '/careers/interns-post-docs.html' to '/careers/post-docs.html'">Interns & PostdocsPostdoctoral<br/>Opportunities</a>
                                                                                                                                                ------------------++++++++++++++++++++++++++++++
```

janakrajchadha:
I think once we've checked a few results from `diff-match-patch`, we can build on top of its result to add the `<ins>` or `<del>` tags.

Mr0grog:
Ok. I suspect this might be better served by an entirely different diffing algorithm that diffs node trees. Trying to parse your way through the output from diff-match-patch and determine what parts are intersecting in what way with HTML elements will certainly not be straightforward.

Mr0grog:
[...]

PR #78, opened 2017-08-10
WIP: Html diff render

janakrajchadha:
~~__WIP__~~

I've tried and tested this differ. It works well on the few test cases I've tried it on. 
There are better ways to implement some parts of the differ but most of the ones I've found include the reconstruction of the HTML tree which leads to loss of attribute data from tags like `<head>` and `<body>`. 
Some changes in the source code of `lxml` would also make this job easier but there's probably a good reason they haven't done that.

Please review @danielballan :)


janakrajchadha:
Fixes #69

Mr0grog:
@janakrajchadha What’s the status on this? Just curious since it sounded like [we came to an agreement](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/69#issuecomment-321712752) over a week ago that we should not just remove `ins` and `del` elements and add our own, but there are no updates to this PR since before then. The tests here still look for the addition of simple `ins` and `del` elements, too.

janakrajchadha:
@Mr0grog I went deeper into lxml's code to understand where we could alter it to add our own tags. Turns out that there's no parameter which allows us to pass our own tags, we'll have to alter the source code in a few places. 
Also found out that lxml uses [`difflib`](https://docs.python.org/3/library/difflib.html) in one of the layers for this diff. I had shared the possible paths to take with @danielballan, waiting for his views on it.
Understanding and reconstructing this based on our needs may take a little longer than earlier expected.

Mr0grog:
@danielballan @janakrajchadha does it make sense to merge and deploy this in the mean time, improving the implementation not to cause problems on pages with `<ins>`/`<del>` tags and the like later?

Mr0grog:
Update: decided in weekly meeting to merge this and iterate later, pending a deeper review of the existing implementation here by @danielballan.

janakrajchadha:
@Mr0grog @danielballan I was planning to create my own version of `htmldiff` based on `lxml`'s source code. As discussed earlier, `lxml` uses `InsensitiveSequenceMatcher` which is inherited from `SequenceMatcher` class of `difflib`. The problem we faced with `lxml`'s implementation is that it does not allow custom tags to be added instead of simple `<ins>/<del>` tags.
Anastasia, who has joined us on Slack, has worked on similar tools and I was lucky enough to come across a solution for our problem while going through her diff repositories.
You can take a look at her implementation of [`htmldiff`](https://github.com/anastasia/htmldiffer/blob/develop/htmldiffer/diff.py).
Also, [`lxml's htmldiff_tokens` ](http://lxml.de/api/lxml.html.diff-pysrc.html#htmldiff_tokens).
I'll start working on this.
Let me know if you have any suggestions.

danielballan:
Sounds good. Shall we leave this PR in limbo in the meantime?

Mr0grog:
Sorry I have been slow to respond. Unless someone has a PR they are ready to submit, I would not like to leave this in limbo. I would like to have *something* for our most critical diff, even if we completely replace it in the near future.

danielballan:
That is a very fair point. Post v0 a lot of our diffing stuff might get superceded by collaborative projects, but this will suffice for v0. I'll add some additional tests. @janakrajchadha, carry on with your collaborative work with Anastasia.

janakrajchadha:
@danielballan I'm not sure why I'm getting the `FileNotFound` error. I'm looking for reasons.
Would you rather have all the test cases defined in the code itself?

danielballan:
Good test cases. As you've discovered, bundling static files into a Python package properly is not straightforward. We'll probably find a good reason to do that later (and I'll show you how) but in this case I think it's easier and better to just define the test cases in the code itself.

janakrajchadha:
@danielballan Fixed, should this be merged now?

danielballan:
Great, thanks @janakrajchadha.

PR #82, opened 2017-08-16
Add raw Internet Archive importing via CDX API

### LINKED ISSUE #3: Need a function that downloads raw captured HTML from Internet Archive
It should:

* Check that URL is archived by the Internet Archive
* Retrieve a list of the URIs and capture timestamps of all versions captured by the IA.
* Formulate an 'Import' request for web-monitoring-db and POST it to the app.

It does _not_ need to harvest any HTML from IA (as previously stated on an early version of this GH issue).  We can just store the IA URI in our database; we don't need to maintain our own copy of it.

### LINKED ISSUE #46: Obtain unaltered copies of archived pages from the Wayback Machine
As noted in passing in #3, the Internet Archive Wayback Machine inserts an HTML toolbar and special JS into the archived pages it serves. These inserted code snippets are delineated with HTML comments, but it's not cleanly done.

* Is there some way to request the raw archived page -- or, even better, the original server response complete with headers?
* If not, is it possible to carefully snip out the added HTML?

The goal is to obtain an unaltered copy of the archived page that can be hashed and compared / de-duplicated against archives collected by other services. This, of course, requires byte-level fidelity.

Mr0grog:
This fixes #3 and fixes #46, and is the beginnings of work towards edgi-govdata-archiving/web-monitoring#23.

I’ve replaced `list_versions()` with an implementation that uses the CDX API and also split the work into two functions:

1. `search_cdx()` is a simple wrapper around the CDX API and iterates through and parses each line of each page of the CDX response.
2. `list_versions()` is a slightly higher-level interface to the same results that adds business logic like de-duping, resolving metadata, and raising meaningful errors.

The parsed CDX responses now also include `.raw_url` and `.view_url` attributes, which are URL from which to retrieve the raw content of the version and the viewable content (with URLs rewritten to point to resources from the archive) of the version, respectively.

Things I’m still planning to try and do here:

- [x] Pull more metadata from either the CDX record or the headers that we get when requesting the content (hard because of the `warc/revisit` problem)
- [x] Add CLI options to limit the timeframe of captures (the implementation is already there in `list_versions`)
- ~Add CLI option to limit results to a list of pages or to pages that are already in the DB~ (Moved this to #86)
- [x] Fix all the `.ipynb` files I probably broke
- ~Validate downloaded version content using the hash from the CDX result (I already added a method to calculate the CDX hash, which is a base 32 encoded SHA-1 digest of just the response body; no headers)~

### DIFF
--- a/web_monitoring/internetarchive.py
+++ b/web_monitoring/internetarchive.py
@@ -255,6 +255,13 @@ def format_version(*, url, dt, uri, version_hash, title, agency, site, status,
         Any relevant HTTP headers from response
     view_url : string
         The archive.org URL for viewing the page (with rewritten links, etc.)
+    redirected_url : string
+        If getting `url` resulted in a redirect, this should be the URL
+        that was ultimately redirected to.
+    redirects : string
+        If getting `url` resulted in any redirects this should be a sequence
+        of all the URLs that were retrieved, starting with the originally
+        requested URL and ending with the value of the `redirected_url` arg.
 
     Returns
     -------
--- a/web_monitoring/internetarchive.py
+++ b/web_monitoring/internetarchive.py
@@ -18,6 +18,7 @@ from collections import namedtuple
 from datetime import datetime
 import hashlib
 import urllib
+import re
 import requests
 from web_monitoring import utils
 
@@ -35,6 +36,8 @@ CDX_SEARCH_URL = 'http://web.archive.org/cdx/search/cdx'
 ARCHIVE_RAW_URL_TEMPLATE = 'http://web.archive.org/web/{timestamp}id_/{url}'
 ARCHIVE_VIEW_URL_TEMPLATE = 'http://web.archive.org/web/{timestamp}/{url}'
 URL_DATE_FORMAT = '%Y%m%d%H%M%S'
+MEMENTO_URL_PATTERN = re.compile(
+    r'^http(?:s)?://web.archive.org/web/\d+(?:id_)?/(.*)$')
 
 CdxRecord = namedtuple('CdxRecord', (
     # Raw CDX values
@@ -50,6 +53,29 @@ CdxRecord = namedtuple('CdxRecord', (
     'raw_url',
     'view_url'
 ))
+
+
+def original_url_for_memento(memento_url):
+    """
+    Get the original URL that a memento URL represents a capture of.
+
+    Examples
+    --------
+    >>> original_url_for_memento('http://web.archive.org/web/20170813195036/https://arpa-e.energy.gov/?q=engage/events-workshops')
+    'https://arpa-e.energy.gov/?q=engage/events-workshops'
+    """
+    try:
+        url = MEMENTO_URL_PATTERN.match(memento_url).group(1)
+    except:
+        raise ValueError(f'"{memento_url}" is not a memento URL')
+
+    # A URL *may* be percent encoded, decode ONLY if so (we don’t want to
+    # accidentally decode the querystring if there is one)
+    lower_url = url.lower()
+    if lower_url.startswith('http%3a') or lower_url.startswith('https%3a'):
+        url = urllib.parse.unquote(url)
+
+    return url
 
 
 def cdx_hash(content):
@@ -199,7 +225,8 @@ def list_versions(url, *, from_date=None, to_date=None, skip_repeats=True):
 
 
 def format_version(*, url, dt, uri, version_hash, title, agency, site, status,
-                   mime_type, encoding, headers=None, view_url=None):
+                   mime_type, encoding, headers=None, view_url=None,
+                   redirected_url=None, redirects=None):
     """
     Format version info in preparation for submitting it to web-monitoring-db.
 
@@ -248,6 +275,10 @@ def format_version(*, url, dt, uri, version_hash, title, agency, site, status,
     if status >= 400:
         metadata['error_code'] = status
 
+    if redirected_url:
+        metadata['redirected_url'] = redirected_url
+        metadata['redirects'] = redirects
+
     return dict(
          page_url=url,
          page_title=title,
@@ -284,13 +315,18 @@ def timestamped_uri_to_version(dt, uri, *, url, site, agency, view_url=None):
         if k.startswith(prefix)
     }
 
-    # TODO: pass along info about redirects? e.g:
-    # pattern = re.compile(r'^http://web.archive.org/web/\d+id_/(.*)$'
-    # map(
-    #   lambda response: pattern.match(response.url).groups(0),
-    #   res.history)
+    redirected_url = None
+    redirects = None
+    if res.url != uri:
+        redirected_url = original_url_for_memento(res.url)
+        redirects = list(map(
+            lambda response: original_url_for_memento(response.url),
+            res.history))
+        redirects.append(redirected_url)
+
     return format_version(url=url, dt=dt, uri=uri,
                           version_hash=version_hash, title=title,
                           agency=agency, site=site, status=res.status_code,
                           mime_type=content_type[0], encoding=res.encoding,
-                          headers=original_headers, view_url=view_url)
+                          headers=original_headers, view_url=view_url,
+                          redirected_url=redirected_url, redirects=redirects)
--- a/web_monitoring/internetarchive.py
+++ b/web_monitoring/internetarchive.py
@@ -69,14 +69,20 @@ def search_cdx(params):
     Note that even URLs without wildcards may return results with different
     URLs. Search results are matched by url_key, which is a SURT-formatted,
     canonicalized URL:
-      - Does not differentiate between HTTP and HTTPS
-      - Is not case-sensitive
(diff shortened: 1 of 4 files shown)


Mr0grog:
For reference, metadata we should be grabbing from responses:

- HTTP status code (amazingly, this is the original responses code, even for 404 or 500 responses!)
- `Content-Type` header (would be nice to take this from CDX, but for “WARC revisits,” we don’t get this info. IA folks have an update they are working on that adds a `resolveRevisits` parameter, but it’s not totally reliable or even released yet)
- `X-Archive-Orig-*` headers:
  - `X-Archive-Orig-content-length`
  - `X-Archive-Orig-strict-transport-security`
  - `X-Archive-Orig-accept-ranges`
  - `X-Archive-Orig-expires`
  - `X-Archive-Orig-server`
  - `X-Archive-Orig-last-modified`
  - `X-Archive-Orig-connection`
  - `X-Archive-Orig-etag`
  - `X-Archive-Orig-cache-control`
  - `X-Archive-Orig-date`
  - etc… maybe do this generically? Not sure if we really care about all of these.
- Chain of redirects (I’m thinking we store the result under the original URL as Versionista does, but we could also generate new pages here from the URLs that were redirected *to*—probably lots to discuss there.)

web_monitoring/internetarchive.py
```diff
@@ -28,90 +31,157 @@ class UnexpectedResponseFormat(WebMonitoringException):
...
-    Yield (version_datetime, version_uri) for all versions of a url.
+    Search archive.org's CDX API for captures of a given URL. This will
+    automatically page through all results for a given search.
+
+    Returns an iterator of CdxRecord objects. The StopIteration value is the
+    total count of found captures.
+
+    Note that even URLs without wildcards may return results with different
+    URLs. Search results are matched by url_key, which is a SURT-formatted,
+    canonicalized URL:
```
danielballan: Nitpick: When we render these docs into HTML we'll need a blank line here.

danielballan:
This looks great. This is a good use of `namedtuple`. Thanks for keeping the docstrings tidy and up to date.

web_monitoring/internetarchive.py
Mr0grog: Will update. I’ve never done doc generation for Python before; is there a standard convention we’re following here? I’m assuming the formatting is reStructuredText?

Mr0grog: Ah, is this Numpy’s style? https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

danielballan:
Was there more you wanted to do here, @Mr0grog?

Mr0grog:
Yeah, the other checkbox items from the PR description. At the very least, it seems like I should be extracting the response status code and mine type for `source_metadata` and I need to fix the notebook file.

Was also planning to do CDX hash validation, though I don't know how critical that is given that, on a redirect, the hash is for the first response, not the last one.

Finally, was planning to add functionality for querying a list of URLs (or file containing a list?), since we could probably do it more efficiently by grabbing more than one page at a time in a CDX search with a wildcard. That could get completely located, though, so I can split it into a different PR.

Mr0grog:
OK, added more metadata. Some questions here:

- We are now storing all the the `X-Archive-Orig-*` headers (minus the `X-Archive-Orig-` prefix). Does that seem right, or should we pick a particular list of ones we care more about?

- Should we keep more information for redirects? Two things I’m thinking of:
  - The final, resolved URL (e.g. a page with the `url: "https://www.epa.gov/climatechange"` redirects to `https://www.epa.gov/sites/production/files/signpost/cc.html`, so we might have `{"resolved_url": "https://www.epa.gov/sites/production/files/signpost/cc.html"} in the metadata)
  - If redirects are involved, the history of redirects, e.g. requesting `` could have metadata:

    ```js
    {
      "redirects": [
        "http://www.epa.gov/climatechange"
        "https://www.epa.gov/climatechange",
        "https://www.epa.gov/sites/production/files/signpost/cc.html"
      ]
    }
    ```

    (Or some subset of that? Not include the original URL since implied by page URL? Not include the final URL since it might be included separately above as `resovled_url` or `final_url`?)

- `timestamped_uri_to_version()` feels like its argument list is getting a little messy and long in the tooth. Should we change it to take a `CdxRecord` instance?

- Likewise `format_version()` feels like it’s getting a little crazy. Should we remove some of these keyword args in favor of an `additional_metadata` dict? I can see both ways, here; maybe not that big a deal.

Mr0grog:
Ok, I added redirect data, since after the analyst meeting yesterday, it seemed pretty valuable. Still looking to hear feedback on the interfaces to `timestamped_uri_to_version()` and `format_version()`.

If there was a redirect, two properties will show up in metadata:

- `redirected_url`: The final URL that was redirected to
- `redirects`: List of URLs in the chain of redirects. The first URL is the orginal URL that was requested, while the last is final, resulting URL (`redirected_url` above)

Also curious if any thoughts on:

- Should `redirected_url` always be present and named something like `resolved_url`?
- Should `redirects` not include the first entry (so it is URLs redirected *to*) or not include the last entry (so it is URLs redirected *from*)?

Mr0grog:
Removed validating hashes from the to-do list here. I implemented it, but it appears to fail in a large number of cases where spot-checking the content appears fine.

I’m not sure what’s wrong and have posted to the Internet Archive Slack, but I don’t feel like it’s worth waiting on.

---

For context, if anyone wants to chew on it, here’s an example problem case:

Searching for https://www.niehs.nih.gov/about/od/ocpl/*

CDX Request: http://web.archive.org/cdx/search/cdx?url=https%3A%2F%2Fwww.niehs.nih.gov%2Fabout%2Fod%2Focpl%2F*&collapse=digest&resolveRevisits=true&showResumeKey=true

Hit the line:

```
gov,nih,niehs)/about/od/ocpl/contact 20130224101247 http://www.niehs.nih.gov:80/about/od/ocpl/contact text/html 301 3UOLV74SEVGVB56Y47BSERHSQCSDTGUR 394
```

`GET http://web.archive.org/web/20130224101247id_/http://www.niehs.nih.gov:80/about/od/ocpl/contact`
  - DIGEST: `FK6D4KAJK36ZT4S6N24YU6AEEWFIOILO`
  - Redirects to http://web.archive.org/web/20130301174940id_/http://www.niehs.nih.gov/about/od/ocpl/contact/

`GET http://web.archive.org/web/20130301174940id_/http://www.niehs.nih.gov/about/od/ocpl/contact/`
  DIGEST: `TJG34KML56CYZQMW2ZFSQWYDFEFHUEV5`

Neither of these is `3UOLV74SEVGVB56Y47BSERHSQCSDTGUR`

Using the following method of calculating the digest, which seems to work most of the time:

1. Calculate SHA-1 digest of response body
2. Base 32 encode the result

Or, in Python: `base64.b32encode(hashlib.sha1(body).digest()).decode()`

Mr0grog:
Updated the notebook file. @danielballan it seems like this notebook was broken to begin with; are we actually maintaining it?

Also, not sure on the best practices for storing them in the repo; I pared down the changes to just the relevant bits of code in each cell. There is a lot of other stateful stuff that comes along for the ride otherwise.

janakrajchadha:
@Mr0grog This notebook was added by me as a part of the documentation for the data formats.
I think the import issue was caused because it was earlier in the `web_monitoring` folder. 
Thanks for the catch!

I usually use notebooks to test code and I thought it might be useful to keep a notebook demo which displays real examples of data from the different sources.
@danielballan definitely knows more about the best practices to store notebooks in a repo.

danielballan:
We have had some notebooks in the repo from the beginning (originally from the NYC event) so I have been a little loose about handling them in git, but soon we should apply better practices. Notebook output (which is often stateful and sometimes includes big base64 blobs) should be stripped before committing. This can be done automatically with git hooks.

If output is desired for documentation, it should be rendered as part of a documentation build process.

But I think we can clean all this up in a separate PR.

Mr0grog:
@danielballan I think the remaining item (limit to pages already in DB) probably requires some work on the DB side, too (e.g. edgi-govdata-archiving/web-monitoring-db#128), so I’m thinking we should just merge this and do that work separately. Thoughts?

Mr0grog:
OK! Splitting off the “limit results to a list of pages or to pages that are already in the DB” work into a separate issue/PR (#86). This needs a final review, but should otherwise be good to merge.

Mr0grog:
Oh, also, if you have feedback on https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/82#issuecomment-324153034, @danielballan, I’m happy to make changes to those parts.

danielballan:
As per discussion at the last webmonitoring dev meeting, we'll merge and iterate. I need to play with this for awhile before I can offer useful opinions, and I don't want to block progress.

I'll address the notebook file stuff in a separate PR.

Mr0grog:
FWIW, I think the notebook file stuff is fine—I fixed it up. (Though it would be nice to have a pre-commit hook that cleans it up automatically.)

Issue #81, opened 2017-08-16
Important change classification - text features


janakrajchadha:
Based on the plan described in [edgi-govdata-archiving/web-monitoring/issues/67](https://github.com/edgi-govdata-archiving/web-monitoring/issues/67), I'm laying out a tentative plan for the first pass of important change classification by looking at text changes. Separate issues will be opened for modifications over time.

To begin with, we need to do two things:

- [x]  [Create a dataset which will be fed to the model](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/83)
- [x]  [Create basic learning model framework for text classification tasks](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/84)

These two tasks can be tackled independently.

After creating a dataset, it will be fed to the model which will then be evaluated.
The evaluation will be done in two ways:
1. Dataset validation - A common Machine Learning practice is to use a part of the dataset as a test/validation set to evaluate the model.
2. Feedback from analysts - The results of the model will be shared with analysts who'll evaluate the performance of the model.

- [ ]  Evaluate model and improve it

The entire process follows a general life cycle described in the picture below
![ml_lifecycle](https://user-images.githubusercontent.com/13016251/29369916-94dcbfa6-82c1-11e7-8ba1-1b0a84d3468d.png)

Updates on the first two tasks soon.

janakrajchadha:
@danielballan @suchthis @mhucka I'm partially done with the framework, which is one of the two independent tasks I've mentioned above. 
I'm also adding functions to `db.py` which will help me to get the content of the versions for the dataset.
More on this soon 😄

PR #85, opened 2017-08-20
WIP: Machine learning work

janakrajchadha:
__WIP:DO NOT MERGE__

Work related to issues described in #81 

Upcoming work

- [x] Dataset creation through version content diffing

- [x] Simple text classification model sample

Requesting review for completed work @danielballan


machine_learning/create_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    -------
+    df : Pandas DataFrame with information about changes between
+                versions as rows
+    """
+    path = os.path.join(os.getcwd(), dirname)
+    filelist = os.listdir(path)
+    os.chdir(path)
+    columns = ['deleted', 'added', 'combined', 'no. of changes']
+    df = pd.DataFrame(columns=columns)
+    for index in range(0,len(filelist),2):
```
mhucka: Trivial style niggle: I think we're following the convention of putting spaces after commas. I'll mention just this one case, but it might be best to search the files for other possible cases. A very low priority item, though.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,147 @@
...
+    filename* : string
+        The csv file which has the urls in the 'Last Two - Side by Side' column
+    dirname* : string
+        The directory name in which you want to download the content, creates a new directory
+    path : string (default : current working directory)
+        Path to the directory where the csv is stored and downloaded content will be stored
+    """
+    if (path == ''):
+        path = os.getcwd()
+    os.mkdir(dirname)
```
mhucka: I'm not sure if the calling context is assumed to handle exceptions, but if not, I think it would be safer to check for possible exceptions from os.mkdir. This is especially the case since no checking is being done on `dirname`, which means accidentally handing it some values may cause `os.mkdir` to be unable to create the directory.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,147 @@
...
+    new_dir = os.path.join(path, dirname)
+    download_log_handler = open(str(dirname)+ '.txt', 'w')
+    filename = os.path.join(os.getcwd(), filename)
+    os.chdir(new_dir)
```
mhucka: The success of this operation should probably be checked, because if the `chdir` doesn't take place, subsequent operations will not write files to the expected locations.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,147 @@
...
+    download_log_handler.close()
+    os.chdir('..')
+    return True
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Run downloader script')
+    parser.add_argument('filename', type=str, help='CSV file which has the urls')
+    parser.add_argument('dirname', type=str, help='Directory name in which you want to download the content')
+    parser.add_argument('--path', type=str, help='Path where you want to create the new directory', default='')
```
mhucka: I'm not sure I understand why there is both a path and a directory name option for the same thing. Maybe I'm missing some underlying issue, but typically, it would be enough to use one argument that would be the path of the directory _including_ its name.  E.g., `"this/is/the/directory"`, which if I interpret the code above correctly, I think is being split up into `"this/is/the/"` and `"directory"`.

machine_learning/create_dataset.py
janakrajchadha: Yep, I was just going to do a round of formatting fixes. I try to make sure I avoid these issues in the initial commit itself, but I tend to miss a few when my mind is occupied with debugging and getting the code to run.

machine_learning/data_downloader.py
janakrajchadha: Ah, yes, I haven't added code which can handle exceptions. I've encountered an issue or two when I tried the code a few times myself, but I don't think that covers all cases. I'll add exception handling for those cases. Any specific cases that you have in mind @mhucka?

machine_learning/data_downloader.py
janakrajchadha: I think I should clarify why I did this. The `path` is where we want to store the data. The `dirname` is the name the user wants to give to the directory where the data will be downloaded.
E.g., "this/is/the/path" is the path given and "this/is/the/path/directory" is where the data will be downloaded. 
I should probably add examples in the code to make this clearer.

janakrajchadha: @mhucka I think my explanation of the argument confused you. I had written in the comments of the function `"Path to the directory where the csv is stored and downloaded content will be stored"`. 
What I meant to say here was `"Path to the directory where the csv is stored and downloaded content directory will be stored"` or better yet `"Path to the directory where the csv is stored and directory 'dirname' will be created"`. Does that clear the confusion?

janakrajchadha:
The tasks I had  in mind for this PR have been completed.
I have created tests as well but I haven't added them as I'm unsure if the test dataset can be shared publicly. Once the tests have been added and @mhucka @danielballan have reviewed the latest commits, I think it's fair to say that this PR can be merged.

danielballan:
See comments on Slack.

circle.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
 
   override:
     - python setup.py install
+    - pip install -U scikit-learn
```
danielballan: Should this be added to the requirements.txt?

danielballan: Should this be added to the requirements.txt?

machine_learning/data_downloader.py
danielballan: I think I follow what you're saying, but I don't understand the reason for using two arguments to specify this instead of one.

circle.yml
janakrajchadha: I had added it to requirements.txt in commit [d9973b0](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/85/commits/d9973b008a013efae7f0cbb3c3afe836bd4ee2fc). If I'm not wrong, the dependencies in requirements.txt are installed according to the order in the file.
The CircleCI test for this commit failed because it was trying to install scikit-learn before scipy. I tried looking for explanations and solutions but this is the only working solution I could find then (I completely forgot about it later).

machine_learning/data_downloader.py
janakrajchadha: Ah, a little something I missed. On line 100, it should be `filename = os.path.join(path, filename)`. I'm assuming that the csv file will be in the same place where you want to create the directory for downloading data. So, `path` is used with both `filename` and `dirname`.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+                                           + to_version_response.text)
+            continue
+
+    download_log_handler.close()
+    os.chdir('..')
+    return True
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Run downloader script')
```
Mr0grog: Can we please stick to one way of doing CLI arg parsing? Existing code uses DocOpt. For example: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/cli.py#L103-L114

Unless we've decided we don't want to use it at all anymore.

machine_learning/dataset_creator.py
```diff
@@ -0,0 +1,80 @@
...
+                     from_version_id, to_version_id])
+        df = df.append(pd.DataFrame([data], columns=columns),
+                       ignore_index=True)
+
+    os.chdir('..')
+    return df, True
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Run dataset creation script')
```
Mr0grog: Same as above issue on parsing.

circle.yml
janakrajchadha: As it turns out, installing scikit-learn causes problems as `setup` tries to do a simple `pip` install from the Python Package Index. A simple install fails if `numpy` and `scipy` aren't already installed on the system. 
Take a look at the installation page for scikit-learn : https://github.com/scikit-learn/scikit-learn/blob/master/doc/install.rst
Also, take a look at this issue and PR on scikit-learn's GitHub repo : 
https://github.com/scikit-learn/scikit-learn/issues/4164
https://github.com/scikit-learn/scikit-learn/pull/6990

Anaconda offers scikit-learn as a part of its distribution.
It may make sense to suggest installing Anaconda(or MiniConda) before running `setup.py`.

machine_learning/data_downloader.py
janakrajchadha: I've made the changes to DocOpt. I think it's better if we stick with it.

janakrajchadha:
@danielballan Any more comments on this?

machine_learning/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    http://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html
+    """
+    if (eval_method == 'accuracy'):
+        return accuracy_score(y_true=y_test, y_pred=y_predicted)
+    elif (eval_method == 'report'):
+        return classification_report(y_true=y_test, y_pred=y_predicted)
+    else :
+        raise ValueError('Please select a valid evalaution method.')
+
+def save_model(clf, filename):
```
danielballan: Ditto here. I have definitely felt the temptation to write conveniences like this, but I think it's better to just direct users to the underlying function. If they already know it, they'll be happy. If they don't know it, you've taught them something that might be useful in other contexts, not just in our library.

scripts/data-downloader
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
```
danielballan: Could the scripts be renamed to begin with `wm-`? I have followed this pattern to avoid colliding names with other libraries that might have similar utilities.

web_monitoring/db.py
```diff
@@ -45,3 +47,89 @@ def query_import_status(import_id):
...
+def get_version_uri(version_id, id_type='db', source_type='versionista', get_previous=False):
+    """
+    Get the uri of a version(snapshot) stored in the db.
+
+    Parameters
+    ----------
+    version_id* : string
+    source_type : string
+    get_previous : boolean
+    * -> required fields
```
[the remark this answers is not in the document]
janakrajchadha: Thanks for the reference, I'll go through it and use it for all future documentation 😃

web_monitoring/db.py
```diff
@@ -45,3 +47,89 @@ def query_import_status(import_id):
...
+                                                           version_id=version_id)
+    else:
+        raise ValueError('Id type should be either "db" or "source"')
+
+    response = requests.get(url,
+                            auth=(settings['db_email'], settings['db_password']))
+    result = response.json()
+
+    if (len(result['data']) == 0):
+        if(get_previous):
```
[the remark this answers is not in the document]
janakrajchadha: I was not too sure about what to do here. Although it does make sense to raise an Exception here for a consumer trying to get a single version uri, it causes problems when a script like `data_downloader` tries to get a big list of uris. Since the downloading script depends on the data in the csv, anyone using it will have to make sure that the data is clean and the versions exist in the db (which is a little tedious).

machine_learning/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    temp_text = re.sub('[^a-zA-Z]', ' ', raw_text)
+
+    #Convert all words to lower case and split into words
+    cleaned_text = temp_text.lower()
+
+    #Join the words together to create a space separated string
+    return cleaned_text
+
+
+def preprocess(vectorizer_func, text, labels, max_features, max_df=1.0, min_df=1):
```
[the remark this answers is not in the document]
janakrajchadha: I agree they are thin wrappers. I found these to be a convenient option when I was learning and already had some sklearn experience. 
Since I was only dealing with text models in the beginning, it was really a matter of choice when it came to these functions. However, as we try out more models, I think it may make sense to have a few preprocessing functions. 
Do you think it's better that we remove them?

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+    ----------
+    url : string
+
+    Returns
+    -------
+    from_version_id, to_version_id : string
+        To and from version ids extracted from the url
+    """
+    if (url == ''):
+        return None
```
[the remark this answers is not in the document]
janakrajchadha: Again, I was not too sure about this. While raising an error is the right thing to do, anyone using this will have to ensure that the csv data is cleaned before using it.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+
+    result = urlparse(url)
+    if (result.netloc == 'versionista.com'):
+        try:
+            ids = [element for element in result.path.split('/') if element != '']
```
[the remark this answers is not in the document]
janakrajchadha: I definitely like your trick more, I think I'll change it.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+            site_id = ids[0]
+            page_id = ids[1]
+            version_ids = ids[2].split(':')
+            to_version_id = version_ids[0]
+            from_version_id = version_ids[1]
+            return {'site_id': site_id, 'page_id': page_id,
+                    'from_version_id': from_version_id, 'to_version_id': to_version_id}
+        except:
```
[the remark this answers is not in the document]
janakrajchadha: Ah, this is all a part of what do we do when the list of uris in the csv isn't clean?
This was added to ensure that bad uris (Eg. when the uri terminates at page id and has no version ids) don't disrupt the script.
All these points returning `None` as of now are error points.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+        To and from version uris to access content in db
+    """
+    ids = parse_url(url)
+
+    if (ids == None):
+        return [None]*3
+    else:
+        from_version_id = ids['from_version_id']
+        to_version_id = ids['to_version_id']
+        if from_version_id == '0':
```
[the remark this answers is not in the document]
janakrajchadha: Some versionista uris have the format `https://versionista.com/site_id/page_id/to_version_id:0/ which automatically directs to the diff with the previous version. However, versions in our db are saved with their true version ids and when the `from_version_id` is `'0'`, this is handled by taking advantage of the optional parameters of `get_version_uri`.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+                                               source_type='versionista')
+            ids['from_version_id'] = valid_from_version_id
+        else :
+            to_version_uri = get_version_uri(version_id=to_version_id,
+                                             id_type='source',
+                                             source_type='versionista')
+            from_version_uri = get_version_uri(version_id=from_version_id,
+                                               id_type='source',
+                                               source_type='versionista')
+        return from_version_uri, to_version_uri, ids
```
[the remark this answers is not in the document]
janakrajchadha: No, I do want the ids to be returned. This was added later once I realised that it's better to save the downloaded data with the ids in the filename. I forgot to update the docstring after I added this.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+
+def download(filename, dirname, path=''):
```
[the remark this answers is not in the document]
janakrajchadha: Ah, alright. So, does this mean that we assume the presence of the file in the current working directory?

web_monitoring/db.py
```diff
@@ -45,3 +47,89 @@ def query_import_status(import_id):
                                                 import_id=import_id)
     return requests.get(url,
                         auth=(settings['db_email'], settings['db_password']))
+
+def get_version_uri(version_id, id_type='db', source_type='versionista', get_previous=False):
```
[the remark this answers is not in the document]
janakrajchadha: Yes, you got that right. Since the purpose of either option is to return the uri, I thought it would make sense to keep both in the same function.
I think splitting it does make it easier to understand but is it the best way when both the functions serve the same purpose.

machine_learning/data_downloader.py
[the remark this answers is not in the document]
danielballan: Cool. Can you put that comment into the code so that future contributors can follow what is happening?

web_monitoring/db.py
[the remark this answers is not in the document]
danielballan: For me, I think the critical consideration is whether the inputs and outputs are the same. In this case, the outputs are compatible, but there are inputs that apply to one case but not another. That's why I would split them up in the case.

web_monitoring/db.py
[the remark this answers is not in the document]
danielballan: A good point. I would say that the right place to swallow the error is in `data_downloader`. When it gets an exception, it can decide to (1) silently ignore it, (2) print a warning, (3) return a list of the uudis of the failures, or (4) blow up. By passing up a warnings (instead of `None`), you will be providing `data_downloader` with more detailed information about the _reason_ for the failure.

machine_learning/utils.py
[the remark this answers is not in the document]
danielballan: It's a judgement call. As long as you are open to _maybe_ removing them in the future (and putting them into documentation or something) I would be OK with merging this now.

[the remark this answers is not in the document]
janakrajchadha: Oh, absolutely! I think once we're sure about their usefulness, we can either keep them or put them somewhere else (documentation).

web_monitoring/db.py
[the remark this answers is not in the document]
janakrajchadha: Right, warnings will be more helpful than just a `None`. I'll print warnings and get a list of failed uuids as well (the logs should come in handy here).

danielballan:
While we iterate on the rest of this, I'm going to cherry-pick the changes to `db.py`. I would like to complete the Python API web-monitoring-db today, and I'll do so on top of your work.

janakrajchadha:
@danielballan Another round on this one? 😄

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+        To and from version uris to access content in db
+    ids : dictionary
+        Site, page, and version ids of the from and two versions extracted from the url
+    """
+    ids = parse_url(url)
+    from_version_id = ids['from_version_id']
+    to_version_id = ids['to_version_id']
+    print(ids)
+    client = Client.from_env()
+    """
```
danielballan: This is technically valid, but it would be better style to make it comment than a string. The _docstring_ immediately below the function signature is granted special significance by the language, but any additional comments further down should be comments, not strings.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+    However, versions in our db are saved with their true version ids and when the from_version_id is '0',
+    this is handled by taking advantage of the optional parameters of get_version_uri.
+    """
+    if from_version_id == '0':
+        get_to_version = client.get_version_by_versionista_id(versionista_id=to_version_id)
+        to_version_uri = get_to_version['data']['uri']
+        valid_from_version_id = get_to_version['data']['source_metadata']['diff_with_previous_url'].split(':')[-1]
+        from_version_uri = client.get_version_by_versionista_id(versionista_id=valid_from_version_id)['data']['uri']
+        ids['from_version_id'] = valid_from_version_id
+    else :
```
danielballan: stray space here

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+        to_version_uri = client.get_version_by_versionista_id(versionista_id=to_version_id)['data']['uri']
+        from_version_uri = client.get_version_uri(versionista_id=from_version_id)['data']['uri']
+    print(from_version_uri, to_version_uri)
```
danielballan: This debugging print should be removed. Library code shouldn't print, as a general rule, because callers using it have no way to suppress that printing.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+        site_id = ids[0]
+        page_id = ids[1]
+        version_ids = ids[2].split(':')
+        to_version_id = version_ids[0]
+        from_version_id = version_ids[1]
+        ids = {'site_id': site_id, 'page_id': page_id,
+               'from_version_id': from_version_id, 'to_version_id': to_version_id}
+        return ids
+    except:
+        warnings.warn('Url "' + url + '" is not in the required format and no data will be downloaded for this row', Warning)
```
danielballan: Why not let this raise so that it's easier to interactively debug?

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+    -------
+    from_version_uri, to_version_uri : string
+        To and from version uris to access content in db
+    ids : dictionary
+        Site, page, and version ids of the from and two versions extracted from the url
+    """
+    ids = parse_url(url)
+    from_version_id = ids['from_version_id']
+    to_version_id = ids['to_version_id']
+    print(ids)
```
danielballan: This stray `print` should be removed.

machine_learning/data_downloader.py
```diff
@@ -0,0 +1,160 @@
...
+    download_log_handler = open(str(dirname) + '.txt', 'w')
+    filename = os.path.join(path, filename)
+    df = pd.read_csv(filename, header=0, error_bad_lines=False)
+    new_dir = os.path.join(path, dirname)
+    os.chdir(new_dir)
+    download_log_handler.write(time.asctime(time.localtime(time.time())) + '\n')
+    urls = df['Last Two - Side by Side']
+
+    for index in range(0, len(df['Last Two - Side by Side'])):
+        try:
```
danielballan: I agree with this. This is a good use of try/except because `download` is a batch operation. I would however try to provide more information about the failure, like this:

```python
try:
   ...
except Exception as exc:
   # include repr(exc) in the log message
```

machine_learning/dataset_creator.py
```diff
@@ -0,0 +1,82 @@
...
+            text1 = f1.read()
+        with open(filelist[index + 1], 'r', encoding='utf-8') as f2:
+            text2 = f2.read()
+        data =  text_diff(text1, text2)
+        data.extend([site_id, page_id,
+                     from_version_id, to_version_id])
+        df = df.append(pd.DataFrame([data], columns=columns),
+                       ignore_index=True)
+
+    os.chdir('..')
```
danielballan: We should ensure that this line gets run (even if an error occurs above) by putting it in a `finally` block:

```python
os.chdir(path)
try:
   ...
finally:
   os.chdir('..')
```

PR #89, opened 2017-09-14
Include version in response.

danielballan:


danielballan:
Via the magic of versionista, the `__version__` attribute will include the current hash. For example, current master `web_monitoring.__version__` is `'0+untagged.200.gd34cc21'`. Tagged releases would look like `1.0.0` etc.

Mr0grog:
Oooh, nice! Quick take: would it be better to return this in a header? Should we use this to fill in [`diff_service_version` in the DB API response](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/controllers/api/v0/diff_controller.rb#L11)?

danielballan:
Definitely. I originally scoped the diff content under the 'diff' key with idea of putting metadata alongside it: the version for `diff_service_version`, an echo of the name of the diffing service as a sanity check, and some error information in the event of a bug in the differ. I'm fine with moving this to a header but I'm not familiar with the conventions for what to call it and how to organize it.

Mr0grog:
> I originally scoped the diff content under the 'diff' key with idea of putting metadata alongside it

Ah. I was thinking you’d want to use headers for metadata so diffs that return non-text content (e.g. images or PDFs) could just have the diff’s bytes for the body. If you weren’t planning on that, I wouldn’t bother with the extra complexity of headers for this.

If we wanted to use headers, I might consider `Server` or `X-Version`, otherwise just make something up like `X-Diff-Version` (and other future things would be `X-Diff-*`).

danielballan:
> Ah. I was thinking you’d want to use headers for metadata so diffs that return non-text content (e.g. images or PDFs) could just have the diff’s bytes for the body.

That makes a lot of sense. I had that in mind at some point too and evidently dropped the thought. In light of that plan, I wonder if we should remove the `{'diff': ...}` scope and just return `...` so that -db and -ui don't have to know to "unbox" results for these particular differs.

Mr0grog:
The more I think about this, the more ambivalent I’m starting to feel about it. I vaguely recall a similar conversation w/r/t the way the DB API wraps these responses and determining that binary data would just have to be base64-encoded there anyway, so maybe this isn’t really that big a deal, since we don’t really expect the UI to be calling a differ directly.

danielballan:
OK. Then I vote for putting version info (etc.) into the response body where it is (IMO) more discoverable. If we do need to pass around binary data efficiently, that's a post-v0 problem.

Mr0grog:
Great; this looks good to me, then.

danielballan:
It was useful to get clarity on our reasoning here.

PR #90, opened 2017-09-16
Add HTML rendered diff as “html_visual_diff”

Mr0grog:
This should have been done in #78, but was apparently missed. Kind of important for actually using this differ ;)


janakrajchadha:
@Mr0grog Thanks for the catch, we definitely want to use the differ 😅

Mr0grog:
Going ahead and merging this since I accidentally included it in #93 and that has the go-ahead.

PR #92, opened 2017-09-17
HTML diff render fix

### LINKED ISSUE #91: HTML diff breaks with versions that have differing number of nodes in `<html>`
If the number of nodes in the root `<html>` element of a page are different, `html_diff_render` raises an error and breaks.

For example try this one locally: http://localhost:8888/html_visual_diff?format=json&a=https%3A%2F%2Fedgi-versionista-archive.s3.amazonaws.com%2Fversionista1%2F69957-6026604%2Fversion-10666282.html&a_hash=db379f787af3c710e13312fbeb68725a10cd1bd16e6dd221da53bceb2f708cad&b=https%3A%2F%2Fedgi-versionista-archive.s3.amazonaws.com%2Fversionista1%2F69957-6026604%2Fversion-11082551.html&b_hash=6a8160fefe30aa3918f159ce7e0b3082bae491b712f8b2ac0c0fdef3a998e019

That’s from the page https://api.monitoring.envirodatagov.org/api/v0/pages/65d625ec-5f15-433e-8600-ae13937f2b5a, versions:

1. https://api.monitoring.envirodatagov.org/api/v0/versions/5afe9b05-6771-4667-bd2a-8df9e831faeb
2. https://api.monitoring.envirodatagov.org/api/v0/versions/fbb8774d-22a4-48f4-93d6-145f819a0f44

The issue is in [lines 129 and 130 of `differs.py`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/differs.py#L129-L130), where we simply iterate over the nodes of the new content and diff them with the correspondingly indexed nodes of the old content:

```py
diff_list = []
for index in range(len(new_content)):
    diff_list.append(htmldiff(old_content[index], new_content[index]))
```

This isn’t just a problem for differing numbers of nodes, but if whitespace were added, say, between the `<head>` and `<body>` and removed after the `<body>` (so same number of nodes in the end), you’d wind up diffing the `<body>` of one version against whitespace from the other.

The right move here is probably to find the `<head>`, `<body>`, and the source code surrounding them from each version and diff those explicitly.

janakrajchadha:
Not sure why the compare is showing commits merged earlier as well.
Anyway, main fix in [cfda84d](https://github.com/janakrajchadha/web-monitoring-processing/commit/cfda84d34734efde77805218056bae36a0fac87c).
Fixes #91


janakrajchadha:
@Mr0grog Do you think we should skip the `<head>` tag as well? Will we miss any crucial information if we did that (not helpful or visible in the visual diff but present in the diffed HTML)?

Mr0grog:
Oh, @janakrajchadha, did you not see that the bug was assigned to me? I was already working on #93.

No, I don’t think we should not diff the `<head>`—it’s important (it does have visual effects) if we want to render side-by-side comparisons from this.

janakrajchadha:
@Mr0grog Aah, I missed that. Off the top of my head, it looked like a simple one-line fix (at least the `<head>, `<body>` issue). I see that you've done some other work as well in PR #93. I'll go through it and I think I'll just close this if the fix here is already covered by 93. 
Sounds good?

Mr0grog:
Sure, that’s fine.

janakrajchadha:
#93 fixed #91 (and made the diff better). Closing this.

PR #93, opened 2017-09-17
Handle versions with different numbers of nodes in HTML diffs

### LINKED ISSUE #91: HTML diff breaks with versions that have differing number of nodes in `<html>`

Mr0grog:
This fixes #91. It also adds a bunch of comments to clarify why we are bending over backwards to do some strange transformations here.

While I was here, I also took the opportunity to clean up how the inserted CSS styling was clobbering character encoding on lots of pages.

### DIFF
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -4,7 +4,6 @@ from lxml.html.diff import htmldiff
 import re
 import web_monitoring.pagefreezer
 import sys
-import os
 import copy
 
 # BeautifulSoup can sometimes exceed the default Python recursion limit (1000).
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -5,6 +5,7 @@ import re
 import web_monitoring.pagefreezer
 import sys
 import os
+import copy
 
 # BeautifulSoup can sometimes exceed the default Python recursion limit (1000).
 sys.setrecursionlimit(10000)
@@ -101,9 +102,15 @@ def html_source_diff(a_text, b_text):
     TIMELIMIT = 2 #seconds
     return compute_dmp_diff(a_text, b_text, timelimit=TIMELIMIT)
 
+
 def html_diff_render(a_text, b_text):
     """
-    HTML Diff for rendering
+    HTML Diff for rendering.
+
+    Please note that the result of this should not be displayed as-is in a
+    browser -- because this contains added and removed sections of the
+    document’s <head>, it may cause a browser to load two different CSS or JS
+    files that are in conflict with each other.
 
     Example
     -------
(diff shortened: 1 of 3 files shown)


web_monitoring/differs.py
```diff
@@ -101,9 +102,15 @@ def html_source_diff(a_text, b_text):
...
+
 def html_diff_render(a_text, b_text):
     """
-    HTML Diff for rendering
+    HTML Diff for rendering.
+
+    Please note that the result of this should not be displayed as-is in a
+    browser -- because this contains added and removed sections of the
+    document’s <head>, it may cause a browser to load two different CSS or JS
+    files that are in conflict with each other.
```
Mr0grog: Just to be clear, this is not a change in behavior—this was already the case. I wanted to make sure it was extra clear, though.

janakrajchadha: I wanted to talk to you about this, especially after I saw the result of the diff between the examples mentioned in issue #91.
Important to have this in the docs.

web_monitoring/tests/test_differs.py
```diff
@@ -67,15 +69,20 @@ def test_html_diff_render():
...
+                    '<html>\n <head>\n  <style type="text/css">\n   ins {text-decoration : none; background-color: #d4fcbc;}\n                        del {text-decoration : none; background-color: #fbb6c2;}\n  </style>\n </head> <body>\n  <h3>Paragraph</h3>\n </body>\n</html>',
+                    '<html>\n <head>\n  <title>HTML <ins>Difference</ins> <del>Diff</del> Render</title>\n  <style type="text/css">\n   ins {text-decoration : none; background-color: #d4fcbc;}\n                        del {text-decoration : none; background-color: #fbb6c2;}\n  </style>\n </head> <body>\n  <h1><ins>Head</ins></h1> <h1><del>Heading</del></h1>\n </body>\n</html>']
 
     for index in range(len(test_data)):
-        test_check.append(wd.html_diff_render(test_data[index][0],
-                                              test_data[index][1]) == test_results[index])
+        diff = wd.html_diff_render(test_data[index][0], test_data[index][1])
+        assert (f'{index}: ' + test_results[index]) == (f'{index}: ' + diff)
```
danielballan: :+1: I was planning on circling back for this. Much easier to understand test failures this way. We might later go one step further and make the `test_results` into a pytest parameterization.

Mr0grog: > We might later go one step further and make the test_results into a pytest parameterization.

Faaaaaaaancy!

web_monitoring/differs.py
```diff
@@ -114,35 +120,97 @@ def html_diff_render(a_text, b_text):
...
+                        del {text-decoration : none; background-color: #fbb6c2;}"""
+    soup_new.head.append(change_styles)
 
     render = soup_new.prettify(formatter=None)
 
     return render
+
+
+def _find_meaningful_nodes(soup):
+    """
```
janakrajchadha: This is great! Thanks for the enhancement @Mr0grog 😄

PR #100, opened 2017-10-02
Remove repeat blank lines from text diffs

### LINKED ISSUE #98: Remove consecutive newlines from text diffs

Mr0grog:
Up to two repeat blank lines are allowed (which helps to get some differentiation between parts of the page), but anything more is collapsed. This also changes the output of side-by-side text to be two text strings rather than a list of text strings (since we need to collapse *across* strings).

So now we have diffs that look like:

screen shot 2017-10-02 at 12 53 15 am

rather than:

screen shot 2017-10-02 at 1 00 59 am

I also went way overboard and made an implementation of something approximating HTML’s whitespace algorithm (collapse all types of whitespace including line feeds into a single space; newlines are triggered by `<br>` and block elements), but it rarely looked better and often looked worse on most actual pages I tried it with. (It turns out that the way people write markup, various nav links and sections of the page tend to clump up meaningfully when we naively collapse blank lines, but get spaced out confusingly uniformly when we follow HTML whitespace). We could definitely get that to be something really good, but it would require a lot of work to get over the hump to be better than the naive approach in this PR, so I tabled it for now.

Fixes #98.

### DIFF
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -12,6 +12,8 @@ sys.setrecursionlimit(10000)
 # Dictionary mapping which maps from diff-match-patch tags to the ones we use
 diff_codes = {'=': 0, '-': -1, '+': 1}
 
+REPEATED_BLANK_LINES = re.compile(r'([^\S\n]*\n\s*){2,}')
+
 def compare_length(a_body, b_body):
     "Compute difference in response body lengths. (Does not compare contents.)"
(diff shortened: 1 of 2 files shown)


danielballan:
Deployed.

Mr0grog:
Re: the HTML whitespace algorithm; #101 might be the better overall approach. I’ve been talking about that or various pieces of it for a while, so I finally got it written down.

PR #103, opened 2017-10-07
Use canonical beautifulsoup4 package.

danielballan:
See https://pypi.python.org/pypi/bs4 for explanation. The 'bs4'
package is a placeholder to prevent namesquatting, and its version
is not meaningful.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-bs4
+beautifulsoup4
 diff_match_patch_python
 docopt
 lxml


danielballan:
Merging as noncontroversial. :- )

Mr0grog:
Idunno, using reliable, versioned references for things? Really not sure about this one. :P

danielballan:
Must you always make things so complicated, @Mr0grog???? 🙃

janakrajchadha:
I think we'll have to replace `bs4` with `beautifulsoup4` in all the scripts where we're using it. Correct?
Or has that been done already?

danielballan:
Actually, no, `bs4` is the still the correct thing to import. Confusingly, a package's name on PyPI does not have to match the name of the thing you import. BeautifulSoup maintains a `bs4` placeholder to ameliorate this confusion.

janakrajchadha:
That's an interesting thing to know.
The things which dazzle my mind and those which make no sense to me appear to stay in equilibrium. 🤷‍♂️

PR #108, opened 2017-10-08
Compare html_diff_render, htmldiffer, and htmltreediff with more example data

danielballan:
* [x] Move the HTML test cases hard-coded as strings in `test_differs.py` into separate HTML files so that they are easier to read. (For very short files, this might seem like a step backward in ease of use, but for more complex examples I would argue it's a step forward.)
* [x] Add several new HTML test cases, focusing on edge cases that break various diffing methods under consideration.
* [x] Apply the tests to our internal `html_diff_render` function (based on ` lxml.html.diff.htmldiff`).
* [x] Apply the tests to [htmldiffer](https://github.com/anastasia/htmldiffer).
* [x] Apply the tests to [htmltreediff](https://github.com/danielballan/htmltreediff/tree/customize) -- forked by me to add py3 compatibility and two minor tweaks.
* [x] Add some realistic (large) test files.
* [ ] Time the performance of the various methods.

Finally -- as noted in a code comment -- these tests do not assert any expected result. They can PASS or ERROR but not FAIL. For now, their purpose is to produce output that's easy to compare and discuss. In the future we might decide on an 'expected' result and assert equality with it, but at this stage I don't think we know exactly what the correct output _is_, and so enforcing stability via tests would just be a form of technical debt.

What do you think of these test cases, @janakrajchadha and @anastasia? What others should we add? I pasted a copy of the test output in [this gist](https://gist.github.com/danielballan/f49897e7aabadc9f0576c61051453ca5).


janakrajchadha:
The tests look good to me and I think we can really see the difference in results. 
I can't think of any specific additions but maybe we should think of some more cases which are likely to break the differ (some of which have been shared by @Mr0grog in the issues).

danielballan:
@janakrajchadha Great. Yes, I think we should think of more cases which are likely to break the differ. I think I incorporated all of the cases @Mr0grog shared. Can you dream up any others?

danielballan:
@Mr0grog I think this is good enough to merge and iterate on (and thread through the UI for additional evaluation).

setup.py
```diff
@@ -13,17 +14,24 @@
...
-    with open(os.path.join(os.path.dirname(__file__), fname)) as f:
+    with open(Path(__file__).parent / fname) as f:
         result = f.read()
     return result
 
 
+# Requirements not on PyPI can't be installed through `install_requires`.
+# They have to be installed manually or with `pip install -r requirements.txt`.
+requirements = [r for r in read('requirements.txt').splitlines()
+                if not r.startswith('git+https://')]
```
Mr0grog: Sorry for being late to the party on this PR. Do we need to amend the README to `pip install ...` instead of/in addition to `python setup.py develop`?

Mr0grog:
> these tests do not assert any expected result. They can PASS or ERROR but not FAIL.

I just want to say 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #109, opened 2017-10-08
Simplify config scheme by removing a layer of indirection.

danielballan:
The purpose of having a 'config' is to make it easy to plug in diffing functions
from external libraries that might be maintained separately from the library where
the server lives. This PR simplifies things by moving that config from a JSON
file to a module global.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -3,12 +3,24 @@ from docopt import docopt
 import hashlib
 from importlib import import_module
 import inspect
-import json
 import tornado.gen
 import tornado.httpclient
 import tornado.ioloop
 import tornado.web
 import web_monitoring
+
+
+# MAP tokens in the REST API to functions in modules.
+# The modules do not have to be part of the web_monitoring package.
+CONFIG = {
+    "length": ["web_monitoring.differs", "compare_length"],
+    "identical_bytes": ["web_monitoring.differs", "identical_bytes"],
+    "pagefreezer": ["web_monitoring.differs", "pagefreezer"],
+    "side_by_side_text": ["web_monitoring.differs", "side_by_side_text"],
+    "html_text_diff": ["web_monitoring.differs", "html_text_diff"],
+    "html_source_diff": ["web_monitoring.differs", "html_source_diff"],
+    "html_visual_diff": ["web_monitoring.differs", "html_diff_render"]
+}
 
 
 def load_config(config):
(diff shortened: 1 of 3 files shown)


web_monitoring/diffing_server.py
```diff
@@ -3,14 +3,26 @@
...
 import tornado.gen
 import tornado.httpclient
 import tornado.ioloop
 import tornado.web
 import web_monitoring
 
 
+# MAP tokens in the REST API to functions in modules.
+# The modules do not have to be part of the web_monitoring package.
+CONFIG = {
```
Mr0grog: I feel like a different name might be clearer at this point, e.g. `DIFF_ROUTES` or `DIFF_ENDPOINTS` or something, this this is really just a list of URL paths and doesn’t hold any other sorts of things one might think of as configuration.

web_monitoring/diffing_server.py
```diff
@@ -3,14 +3,26 @@
...
+    "length": ["web_monitoring.differs", "compare_length"],
+    "identical_bytes": ["web_monitoring.differs", "identical_bytes"],
+    "pagefreezer": ["web_monitoring.differs", "pagefreezer"],
+    "side_by_side_text": ["web_monitoring.differs", "side_by_side_text"],
+    "html_text_diff": ["web_monitoring.differs", "html_text_diff"],
+    "html_source_diff": ["web_monitoring.differs", "html_source_diff"],
+    "html_visual_diff": ["web_monitoring.differs", "html_diff_render"]
+}
```
Mr0grog: Any reason not to dump the strings and just make the values actual references here?

```py
{
  "length": web_monitoring.differs.compare_length,
  # ...etc...
}
```

If this is Python code, that seems both more convenient (can remove `load_config()`) and exposes our dependencies to anything that can analyze Python.

web_monitoring/diffing_server.py
```diff
@@ -132,18 +144,27 @@ def caller(func, a, b, **query_params):
                                "".format(func.__name__, name))
     return func(**kwargs)
 
+class IndexHandler(tornado.web.RequestHandler):
+
+    @tornado.gen.coroutine
+    def get(self):
+        # Return a list of the differs.
+        # TODO Show swagger API instead.
+        self.write(repr(list(CONFIG)))
```
Mr0grog: I don’t think anything needs to change here, but this reminds me we should sit down and figure out whether this server should be publicly accessible (in which case we need to take time to work on simple security; e.g. the differ shouldn’t accept any old arbitrary URLs to diff). If not public, the utility of rendering Swagger docs here like we do for the DB API is pretty limited and documenting routes in a more readable text format, like Markdown, might be better.

danielballan: For v0, I'm pretty we should not worry about making it publicly accessible. I'd be happier with Markdown myself but willing to do swagger for consistency if we think it will eventually become public.

web_monitoring/diffing_server.py
danielballan: Yeah, I guess the utility of this whole scheme is gone now that we're not using a config file.

web_monitoring/diffing_server.py
Mr0grog: Ok, let’s make sure we talk about this at the Wednesday meeting or sooner. Either way, I don’t think it needs to affect this PR.

Mr0grog:
👍 :shipit:

PR #113, opened 2017-10-09
Uuids from versionista links

janakrajchadha:
@danielballan


danielballan:
I think that this works, but it's better to pass query parameters -- the stuff after the `?` -- to `requests.get`'s `params` argument, rather than include them into the URL string. See #114 for some examples of this. I think we should not merge this, but I tried to fold the spirit of it (in addition to your actual commits from prior work) in #114.

janakrajchadha:
I'm working on those changes. While we are here, should I take the liberty to split `get_version_uri` into two functions as you had suggested or should I keep that a part of #85?

danielballan:
A keyword argument that modifies the return type (single result vs tuple of results) is somewhat of an anti-pattern in Python. What if we instead had a function like:

```python
def get_previous(version_id):
    """
    Get the Version immediately preceding this one.

    Parameters
    -----------
    version_id : string

    Returns
    -------
    version : dict
    """
    ...
```

I could also imagine:

```python
def get_first_after(version_id, date)
    """
    Get the first Version after a given date.

    Parameters
    ---------
    version_id : string
    date : datetime

    Returns
    -------
    version : dict
    """
```

Either or both of these might accept an optional `same_source` flag if we want to restrict the search to versions with the same `source_type`.

Can you work on these on top of #114? It might be better to wait a day or two to start. I'm going to make some major changes in response to Rob's review.

janakrajchadha:
> Can you work on these on top of #114?

Absolutely! I've already pulled the work in #114 and I'm using the patterns you had suggested.
I'll take the next couple of days to understand the design pattern in 114 a little more. I'll finish and push changes I'm working on after you've made the planned changes.

Mr0grog:
Was this fully rolled into #114? Can we close it?

janakrajchadha:
There are two parallel issues discussed here. `get_version_uri` was supposed to be split up into two different functions and that has been handled by #114. I can't see a function which gets the uuid, have you added it @danielballan?

web_monitoring/db.py
```diff
@@ -126,7 +126,7 @@ class method :meth:`Client.from_env`.
     url : string, optional
         Default is ``https://api.monitoring.envirodatagov.org``.
     """
-    def __init__(self, email, password, url=DEFAULT_URL):
+    def __init__(self, email='', password='', url=DEFAULT_URL):
```
janakrajchadha: @danielballan In case the user wants to use `from_env` later, there shouldn't be an error message when the user tries to create an object like this `obj1 = Client()'.

janakrajchadha:
Test failing here due to `VCR` overwrite issue. I'll read the docs and fix this.
@danielballan A lot of new work from you this week, need time to catch up. 😅

web_monitoring/db.py
Mr0grog: I don’t think that’s right—`from_env` is a class method, so you would never use it “later;” it always creates a fresh instance of `Client`.

> there shouldn't be an error message when the user tries to create an object like this `obj1 = Client()`

Credentials should *always* be provided to any `Client` instance. You can’t access the DB without them (that update hasn’t been deployed yet, but will be soon). I‘m pretty sure these arguments are required on purpose.

web_monitoring/db.py
```diff
@@ -657,3 +657,29 @@ def get_version_by_versionista_id(self, versionista_id):
         # Make result look like the result of `get_version` rather than the
         # result of `list_versions`.
         return {'data': result['data'][0]}
+
+    def get_db_uuid_by_versionista_id(self, versionista_id):
```
Mr0grog: I don’t think this method makes much sense to add. What meaningful utility does it offer over just calling:

```py
client.get_version_by_versionista_id(12345)['data']['uuid']?
```

janakrajchadha: @danielballan wanted this to be added, as we had discussed on slack.

web_monitoring/db.py
janakrajchadha: @Mr0grog if credentials should always be provided, then the option to use ‘from_env’ should not be there or it should be a part of the constructor.
Shouldn’t it?

web_monitoring/db.py
Mr0grog: Ok, then at the very least, this method shouldn’t duplicate the implementation of `get_version_by_versionista_id()`. It should leverage it. Is there any reason this whole method shouldn’t just be:

```py
def get_db_uuid_by_versionista_id(self, versionista_id):
    """docstring"""
    return self.get_version_by_versionista_id(versionista_id)['data']['id']
```

?

web_monitoring/db.py
Mr0grog: I think the intent here is to not make it part of the constructor so that what you are asking for is explicit. It also makes the contract of the API more clear: you can’t misunderstand that you must provide credentials—in some form—in order to create a client.

It also leaves open the opportunity to have a client *without* credentials (we just make them optional instead of required) when we eventually relax the login requirements on the DB side. You don’t have to worry about revising the constructor to no longer fill in default values from the environment and also fixing all code other people have written not in this project that might have depended on that behavior.

janakrajchadha: Oh, I see. I suppose any user must be clear about the idea of creating a class with either the constructor or by using `from_env` if they're using this API. 
Though a class method should not be used later, it is possible to use it, and that has bothered me earlier as well.

web_monitoring/db.py
janakrajchadha: I'm not sure about the purpose of adding this method either. @danielballan Is there any specific reason behind this addition?

danielballan: I agree that we don't really need this now. The day before I dove into #114, as I was still catching on the state of things, I was asking whether we already had an easy way to translate from a Versionista `version_id` to our Version UUID. Now, having finished #114, I think `get_version_by_versionista_id` is sufficiently convenient.

In [my comment above](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/113#issuecomment-335478685) I meant to redirect your effort to other convenience methods that might add more utility than this one.

Sorry for creating confusion here. I was still catching up!

web_monitoring/db.py
danielballan: > It also makes the contract of the API more clear: you can’t misunderstand that you must provide credentials—in some form—in order to create a client.

Exactly.

> Though a class method should not be used later, it is possible to use it, and that has bothered me earlier as well.

That's a fair criticism of this pattern: the fact that it's possible to do:

```python
new_client_instance = client_instance.from_env()
```

is a bit weird. Nonetheless, I find that using class methods to implement "alternative" constructors is a fairly popular pattern. An alternative approach would be to implement this as a separate function, like `make_client_from_env()`.

janakrajchadha: It is quite popular. I suppose we can add a little something about class methods in the docstring.

web_monitoring/db.py
janakrajchadha: Yeah, the db API hadn't been updated in a while. I guess it's best to add other convenience functions. @danielballan  do you have any in mind, other than the ones you've already mentioned?

danielballan: On second thought, `get_first_version_after(page_id, date)` is just `get_versions(page_id=page_id, start_date=date, chunk_size=1)['data'][0]` which doesn't seem worth a separate function. But `get_previous` still seems useful. Nothing else occurs to me yet.

Mr0grog: Hmmmm, that query would still have versions in descending date order, though, right? So it would be a little more complex (but the solution to that is probably to make order a parameter in the REST API rather than make a complex algorithm here).

`get_previous(version_id)` would just be:

```py
previous_url = get_version(version_id)['links']['previous']
return previous_url and requests.get(previous_url).json or None

# --- or ---
previous_url = get_version(version_id)['links']['previous']
return previous_url and get_version(previous_url.rsplit('/', 1)[1]) or None
```

So maybe complex enough for a method, maybe not.

Mr0grog: FYI, edgi-govdata-archiving/web-monitoring-db#164 and edgi-govdata-archiving/web-monitoring-db#165

Mr0grog: …and it looks like someone showed up out of the blue and is already working on those. So yeah, that’ll turn a bunch more things into one-liners, I guess.

janakrajchadha: @Mr0grog I'm not sure if returning both the previous url and its version content is necessary.
The response of the previous url will have its version id and it can be used to create its url (though I don't see why it is needed because we already have the response).
Also, it is really a matter of usefulness, I suppose. How often would one try to get the previous or the next version? If it'll be used quite frequently, we should definitely add this method (even if it is a one-liner).

danielballan: Yeah, I agree we don't really know yet how frequently certain patterns will be used. I feel confident that the current master branch provides a decent API (not to say optimal but decent) to the db. At this moment, until we know more, it's probably more urgent and productive to focus on the differs. But if you feel inspired to embellish this API, that would be fine too.

danielballan: Yeah, I agree we don't really know yet how frequently certain patterns will be used. I feel confident that the current master branch provides a decent API (not to say optimal but decent) to the db. At this moment, until we know more, it's probably more urgent and productive to focus on the differs. But if you feel inspired to embellish this API, that would be fine too.

danielballan:
Echoing my last inline comment, I'm going to close this:

> Yeah, I agree we don't really know yet how frequently certain patterns will be used. I feel confident that the current master branch provides a decent API (not to say optimal but decent) to the db. At this moment, until we know more, it's probably more urgent and productive to focus on the differs. But if you feel inspired to embellish this API, that would be fine too.

PR #114, opened 2017-10-10
Complete Python API to web-monitoring-db

danielballan:
This PR fills out our Python API to web-monitoring-db's REST API.
It should be fairly easy for @trinberg, @ambergman, and other Python-savvy
analysts or DataKind collaborators to use these functions whip up ad hoc
analysis in a Jupyter notebook.  Each function has numpy-style docstring, which
we can soon publish in HTML documentation using sphinx.

This cherry-picks relevant commits from #85 and builds on them.


web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+        number of items per chunk
+    start_date : datetime, optional
+    end_date : datetime, optional
+    source_type : string, optional
+    version_hash : string, optional
+    source_metadata : dict, optional
+        Examples:
+
+        * ``{'version_id': 12345678}``
+        * ``{'account': 'versionista1', 'has_content': True}``
```
Mr0grog: Would it make more sense if this and the page-specific version above were the same function with an optional `page=` kwarg? There shouldn’t be any difference between the two as far as the API is concerned.

(In fact, I believe you can GET `/api/v0/versions?page_id={page_id}` and get the same thing as `/api/v0/pages/{page_id}/versions`, though the former should *probably* be discouraged.)

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    result = res.json()
+    data = result['data']
+    data['capture_time'] = parse_timestamp(data['capture_time'])
+    data['updated_at'] = parse_timestamp(data['updated_at'])
+    data['created_at'] = parse_timestamp(data['created_at'])
+    return result
+
+
+@ensure_credentials
+def post_version(page_id, uuid, capture_time, uri, version_hash,
```
Mr0grog: Since you aren’t going by HTTP verbs for lists, would `submit` or `create` or `add` be clearer here?

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    Submit one new Version.
+
+    See :func:`post_versions` for a more efficient bulk importer.
+
+    Parameters
+    ----------
+    page_id : string
+        Page to which the Version is associated
+    uuid : string
+        a new, unique Version ID (UUID4)
```
Mr0grog: Is this intentionally non-optional?

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    start_date : datetime
+    end_date : datetime
```
Mr0grog: I think these two are copy/paste errors.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    uri : string
+        URI of content (such as an S3 bucket or InternetArchive URL)
+    version_hash : string
+        SHA256 of Version content
+    source_type : string
+        such as 'versionista' or 'internetarchive'
+    source_metadata : dict, optional
+        free-form metadata blob provided by source
```
Mr0grog: `title` should also be a valid parameter here, right?

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    url = f'{api_url()}/imports'
+    res = requests.post(url, auth=auth(),
+                        headers={'Content-Type': 'application/x-json-stream'},
+                        data='\n'.join(map(json.dumps, versions)))
+    res.raise_for_status()
+    return res.json()
 
 
+@ensure_credentials
 def query_import_status(import_id):
```
Mr0grog: Seems like renaming to `get_import_status` would be more in-line with all the new functions.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    version_id : string
+
+    Returns
+    -------
+    content : bytes
+    """
+    url = f'{api_url()}/versions/{version_id}'
+    res = requests.get(url, auth=auth())
+    assert res.ok
+    content_uri = res.json()['data']['uri']
```
Mr0grog: Why not leverage `get_version()` here?

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    return requests.get(content_uri).content
```
Mr0grog: Would it make sense to return `text` or `content` based on the `Content-Type` header?

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+        False by default
+
+    Returns
+    -------
+    source_uri or (source_uri, previous_id) depending on ``get_previous``
+    """
+    if (id_type == 'db'):
+        result = get_version(version_id=version_id)
+    elif (id_type == 'source'):
+        result = get_version_by_versionista_id(versionista_id=version_id)
```
Mr0grog: “source” = “versionista” seems like a pretty brittle equivalence for the future.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+        start_str = ''
+    if end_date is not None:
+        end_str = end_date.isoformat() + 'Z'
+    else:
+        end_str = ''
+    return f'{start_str}..{end_str}'
+
+
+def parse_timestamp(capture_time):
+    return datetime.strptime(capture_time, '%Y-%m-%dT%H:%M:%S.%fZ')
```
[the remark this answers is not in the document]
Mr0grog: (Also, maybe better to actually use an ISO 8601 parser? Compatibility with the W3C note version of the spec is the only guarantee DB makes; implementation changes might change this output slightly in the future.)

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    params = {'chunk': chunk,
+              'chunk_size': chunk_size,
+              'site': site,
+              'agency': agency,
+              'url': url,
+              'title': title,
+              'include_versions': include_versions}
+    if include_versions is not None:
+        params.update({'source_type': source_type,
+                       'version_hash': version_hash,
```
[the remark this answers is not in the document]
danielballan: I hesitated to shadow the builtin but since it's not a _keyword_ we technically can shadow it inside the function scope and nothing bad will happen.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+        pagination parameter
+    chunk_size : integer, optional
+        number of items per chunk
+    start_date : datetime, optional
+    end_date : datetime, optional
+    source_metadata : dict, optional
+        Examples:
+
+        * ``{'version_id': 12345678}``
+        * ``{'account': 'versionista1', 'has_content': True}``
```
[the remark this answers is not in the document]
danielballan: Copied straight from the swagger docs. Will add those in both places.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+              'title': title,
+              'include_versions': include_versions}
+    if include_versions is not None:
+        params.update({'source_type': source_type,
+                       'version_hash': version_hash,
+                       'capture_time': time_range_string(start_date,
+                                                          end_date)})
+    url = f'{api_url()}/pages'
+    res = requests.get(url, auth=auth(), params=params)
+    res.raise_for_status()
```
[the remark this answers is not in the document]
danielballan: Which `errors` property? I don't see it in the [API docs](http://docs.python-requests.org/en/master/api/#requests.Response) or in vars:

```python
In [11]: res = requests.get('https://api.monitoring.envirodatagov.org/api/v0/error')

In [12]: vars(res)
Out[12]: 
{'_content': b'<!DOCTYPE html>\n<html>\n<head>\n  <title>The page you were looking for doesn\'t exist (404)</title>\n  <meta name="viewport" content="width=device-width,initial-scale=1">\n  <style>\n  body {\n    background-color: #EFEFEF;\n    color: #2E2F30;\n    text-align: center;\n    font-family: arial, sans-serif;\n    margin: 0;\n  }\n\n  div.dialog {\n    width: 95%;\n    max-width: 33em;\n    margin: 4em auto 0;\n  }\n\n  div.dialog > div {\n    border: 1px solid #CCC;\n    border-right-color: #999;\n    border-left-color: #999;\n    border-bottom-color: #BBB;\n    border-top: #B00100 solid 4px;\n    border-top-left-radius: 9px;\n    border-top-right-radius: 9px;\n    background-color: white;\n    padding: 7px 12% 0;\n    box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);\n  }\n\n  h1 {\n    font-size: 100%;\n    color: #730E15;\n    line-height: 1.5em;\n  }\n\n  div.dialog > p {\n    margin: 0 0 1em;\n    padding: 1em;\n    background-color: #F7F7F7;\n    border: 1px solid #CCC;\n    border-right-color: #999;\n    border-left-color: #999;\n    border-bottom-color: #999;\n    border-bottom-left-radius: 4px;\n    border-bottom-right-radius: 4px;\n    border-top-color: #DADADA;\n    color: #666;\n    box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);\n  }\n  </style>\n</head>\n\n<body>\n  <!-- This file lives in public/404.html -->\n  <div class="dialog">\n    <div>\n      <h1>The page you were looking for doesn\'t exist.</h1>\n      <p>You may have mistyped the address or the page may have moved.</p>\n    </div>\n    <p>If you are the application owner check the logs for more information.</p>\n  </div>\n</body>\n</html>\n',
 '_content_consumed': True,
 '_next': None,
 'connection': <requests.adapters.HTTPAdapter at 0x10eb6ebe0>,
 'cookies': <RequestsCookieJar[]>,
 'elapsed': datetime.timedelta(0, 0, 659360),
 'encoding': 'UTF-8',
 'headers': {'Server': 'Cowboy', 'Date': 'Tue, 10 Oct 2017 14:33:12 GMT', 'Connection': 'keep-alive', 'Content-Type': 'text/html; charset=UTF-8', 'X-Request-Id': 'fca7380c-20fb-4835-ad4e-ebd6642e56ce', 'X-Runtime': '0.002000', 'Strict-Transport-Security': 'max-age=15552000; includeSubDomains', 'Vary': 'Origin', 'Content-Length': '1564', 'Via': '1.1 vegur'},
 'history': [],
 'raw': <urllib3.response.HTTPResponse at 0x10eb633c8>,
 'reason': 'Not Found',
 'request': <PreparedRequest [GET]>,
 'status_code': 404,
 'url': 'https://api.monitoring.envirodatagov.org/api/v0/error'}

In [13]: res.raise_for_status()
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
<ipython-input-13-6ec835a6a98f> in <module>()
----> 1 res.raise_for_status()

~/miniconda3/envs/dd/lib/python3.6/site-packages/requests/models.py in raise_for_status(self)
    933 
    934         if http_error_msg:
--> 935             raise HTTPError(http_error_msg, response=self)
    936 
    937     def close(self):

HTTPError: 404 Client Error: Not Found for url: https://api.monitoring.envirodatagov.org/api/v0/error
```

web_monitoring/db.py
danielballan: @janakrajchadha What motivated your choice of the word `'source'` as opposed to, say, `'versionista'` here?

web_monitoring/db.py
danielballan: Makes sense. (Would you find HTTP verbs much more clear? My motivation here was distinguish between functions that give a list-like result for `data` as opposed to a single object.)

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: Oh, that makes sense, but I was more getting at the fact that this line doesn’t *work.* Needs to be:

```py
{
  'hash': version_hash
}
```

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: Sorry, I mean the API returns meaningful data on errors. e.g. this request: `https://api.monitoring.envirodatagov.org/api/v0/pages?capture_time=2017-07-30T20:00:00.000Z..2017-08-0`

Returns a 400 status code, but it *also* returns JSON data that is more specific and helpful:

```json
{
  "errors": [
    {
      "status": 400,
      "title": "Invalid date: '2017-08-0'"
    }
  ]
}
```

All routes in the API do this, though in a few rare cases (mostly auth, where it was harder to change the underlying system), it’ll be in the format `{"error": {"message": "string"}}`

web_monitoring/db.py
danielballan: Sure, we can generate one if `None`.

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: (Note if you are getting HTML in the response, that is because you are in development mode, where the HTML is pretty useful because it includes an interactive interpreter and formatted stack traces. Open to changing that if you don’t think it’s worth keeping.)

web_monitoring/db.py
Mr0grog: Nah, I don’t think HTTP verbs would necessarily be more clear. I generally agree that names that distinguish the kind of action taken/content returned are better.

(In that sense, the `list_*/get_*` names are great. I was mostly reacting to the fact that the name here is `post_*`, but the first line of the docstring is `Submit a…`, implying that “post” was not really the clearest verb.)

web_monitoring/db.py
Mr0grog: I guess I was more noting that the API server will already do exactly that logic for you.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+    result = res.json()
+    data = result['data']
+    # In place, replace datetime strings with datetime objects.
+    for page in data:
+        page['created_at'] = parse_timestamp(page['created_at'])
+        page['updated_at'] = parse_timestamp(page['updated_at'])
+        if 'latest' in page:
+            page['latest']['capture_time'] = parse_timestamp(
+                page['latest']['capture_time'])
+        if include_versions:
```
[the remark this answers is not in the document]
Mr0grog: Huh. I could have sworn when I reviewed this that the above `if` was not present. Oh well.

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: I guess we could also rename this in the API. I think I just felt like `hash` was just as clear and more concise.

web_monitoring/db.py
```diff
@@ -1,47 +1,608 @@
...
+
+    Returns
+    -------
+    capture_time_query : None or string
+        If None, do not query ``capture_time``.
+    """
+    if start_date is None and end_date is None:
+        return None
+    if start_date is not None:
+        start_str = start_date.isoformat() + 'Z'
```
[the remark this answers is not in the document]
danielballan: Fair. I have revised this with I believe is a rigorous solution that accepts tzaware-datetimes (and passes their datetime into the ISO format) and also accepts tzaware-datetimes (which it localizes and again passes that local timezone into ISO format).

web_monitoring/db.py
danielballan: Then now might be a good time to remove this, as we are already aggressively breaking the API. It was a useful iteration for clarifying our thoughts about what we need. I expect your experiments will prompt more convenience methods in the future.

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
+            title=title,
+            source_metadata=source_metadata)
+        url = f'{self._api_url}/pages/{page_id}/versions'
+        res = requests.post(url, auth=self._auth,
+                            headers={'Content-Type': 'application/json'},
+                            data=json.dumps(version))
+        _process_errors(res)
+        return res.json()
+
+    def add_versions(self, versions):
```
Mr0grog: There is one more param the API supports; don’t know if we want to add it here: `update`, which can be:
- `skip` (default, leaves already existing versions in place)
- `replace` (replaces existing version with the one in this import)
- `merge` (leaves old fields intact where new values are null and merges `source_metadata` objects).

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
-    import_id : integer
+    url : string
+    email : string
+    password : string
     """
-    url = WEB_MONITORING_SHOW_IMPORT_API.format(db_url=settings['db_url'],
-                                                import_id=import_id)
-    return requests.get(url,
-                        auth=(settings['db_email'], settings['db_password']))
+    def __init__(self, url, email, password):
```
Mr0grog: Should `url` have a default (`https://api.monitoring.envirodatagov.org`)? I wouldn’t expect one for `email` and `password`, but I probably would for the API host.

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
+        """
+        Instantiate a :class:`Client` by obtaining its authentication info from
+        these environment variables:
+
+            * ``WEB_MONITORING_DB_URL``
+            * ``WEB_MONITORING_DB_EMAIL``
+            * ``WEB_MONITORING_DB_PASSWORD``
+        """
+        try:
+            url = os.environ['WEB_MONITORING_DB_URL']
```
Mr0grog: Same thought about default URL.

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
 
+class WebMonitoringDbError(Exception):
+    ...
+
+
+def _process_errors(res):
+    # If the app gives us errors, raise a custom exception with those.
+    # If not, fall back on requests, which will raise an HTTPError.
+    if res.ok:
+        return
```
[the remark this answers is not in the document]
danielballan: This might be the way to go, but I don't feel 100% sure that we'll always be interested in JSON (we might, for example, be interested in proxied content), and the individual functions have more information about what they want/expect. Could refactor later if this doesn't pan out.

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
+        url = f'{self._api_url}/imports'
+        validated_versions = [_build_importable_version(**v) for v in versions]
+        res = requests.post(
+            url, auth=self._auth,
+            headers={'Content-Type': 'application/x-json-stream'},
+            data='\n'.join(map(json.dumps, validated_versions)))
+        _process_errors(res)
+        return res.json()['data']['id']
+
+    def add_versions_batched(self, versions, batch_size=1000):
```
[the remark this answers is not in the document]
danielballan: I like that the `/pages/{page_id}/versions'` endpoints lets us specify the `page_id` explicitly so I think we should expose both for now, but revisit this if we can make `/imports` and `/pages/{page_id}/versions'` more interchangeable.

web_monitoring/db.py
```diff
@@ -1,47 +1,657 @@
...
+        """
+        Get a Changes between two Versions.
+
+        Parameters
+        ----------
+        page_id : string
+        to_version_id : string
+        from_version_id : string, optional
+            If from_version_id is not given, it will be treated as version
+            immediately prior to ``to_version``.
```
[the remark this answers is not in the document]
danielballan: Sure. Seems sensible but not high prio.

web_monitoring/db.py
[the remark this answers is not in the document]
danielballan: Ignore the above: it applies to `add_version` and `add_versions`. There is no good reason for needing `add_versions` and `add_versions` batched and I agree they should be combined.

danielballan:
Thanks for the thorough and prompt review.

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: Oh, that is a very good point. :thumbsup:

web_monitoring/db.py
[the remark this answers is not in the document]
Mr0grog: Cool. Logged on DB as edgi-govdata-archiving/web-monitoring-db#157

web_monitoring/db.py
```diff
@@ -1,47 +1,658 @@
...
+            url = os.environ['WEB_MONITORING_DB_URL']
+            email = os.environ['WEB_MONITORING_DB_EMAIL']
+            password = os.environ['WEB_MONITORING_DB_PASSWORD']
+        except KeyError:
+            raise MissingCredentials("""
+Before using this method, database credentials must be set via environmental
+variables:
+
+   WEB_MONITORING_DB_URL
+   WEB_MONITORING_DB_EMAIL
```
Mr0grog: Did you mean to remove `WEB_MONITORING_DB_PASSWORD` or `WEB_MONITORING_DB_URL`? (Thinking maybe best to keep both, but specify the latter is optional.)

web_monitoring/db.py
```diff
@@ -1,47 +1,658 @@
...
+        Instantiate a :class:`Client` by obtaining its authentication info from
+        these environment variables:
+
+            * ``WEB_MONITORING_DB_URL``
+            * ``WEB_MONITORING_DB_EMAIL``
+            * ``WEB_MONITORING_DB_PASSWORD`` (optional -- defaults to
+              ``https://api.monitoring.envirodatagov.org``)
+        """
+        try:
+            url = os.environ['WEB_MONITORING_DB_URL']
```
Mr0grog: This needs to be `os.environ.get()`, otherwise it’ll raise a `KeyError` even though this is now optional.

web_monitoring/db.py
```diff
@@ -1,47 +1,658 @@
...
+            Specifies how versions that are already in the database (i.e.
+            versions with the same ``capture_time`` and ``source_type``) should
+            be handled:
+
+                * ``'skip'`` (default) -- Don’t import the version or modify
+                  the existing database entry.
+                * ``'replace'`` -- Replace the existing database entry with the
+                  imported one
+                * ``'merge'`` -- Similar to `replace`, but merges the values in
+                  ``source_metadata``
```
Mr0grog: FWIW, this will merge other fields, too, in the sense that, if they aren’t specified, they won’t be updated (unlike replace, which clears any unspecified fields in the target record). So you could use this behavior to change just the metadata or just the source type without amending the URI or hash, etc. Dunno if that matters here, though, since I think `_build_importable_version` will never create an object with unspecified fields.

titaniumbones:
super fun to read through this -- excited at the prospect of being able to do ad hoc analysis in Jupyter or similar environment -- thanks for pointing the slack channel  here & making me aware of it @danielballan !

web_monitoring/db.py
danielballan: Oops. Agreed, and fixed.

web_monitoring/cli.py
```diff
@@ -3,20 +3,30 @@
...
 # CLI. They also print. To access this functionality programmatically, it is
 # better to use the underlying library code.
 
 
+def _add_and_monitor(versions):
+    cli = db.Client.from_env()  # will raise if env vars not set
+    # Wrap verions in a progress bar.
+    versions = tqdm(versions, desc='importing', unit=' versions')
+    print('Submitting Versions to web-monitoring-db...')
+    import_ids = cli.add_versions(versions)
```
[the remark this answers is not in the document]
danielballan: Ha! I nearly did that (I changed it and changed it back) but I misconstrued that you thought `50000` was a reasonable batch size. Changed to `1000` now.

web_monitoring/db.py
danielballan: Let's leave this as is but revise the description in swagger.yml (where I cribbed this from).

web_monitoring/cli.py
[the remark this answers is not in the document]
Mr0grog: 👍 Nah, I was being dumb and getting myopically focused on the default being non-batched because the basic original implementation was non-batched.

web_monitoring/tests/test_db.py
```diff
@@ -0,0 +1,240 @@
...
+@db_vcr.use_cassette()
+def test_get_version():
+    cli = Client(**AUTH)
+    res = cli.get_version(TO_VERSION_ID)
+    assert res['data']['uuid'] == TO_VERSION_ID
+    assert res['data']['page_uuid'] == PAGE_ID
+
+
+@db_vcr.use_cassette()
+def test_get_version_by_versionista_id():
```
janakrajchadha: I'm not sure how this test passed.
If you use the API to get the data for the versionista id, you'll see that while other details match, the `uuid` and `page_uuid` are different.

janakrajchadha: I think the data in the cassette for this test is incorrect. I could be wrong, still diagnosing this issue.

danielballan: I ran this against a local deployment using the seed data, not the staging deployment. I'll add a comment to make that clear.

janakrajchadha: Should all tests have data from the same deployment?
I'm not familiar with the use of vcr, so I'm still figuring this part out.

Mr0grog: > Should all tests have data from the same deployment?

No. Data in the tests should not need to bear any relationship whatsoever to the data in any given deployment. It shouldn’t matter where it came from. We only care that the data is *valid* (not the same as production) and its *structure* matches that of the latest DB version.

Actual data from *any* DB (so long as its code is up-to-date) will satisfy the above conditions. VCR just records a request so it can be played back, meaning it will cover the above conditions without someone having to hand-code fixture data.

PR #115, opened 2017-10-11
Minor revisions to internetarchive module (explained below)

danielballan:
- `except` catches `BaseException`, including `KeyboardInterrupt`. Catch `Exception`.
- Very minor docstring style conformance
- Save a variable assignment in the hot loop by reusing `last_hashes`.
- Add parameters to docstring
- Make order consistent (does not break API because these args were/are keyword-only)

### DIFF
--- a/web_monitoring/internetarchive.py
+++ b/web_monitoring/internetarchive.py
@@ -61,6 +61,7 @@ def original_url_for_memento(memento_url):
 
     Examples
     --------
+    Extract original URL.
     >>> original_url_for_memento('http://web.archive.org/web/20170813195036/https://arpa-e.energy.gov/?q=engage/events-workshops')
     'https://arpa-e.energy.gov/?q=engage/events-workshops'
     """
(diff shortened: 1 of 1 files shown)


web_monitoring/internetarchive.py
```diff
@@ -208,17 +211,15 @@ def list_versions(url, *, from_date=None, to_date=None, skip_repeats=True):
...
     for version in search_cdx(params):
         # TODO: may want to follow redirects and resolve them in the future
         if not skip_repeats or last_hashes.get(version.url) != version.digest:
-            has_versions = True
             last_hashes[version.url] = version.digest
             # TODO: yield the whole version
             yield version
 
-    if not has_versions:
+    if not last_hashes:
```
Mr0grog: Huh, I was about to say this would never pass, but now I see empty dicts are falsy. How did I never know that was the case in Python??? 👍

Mr0grog:
Is this the style you are following? https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

danielballan:
Yes. I'm using the `numpydoc` sphinx extension to render the docstrings into HTML. I'll add that link to the contribution guide.

PR #120, opened 2017-10-17
Address post-facto review of documentation and example-data.

danielballan:


danielballan:
@Mr0grog I think I caught all your in-line suggestions here. Thanks.

PR #122, opened 2017-10-26
Clean up issues encountered doing an IA import

Mr0grog:
While doing a one-time import of a few pages for the analyst team, I encountered several issues. I'm not sure how we didn't catch these in the initial DB client work.

Note some of these are good fixes, but others are probably not right (they are more like quick hacks to get it working). This may need some refinement.

/cc @danielballan

### DIFF
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -83,8 +83,8 @@ def _build_version(*, page_id, uuid, capture_time, uri, hash, source_type, title
     return version
 
 
-def _build_importable_version(*, page_url, uuid, capture_time, uri, hash,
-                             source_type, title, source_metadata=None):
+def _build_importable_version(*, page_url, uuid=None, capture_time, uri, version_hash,
+                             source_type, page_title, site_agency, site_name, source_metadata=None):
     """
     Build a Version dict from parameters, performing some validation.
 
@@ -99,10 +99,12 @@ def _build_importable_version(*, page_url, uuid, capture_time, uri, hash,
                'uuid': uuid,
                'capture_time': capture_time,
                'uri': str(uri),
-               'hash': str(hash),
+               'hash': str(version_hash),
                'source_type': str(source_type),
-               'title': str(title),
-               'source_metadata': source_metadata}
+               'title': str(page_title),
+               'source_metadata': source_metadata,
+               'site_agency': site_agency,
+               'site_name': site_name}
     return version
 
 
@@ -405,7 +407,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         for batch in toolz.partition_all(batch_size, versions):
             # versions might be a generator. This comprehension will pull on it
             validated_versions = [_build_importable_version(**v)
-                                  for v in versions]
+                                  for v in batch]
             res = requests.post(
                 url, auth=self._auth,
                 headers={'Content-Type': 'application/x-json-stream'},
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -7,6 +7,7 @@ import tornado.httpclient
 import tornado.ioloop
 import tornado.web
 import web_monitoring
+import web_monitoring.differs
 
 
 # Map tokens in the REST API to functions in modules.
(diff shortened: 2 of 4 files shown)


web_monitoring/db.py
```diff
@@ -83,8 +83,8 @@ def _build_version(*, page_id, uuid, capture_time, uri, hash, source_type, title
     return version
 
 
-def _build_importable_version(*, page_url, uuid, capture_time, uri, hash,
-                             source_type, title, source_metadata=None):
+def _build_importable_version(*, page_url, uuid=None, capture_time, uri, version_hash,
+                             source_type, page_title, site_agency, site_name, source_metadata=None):
```
Mr0grog: This signature change (and the associated implementation changes) are probably not quite right, but they were more expedient than changing where we call `_build_importable_version(**version_from_ia_module)`. The params we expect here are not really compatible with what the `internetarchive` module spits out, so we need to figure out what to change where for that.

web_monitoring/diffing_server.py
```diff
@@ -7,6 +7,7 @@
 import tornado.ioloop
 import tornado.web
 import web_monitoring
+import web_monitoring.differs
```
Mr0grog: Is the server currently working and deployed without this??? It seems to be required for me.

web_monitoring/internetarchive.py
```diff
@@ -141,6 +143,12 @@ def search_cdx(params):
         except Exception:
             raise UnexpectedResponseFormat(text)
 
+        clean_url = REDUNDANT_HTTPS_PORT.sub(
+            r'\1\2', REDUNDANT_HTTP_PORT.sub(
+                r'\1\2', data.url))
+        if clean_url != data.url:
+            data = data._replace(url=clean_url)
```
Mr0grog: IA will frequently wind up with URLs that have redundant port information in them. Not sure what in their implementation leads to this, but it seemed sensible to clean these up.

There is a related issue of sometimes having both `http:` and `https:` versions of an otherwise identical URL and content, but I thought I’d hold off on trying to think of a fix for that since there’s much less clarity on what’s correct there.

Mr0grog:
@danielballan looks like maybe the `WEB_MONITORING_DB_STAGING_*` env vars are wrong on Circle? The tests are failing due to authentication errors. (They run fine for me locally.)

web_monitoring/diffing_server.py
danielballan: I haven't re-deployed since the commit where we switched from string-based config to literal references to the modules.

web_monitoring/db.py
```diff
@@ -405,7 +407,7 @@ def add_versions(self, versions, *, update='skip', batch_size=1000):
         for batch in toolz.partition_all(batch_size, versions):
             # versions might be a generator. This comprehension will pull on it
             validated_versions = [_build_importable_version(**v)
-                                  for v in versions]
+                                  for v in batch]
```
danielballan: !

web_monitoring/db.py
danielballan: Until some point in the middle of my work on `Client`, I lost track of the fact that bulk-import mashes attributes in with version attributes. I still find this confusing and potentially bad: who checks that the site_name is consistent? But agreed that this fix is needed at least for now and that my original was just a mistake.

danielballan:
@Mr0grog The env vars are correct. I bet you'll be able to reproduce this locally if you clear the test's cache dir:

```
rm -rf ~/.cache/web-monitoring-processing/tests/
```

This used to pass on Circle; I suspect it's related to the "auth everywhere" change on api-staging. We are authenticating this request (and all requests from the `Client`) so the problem is not yet clear.

danielballan:
Now that the tests are passing, I'm going to merge this as is in order to get the critical fixes in. I know that more changes the version "building" are coming.

janakrajchadha:
@danielballan How did the `WEB_MONITORING_DB_STAGING_*` env var issue get fixed?

Mr0grog:
@janakrajchadha the env vars were in fact not correct :P

They never caused a problem before because they never got checked on the server since we never hit any authenticated routes. Now that all routes are authenticated, they got checked and failed.

PR #126, opened 2017-11-02
Ignore test cache/output directories

Mr0grog:
These were turning into a lot of noise in my local checkout. Seems like they should be ignored.


danielballan:
Yes, it is correct to ignore these.

Issue #124, opened 2017-10-26
Reduce list-of-links issue (from CDC) to a clear test case


Mr0grog:
The LXML-based visual diff (i.e. `html_diff_render`) of this change:

https://api.monitoring.envirodatagov.org/api/v0/pages/376d7f27-fa20-448a-a5ab-0ec0e78aa1ff/changes/f96bf124-57fd-46e8-a2e4-7aeef185d4ea..db9e5f92-8d16-4038-9b88-27ce723396ed

(View in the UI at: https://monitoring.envirodatagov.org/page/376d7f27-fa20-448a-a5ab-0ec0e78aa1ff/f96bf124-57fd-46e8-a2e4-7aeef185d4ea..db9e5f92-8d16-4038-9b88-27ce723396ed)

Does a pretty poor job with the list of alphabetical links. Here’s an example snippet:

```html
<div><ul><li><a href=\"https://www.cdc.gov/az/a.html\">A <ins> Link: https://www.cdc.gov/az/a.html </ins><li><ins><a href=\"https://www.cdc.gov/az/b.html\">B Link: https://www.cdc.gov/az/b.html </a></ins></li><ins> </ins><li><ins><a href=\"https://www.cdc.gov/az/c.html\">C Link: https://www.cdc.gov/az/c.html </a></ins></li>

...

<li><ins><a href=\"https://www.cdc.gov/az/0.html\"># Link: https://www.cdc.gov/az/0.html </a></ins></li> <del> Link: //www.cdc.gov/az/a.html </del><li><del><a href=\"//www.cdc.gov/az/b.html\">B Link: //www.cdc.gov/az/b.html </a></del></li><del> </del><li><del><a href=\"//www.cdc.gov/az/c.html\">C Link: //www.cdc.gov/az/c.html </a></del></li>
...
```

Note how the first `<li>` doesn’t get closed properly… but actually, the first deletion (after all the additions), which is from that same _first_ `<li>`, happens outside an `<li>` entirely and is followed by  the close tag for the first `<li>` in the list.

This list-of-links/open-and-close-tag-nesting issue seems to be a fairly common pattern and problem.

We need to reduce this to a clearer and simpler test case to run against all our differs.

weatherpattern:
@Mr0grog i'm working on this ticket and create a general test case for this.
Question: where (or which dates) did you get your example snippet? When I inspect the source, I don't see any changes to the a-z menu.

[...]

Issue #94, opened 2017-09-25
html_visual_diff transforms quotes, etc. to entities in style/script elements


Mr0grog:
Was trying to figure out what was going on when testing the visual diff and seeing views like this:

![screen shot 2017-09-25 at 2 12 25 pm](https://user-images.githubusercontent.com/74178/30831323-d9335f3a-a1fb-11e7-85ad-a16c8310cc56.png)

Turns out one of the issues (they are myriad) is that the differ is treating the contents of `<script>` and `<style>` elements as HTML, transforming things like quotes into entities (e.g. `&quo;`).

We really need to take the time to look at @anastasia’s [HTML differ](https://github.com/anastasia/htmldiffer) and see if we should switch to it. I have a feeling it likely does not suffer from this issue.

Mr0grog:
(Minor update here: https://github.com/anastasia/htmldiffer does correctly handle the entity issue, but has a couple other bugs that can trigger `combined_diff` to output bad HTML, so there are definitely some things that require being careful there.)

danielballan:
Seems reasonable to deploy them both while we're in the beta phase, to collect more anecdotes about corner cases. I'll get on that...

janakrajchadha:
@Mr0grog What other bugs in @anastasia's code can result in bad HTML?

Mr0grog:
Bugs are filed on its repo.

anastasia:
(heads up: https://github.com/anastasia/htmldiffer/issues?q=is%3Aissue+is%3Aclosed)

danielballan:
Thanks for the update @anastasia. I encoded the issues that Rob found and some others in #108.

PR #127, opened 2017-11-02
WIP: Visual diff workarounds/hacks/fixes

Mr0grog:
**Work in Progress; Do Not Merge**

These are a few quick-and-dirty fixes and hacks for the `html_diff_render` visual diff so that it works better in real-world scenarios where we hit lots of problems. This will be a little bit of a grab-bag, but I intend to keep this relatively small so it can be merged by the end of the week and then iterated on.

- [x] Handle `<script>` and `<style>` elements well (#94 plus more)
- [x] Don’t diff stuff in the `<head>` (except may be `<title>`); just use the new version (maybe keep old stuff in a `<template>`, which is inert).
- [x] Remove the “Link: [url]” text after changed links (ideally, still include it as title text or something a script can turn into useful display)
- [x] See if there’s an obvious fix for #124 (list of links). Only address here if it’s easy.
- [ ] Look into character encoding issues @trinberg mentioned.
- [x] ¿Split this differ and all its associated support functions into a separate module? (There is a lot of code involved in cleaning up and working around its issues.)

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -50,7 +50,7 @@ def html_diff_render(a_text, b_text):
         soup_new.html.insert(0, head)
 
     # htmldiff will unfortunately try to diff the content of elements like
-    # <script> or <style> that embed foreign cnontent that shouldn't be parsed
+    # <script> or <style> that embed foreign content that shouldn't be parsed
     # as part of the DOM. We work around this by replacing those elements
     # with placeholders, but a better upstream fix would be to have
     # `flatten_el()` handle these cases by creating a special token, e.g:
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -139,8 +139,17 @@ def _add_undiffable_content(soup, replacements):
     from the replacements dict.
     """
     for element in soup.select('[wm-diff-replacement]'):
-        replacement = replacements[element['wm-diff-replacement']]
+        replacement_id = element['wm-diff-replacement']
+        replacement = replacements[replacement_id]
         if replacement:
+            if replacement_id.startswith('old-'):
+                replacement['class'] = 'wm-diff-deleted-active'
+                wrapper = soup.new_tag('template')
+                wrapper['class'] = 'wm-diff-deleted-inert'
+                wrapper.append(replacement)
+                replacement = wrapper
+            else:
+                replacement['class'] = 'wm-diff-inserted-active'
             element.replace_with(replacement)
 
     return soup
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -1,0 +1,209 @@
+from bs4 import BeautifulSoup, Comment
+import copy
+from lxml.html.diff import (tokenize, htmldiff_tokens, fixup_ins_del_tags,
+                            href_token)
+from .differs import compute_dmp_diff
+
+
+def html_diff_render(a_text, b_text):
+    """
+    HTML Diff for rendering. This is focused on visually highlighting portions
+    of a page’s text that have been changed. It does not do much to show how
+    node types or attributes have been modified (save for link or image URLs).
+
+    The overall page returned primarily represents the structure of the "new"
+    or "B" version. However, it contains some useful metadata in the `<head>`:
+
+    1. A `<template id="wm-diff-old-head">` contains the contents of the "old"
+       or "A" version’s `<head>`.
+    2. A `<style id="wm-diff-style">` contains styling diff-specific styling.
+    3. A `<meta name="wm-diff-title" content="[diff]">` contains a renderable
+       HTML diff of the page’s `<title>`. For example:
+
+        `The <del>old</del><ins>new</ins> title`
+
+    NOTE: you may want to be careful with rendering this response as-is;
+    inline `<script>` and `<style>` elements may be included twice if they had
+    changes, which could have undesirable runtime effects.
+
+    Example
+    -------
+    text1 = '<!DOCTYPE html><html><head></head><body><p>Paragraph</p></body></html>'
+    text2 = '<!DOCTYPE html><html><head></head><body><h1>Header</h1></body></html>'
+    test_diff_render = html_diff_render(text1,text2)
+    """
+    soup_old = BeautifulSoup(a_text, 'lxml')
+    soup_new = BeautifulSoup(b_text, 'lxml')
+
+    # Remove comment nodes since they generally don't affect display.
+    # NOTE: This could affect display if the removed are conditional comments,
+    # but it's unclear how we'd meaningfully visualize those anyway.
+    [element.extract() for element in
+     soup_old.find_all(string=lambda text:isinstance(text, Comment))]
+    [element.extract() for element in
+     soup_new.find_all(string=lambda text:isinstance(text, Comment))]
+
+    # Ensure the new soup (which we will modify and return) has a `<head>`
+    if not soup_new.head:
+        head = soup_new.new_tag('head')
+        soup_new.html.insert(0, head)
+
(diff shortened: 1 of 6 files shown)


web_monitoring/tests/test_differs.py
```diff
@@ -60,3 +64,55 @@ def test_get_visible_text():
     html = '<!--First comment--><h1>First Heading</h1><p>First paragraph.</p>'
     actual = wd._get_visible_text(html)
     assert actual == 'First Heading First paragraph.'
+
+
+# TODO: extend this to other html differs
```
danielballan: It makes sense to me to keep the tests that generate output for humans in a separate file from the fully automated ones that can FAIL. Maybe we can rename them later so this is more clear.

Mr0grog:
OK, this is mostly there. I need to take a little time to look into the last two issues, but we’ve said they are both speculative and aren’t worth a *huge* amount more work to solve here if that’s what they require.

I have moved all of `html_diff_render` into its own module. The routine can now be gotten from `web_monitoring.html_diff_render.html_diff_render`. I’m not in love with the repetitiveness there and am totally open to different naming, though this isn’t uncommon in Python.

There are a bunch of changes to the structure of diffs in a7367712326073e7c995ba028c04c49ac85e1295 that should help clean things up and get better renderable results that are easier for both humans and machines to parse. Excerpted from the commit description:

Diff only the `<body>`, but add some structured metadata to the head. After learning a lot more about LXML’s htmldiff, it doesn’t seem relevant to try and diff the content of the `<head>` when htmldiff doesn't really handle structure correctly and doesn’t actually “see” changes to elements themselves (only their content). There’s nothing in the `<head>` for it to do much work with (unless we made deep, extensive modifications) and we have no real way to meaningfully visualize it, either. So just don’t bother.

However, to support UIs that want to split out the removals and additions to view them side-by-side, this adds some helpful metadata:

1. The diff styling now has an `id` so other programs can (re)move or change it (it is `wm-diff-style`)

2. A diff of the page title is stored as a `<meta name="wm-diff-title" content="[diff]">`, where the content is HTML (e.g. `The <del>old</del><ins>new</ins> title`)

3. The “old” or “A” version’s `<head>` is now stored in `<template id="wm-diff-old-head">` in the `<head>`. We used to use `<ins>/<del>` elements, but those aren’t supported inside `<head>` and caused issues. This simplifies things because `<template>` is inert, so you don’t have to worry about scripts and styles coming from the old version. If you need to render the “old” version, just replace the current `<head>` with the contents of the template.

**NOTE: I’ve got some changes to the UI to better handle the improved diffs here that I need to push up as a PR over there.**

---

If you have time to pre-review this while I poke at the last two potential issues (list of links and encoding), @danielballan, that’d be lovely.

Mr0grog:
…and it looks like the general cleanup and rigor in parsing here may have unintentionally resolved #124 for us! Yay :)

Mr0grog:
Fixes #124.

Mr0grog:
Anyway, I also pushed up https://github.com/edgi-govdata-archiving/web-monitoring-processing/commit/89d334ee34f7ad435b0f34c6589e1239c1c4cbc8 to separate branch for safe-keeping. It’s also nutty enough that I might not remember how it works and what works and doesn’t work about it in the morning. But it’s there for future me or others to try and clean up (the code is truly utterly horrifying).

web_monitoring/html_diff_render.py
```diff
@@ -0,0 +1,209 @@
...
+    old_tokens = tokenize(old)
+    new_tokens = tokenize(new)
+    old_tokens = [_customize_token(token) for token in old_tokens]
+    new_tokens = [_customize_token(token) for token in new_tokens]
+    result = htmldiff_tokens(old_tokens, new_tokens)
+    result = ''.join(result).strip()
+    return fixup_ins_del_tags(result)
+
+
+class MinimalHrefToken(href_token):
```
danielballan: :+1:

web_monitoring/html_diff_render.py
```diff
@@ -0,0 +1,209 @@
...
+    return soup
+
+
+def _get_title(soup):
+    return soup.title and soup.title.string or ''
+
+
+def _diff_title(old, new):
+    return ''.join(map(
+        lambda change: ((change[0] == -1 and '<del>{}</del>')
```
danielballan: This is a clever way to implement a switch. I like it, but it took me a minute to parse it. A docstring and/or comment to help future readers might be a good idea.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -0,0 +1,63 @@
+"""
+This module hosts tests for visual HTML diffs that check validity -- that is,
+they focus on whether the diff will render properly or outputs correct/working
+markup. Unlike the tests in `test-html_diff.py`, these tests can actually fail
```
danielballan: inconsequential typo: `-` -> `_`

web_monitoring/html_diff_render.py
```diff
@@ -0,0 +1,209 @@
...
+    [element.extract() for element in
+     soup_new.find_all(string=lambda text:isinstance(text, Comment))]
+
+    # Ensure the new soup (which we will modify and return) has a `<head>`
+    if not soup_new.head:
+        head = soup_new.new_tag('head')
+        soup_new.html.insert(0, head)
+
+    # htmldiff will unfortunately try to diff the content of elements like
+    # <script> or <style> that embed foreign cnontent that shouldn't be parsed
```
lightandluck: typo: cnontent -> content

Mr0grog:
> I didn't expect that dropping down just one layer into lxml.html.diff would make it possible to customize the output so well.

Me, too! On the flip side, though, I think 89d334ee34f7ad435b0f34c6589e1239c1c4cbc8 shows that if you drop down one *more* level, things become more tightly coupled and harder to manipulate pretty quickly.

> I wonder how deeply embedded the choice of 'ins' and 'del' is, and whether it's possible to choose custom tags without changing the C extensions.

It’s hard-coded, but it’s all in the Python layer. In fact, almost everything about this diffing algorithm that matters in pure Python—it’s ultimately a wrapper around the built-in `difflib.SequenceMatcher`. 

So! Very changeable with only Python code, but doing so would mean pretty much forking the whole thing (which might be a good idea anyway).

web_monitoring/html_diff_render.py
```diff
@@ -149,12 +150,24 @@ def _get_title(soup):
     return soup.title and soup.title.string or ''
 
 
+def _html_for_dmp_operation(operation):
```
danielballan: 👍

danielballan:
> So! Very changeable with only Python code, but doing so would mean pretty much forking the whole thing (which might be a good idea anyway).

:+1:

Revisions look good. Merge and iterate from here?

Mr0grog:
OK! I wanted to get that last commit about making removed scripts/styles inert in, but also get this related PR on UI up: https://github.com/edgi-govdata-archiving/web-monitoring-ui/pull/152

It should probably be merged before this.

Otherwise, I’m cool merging this now. I’ll make a second PR for a cleaned-up version of the crazy deep changes (they’re definitely necessary to really move this diff to where it needs to be).

web_monitoring/html_diff_render.py
```diff
@@ -0,0 +1,231 @@
...
+    HTML Diff for rendering. This is focused on visually highlighting portions
+    of a page’s text that have been changed. It does not do much to show how
+    node types or attributes have been modified (save for link or image URLs).
+
+    The overall page returned primarily represents the structure of the "new"
+    or "B" version. However, it contains some useful metadata in the `<head>`:
+
+    1. A `<template id="wm-diff-old-head">` contains the contents of the "old"
+       or "A" version’s `<head>`.
+    2. A `<style id="wm-diff-style">` contains styling diff-specific styling.
```
lightandluck: I think you can get rid of the first `styling` here

PR #128, opened 2017-11-07
Limit coverage reporting to our own code

Mr0grog:
Checking coverage locally or in CI gets kinda crazy because our coverage reports are including all our dependencies. This is just a simple switch to limit coverage to code in the `web_monitoring` directory. Not sure if I should also be excluding `web_monitoring/tests`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+source =
+    web_monitoring
+[report]
+omit =
+    */python?.?/*
+    */site-packages/nose/*
+    # ignore _version.py and versioneer.py
+    .*version.*
+    *_version.py
+
+exclude_lines =
+    if __name__ == '__main__':
--- a/circle.yml
+++ b/circle.yml
@@ -12,7 +12,7 @@ dependencies:
 
 test:
   override:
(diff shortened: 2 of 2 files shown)


Mr0grog:
Hmmm, maybe also exclude `web_monitoring/_version.py`, since it’s autogenerated and [hopefully] not really germane to our functionality?

danielballan:
I piled on a commit to skip versionista's stuff.

>  Not sure if I should also be excluding `web_monitoring/tests`.

To summarize versions of this argument I've had in the past, the purist position is ,"tests shouldn't count toward coverage" and the pragmatist position is, "it's useful to verify that your tests are actually being run". I lean with the pragmatists.

Mr0grog:
Ooooh, did not know about `exclude_lines`. That’s neat.

Do we still need the line I added to the Circle script now that you added the config file?

> To summarize versions of this argument I've had in the past, the purist position is ,"tests shouldn't count toward coverage" and the pragmatist position is, "it's useful to verify that your tests are actually being run". I lean with the pragmatists.

:thumbsup:

Mr0grog:
Removed no-longer-relevant CLI `--source` option; merging.

PR #131, opened 2017-12-04
Add basic version of links diff

### LINKED ISSUE #102: New differ: list the URLs of the links before and after
See https://github.com/edgi-govdata-archiving/web-monitoring-ui/issues/113

We want something like

```python
def diff_links(a_html, b_html):
    ...
    # return something like [(url1, url2, url3), (url4, utl5)]
```

Mr0grog:
Nobody has moved forward on this, so I went ahead and did a basic implementation. There’s probably plenty to quibble with here, but this gets us something working.

Fixes #102.

screen shot 2017-12-04 at 12 05 27 am

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -9,6 +9,7 @@ import tornado.web
 import web_monitoring
 import web_monitoring.differs
 import web_monitoring.html_diff_render
+import web_monitoring.links_diff
 
 
 # Map tokens in the REST API to functions in modules.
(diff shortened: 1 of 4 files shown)


Mr0grog:
Was about to merge when I realized this left the `links_diff` module at 98% coverage, so I made a slight tweak (just to the tests) to move that up to 100%.

LMK if this still looks good. Otherwise I’ll merge in late afternoon (Pacific), since the change is pretty minor.

Issue #132, opened 2017-12-08
Clean up differ names and responses


danielballan:
* Revisit the names of the diff services and try to make them more clear.
* Make routes in -db and routes in -processing consistent.
* Include `type` in all differ responses.

PR #133, opened 2017-12-08
Add new route names aiming to be more clear.

danielballan:
The old names are maintained for back-compat for now.


danielballan:
Addresses part of #132

PR #135, opened 2017-12-10
added list of link tests

weatherpattern:
Created simple test cases for list of links described in #124 

Added cases for changes in text and urls, and adding and remove list items.

### DIFF
--- a/web_monitoring/example_data/add-list.after
+++ b/web_monitoring/example_data/add-list.after
@@ -1,0 +1,11 @@
+<html>
+  <head>
+  </head>
+  <body>
+    <ul>
+    	<li><a href="/a">Link a</a></li>
+    	<li><a href="/b">Link b</a></li>
+    	<li><a href="/c">Link c</a></li>
+    </ul>
+  </body>
+<html>
--- a/web_monitoring/example_data/add-list.before
+++ b/web_monitoring/example_data/add-list.before
@@ -1,0 +1,10 @@
+<html>
+  <head>
+  </head>
+  <body>
+    <ul>
+    	<li><a href="/a">Link a</a></li>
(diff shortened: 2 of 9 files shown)


danielballan:
Ditto -- sorry for taking too long to look at this. Yep, gotta add these files to the list of cases, and then this should be good to merge.

weatherpattern:
Sorry to have jumped the gun on this. When I run run_test.py, test_html_diff.py passes but I get 4 failures and a few others from other tests. I am having trouble understanding the error output. Could we book some time to go over it?
Test output is attached.

[run_test_output.txt](https://github.com/edgi-govdata-archiving/web-monitoring-processing/files/1557067/run_test_output.txt)

danielballan:
The Circle-CI logs show that the tests are now being run and that they pass, so this is good to merge. It looks like you have a local dev issue. Are you setting environment variables with a valid web-monitoring-db username and password? See `.env.example` for a template. I'm happy to provide follow-up support over Slack. It's possible our documentation on this is not complete or clear.

Mr0grog:
Heh, we actually solved this at the start of the meeting, @danielballan. It was a problem with CircleCI settings and the fact that @weatherpattern is developing on his own fork (so env vars were not shared).

I set it to share vars temporarily (back off now), but that’s not really safe. I wish we could create a list of privileged forks in Circle, but it doesn’t seem like you can. I guess the solutions are:

1. Work on a branch in this repo instead of a fork
2. Set up CircleCI specifically for your fork with the right env vars (@weatherpattern was worried this might be complicated, but I don’t think it should be too bad—the hard part is just knowing the right vars and values to set)

weatherpattern:
For now I'm happy to work with on a work on a branch in this repo.
I'll put setting up circleci on my to do list.

PR #136, opened 2017-12-12
Add 'type' and 'change_count' (where applicable and easy) to response.

danielballan:


danielballan:
A handful of tests still need to be updated.

web_monitoring/diffing_server.py
```diff
@@ -81,8 +81,17 @@ def get(self, differ):
 
         # Pass the bytes and any remaining args to the diffing function.
         executor = concurrent.futures.ProcessPoolExecutor()
-        res = yield executor.submit(caller, func, res_a, res_b, **query_params)
-        self.write({'diff': res, 'version': web_monitoring.__version__})
+        count, diff = yield executor.submit(caller, func, res_a, res_b,
+                                            **query_params)
```
Mr0grog: So my biggest concern here is: should this be a dict instead of a tuple?

I wonder if we might want to add other things in the future (either standard across all diffs like `change_count` here or something particular to one kind of diff) and it seems to me that returning tuples this way, while it makes for nicer syntax, would make those kinds of future changes harder.

web_monitoring/html_diff_render.py
```diff
@@ -102,7 +102,7 @@ def html_diff_render(a_text, b_text):
     replacements_new.update(replacements_old)
     soup_new = _add_undiffable_content(soup_new, replacements_new)
 
-    return soup_new.prettify(formatter='minimal')
+    return len(replacements_new), soup_new.prettify(formatter='minimal')
```
Mr0grog: This is not the right thing to count for this diff, but this diff is also undergoing massive changes, so we can fix it as part of those.

web_monitoring/diffing_server.py
danielballan: Fair. I usually go by the rule, don't return a tuple unless both (1) the order is obvious and (2) the number of elements is small. This might violate (1) -- a matter of taste, I guess -- and could grow to violate (2) in the future. I'll make it a dict.

web_monitoring/html_diff_render.py
danielballan: Gotcha. I tried to figure this out but didn't think about it too hard. I should take the time to understand these internals better.

Mr0grog: :thumbsup: it’s a little complicated for sure. Hoping to add copious explanation and comments as I clean up #129.

danielballan:
Tests pass.

web_monitoring/html_diff_render.py
```diff
@@ -102,7 +102,8 @@ def html_diff_render(a_text, b_text):
     replacements_new.update(replacements_old)
     soup_new = _add_undiffable_content(soup_new, replacements_new)
 
-    return soup_new.prettify(formatter='minimal')
+    # TODO Count changes.
+    return {'diff': soup_new.prettify(formatter='minimal')}
```
Mr0grog: I think, in this version, the right thing is `len(soup_new.find_all('ins')) + len(soup_new.find_all('del'))`. We’d probably be fine without it, too.

PR #141, opened 2017-12-30
Add Dockerfile

danielballan:
Based on Docker best practices as documented here:
https://docs.docker.com/get-started/part2/


Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use an official Python runtime as a parent image
+FROM python:3.6-slim
```
Mr0grog: Don’t know if it *really* makes much difference, but a couple friends who do more Dockering than I have recently noted that Alpine is great and much tinier than Slim.

This post seems to be a big fan (plus has some other tips): https://blog.codeship.com/alpine-based-docker-images-make-difference-real-world-apps/

…but this one pooh-poohs it a bit and basically says “you can build your own better, lighter debian”: https://www.turnkeylinux.org/blog/alpine-vs-debian

…and this seems to suggest it’s quite a bit slower at runtime: https://superuser.com/questions/1219609/why-is-the-alpine-docker-image-over-50-slower-than-the-ubuntu-image

But maybe worth thinking about? ¯\\\_(ツ)\_/¯

If you know much about Alpine Linux or other people have experience on this one, that might be helpful.

danielballan: Interesting reading. I'm sympathetic with the argument that Alpine is a savvy move for Docker but not necessarily a net savings for human dev time. Security, runtime performance, familiarity all point toward Debian. We could try that much lighter Debian build in a future iteration, but I'd rather keep it simple at this stage.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git gcc g++
+
+# Set the working directory to /app
+WORKDIR /app
+
+# Copy the current directory contents into the container at /app
+ADD . /app
```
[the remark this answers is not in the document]
danielballan: Thanks, I don't know how I missed this -- I've used this pattern before.

Issue #145, opened 2018-01-11
html_token: Stop using `lxml.html.diff` for tokenization


Mr0grog:
In the `html_token` diff, the tokenization step is last part we are still relying directly on `lxml.html.diff` for and not our own forked code. There are quite a few reasons we need to bring this under our control:

- It does a needless serialize → parse step when we already have a parsed document

- When “cleaning up” the HTML, it removes `<ins>` and `<del>` elements that may have been in the source page and were probably meaningful there

- When “cleaning up” the HTML, if can make malformed HTML worse. In diff from #143 and #144, it removes all the styling and scripts from the page. (See #144 for more notes. Diff in question: https://monitoring.envirodatagov.org/page/12b5a7ea-cbe2-43b4-bf45-d39d8a4da44f/fdd561d2-b788-4c57-97a1-075dd4ca535d..a5096049-fee6-4cf2-9804-6168e75355fc

- We can do a *whole* lot more with handling more things we care about in the HTML if we control parsing (pay attention to `title` attributes and `alt` attributes on images, for example). This also means we don’t need a second pass to “customize” the generated tokens.

- We can handle non-diffable elements like `<script>`, `<svg>`, etc. right instead of doing the terrible thing we do now (rip those elements out before diffing and then carefully put them back afterward)

- We can ensure that structure of pre- and post- tags is correct instead of doing a second pass to balance it out: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/d647c53957fde542a3a4fdabc3335c2b5bd19051/web_monitoring/html_diff_render.py#L502-L522

Mr0grog:
Need to break this up into separate issues so that, when we get this forked, we don’t lose track of all the other things we still need to do that the forking *enables.*

PR #152, opened 2018-01-24
Fully fork lxml.html.diff; fix the foreign/undiffable content problem

### LINKED ISSUE #146: html_token: SVGs that have not changed look like they have and it’s a real problem
In this example image, SVG “share” links are getting highlighted as changes even though they have not changed at all:

screen_shot_2018-01-14_at_1 03 11_pm

This is a known issue that needs #145 (forking the LXML differ’s tokenization) in order to be solved and that we had chosen to wait on. It’s becoming a real issue for analysts, though, so it might make sense to prioritize even a minimal fix for it.

Mr0grog:
This finally fully forks `lxml.html.diff` (#145); we no longer utilize any parts of it (thought we do have large chunks of its tokenization routine now sitting in our module—but we will probably change those). 

This also fixes #146, where we always show SVG, MathML, styles, scripts, and other non-HTML content as changing, even if it didn’t. That was because of my earlier clever-but-horrible hack to remove that content before diffing (because the tokenizer would always try to tokenize the contents of those elements) and put it back in after. Now, because we own the tokenizing routine instead of using lxml’s, we can prevent the contents of those elements from being tokenized so they *actually get diffed,* but as single, complete tokens that never get broken up.

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -220,39 +220,16 @@ def html_diff_render(a_text, b_text, include='combined'):
         head = soup_new.new_tag('head')
         soup_new.html.insert(0, head)
 
-    # htmldiff will unfortunately try to diff the content of elements like
-    # <script> or <style> that embed foreign content that shouldn't be parsed
-    # as part of the DOM. We work around this by replacing those elements
-    # with placeholders, but a better upstream fix would be to have
-    # `flatten_el()` handle these cases by creating a special token, e.g:
-    #
-    #  class undiffable_tag(token):
-    #    def __new__(cls, html_repr, **kwargs):
-    #      # Make the value this represents for diffing an empty string
-    #      obj = token.__new__(cls, '', **kwargs)
-    #      # But keep the actual source around for serializing when done
-    #      obj.html_repr = html_repr
-    #
-    #    def html(obj):
-    #      return self.html_repr
-    soup_old, replacements_old = _remove_undiffable_content(soup_old, 'old')
-    soup_new, replacements_new = _remove_undiffable_content(soup_new, 'new')
-
     metadata, diff_bodies = diff_elements(soup_old.body, soup_new.body, include)
     results = metadata.copy()
 
     for diff_type, diff_body in diff_bodies.items():
         soup = None
-        replacements = None
         if diff_type == 'deletions':
             soup = copy.copy(soup_old)
-            replacements = copy.copy(replacements_old)
         elif diff_type == 'insertions':
             soup = copy.copy(soup_new)
-            replacements = copy.copy(replacements_new)
         else:
-            replacements = copy.copy(replacements_new)
-            replacements.update(replacements_old)
             soup = copy.copy(soup_new)
             title_meta = soup.new_tag(
                 'meta',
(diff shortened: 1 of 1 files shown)


web_monitoring/html_diff_render.py
```diff
@@ -1323,6 +1567,23 @@ def flatten_groups(groups, include_non_groups=True):
...
+class InsensitiveSequenceMatcher(difflib.SequenceMatcher):
+    """
+    Acts like SequenceMatcher, but tries not to find very small equal
+    blocks amidst large spans of changes
+    """
+
+    threshold = 2
+
+    def get_matching_blocks(self):
+        size = min(len(self.b), len(self.b))
```
danielballan: Is there a typo here? One of these should be `self.a` I bet. Obviously the value of `size`, as written, is always `len(self.b)`.

web_monitoring/html_diff_render.py
```diff
@@ -772,7 +1006,9 @@ def merge_changes(change_chunks, doc, tag_type='ins'):
         # FIXME: explicitly handle elements that can't have our markers as
         # direct children.
         if chunk[0] == '<':
-            name = chunk.split()[0].strip('<>/')
+            # NOTE: split first by `>` because we could have undiffable items
+            # with content here, i.e. more than one tag.
+            name = chunk.split('>')[0].split()[0].strip('<>/')
```
danielballan: Since we only care about the first element of `chunk.split('>')`, we can use `chunk.split('>', 1)` to save time. Same for `split()[0]` -> `split(None, 1)`.

web_monitoring/html_diff_render.py
```diff
@@ -1025,7 +1264,9 @@ def merge_change_groups(change_chunks, doc, tag_type=None):
...
```
danielballan: See previous comment about `split` optimization.

web_monitoring/html_diff_render.py
```diff
@@ -830,7 +1066,10 @@ def merge_changes(change_chunks, doc, tag_type='ins'):
             current_content = []
 
         doc.append(chunk)
-        if inline_tag and inline_tag_name not in empty_tags:
+        # Note the undiffable_content_tags check here. We assume tokens for
+        # those tags represent a whole element, not just a start or end tag,
+        # so we don't consider them "open" as part of `current_content`.
+        if inline_tag and inline_tag_name not in undiffable_content_tags and inline_tag_name not in empty_tags:
```
danielballan: Clarity suggestion:

```py
if (inline_tag and 
        (inline_tag_name not in undiffable_content_tags) and
        (inline_tag_name not in empty_tags)):
```

web_monitoring/html_diff_render.py
```diff
@@ -435,6 +371,304 @@ def _count_changes(opcodes):
     }
 
 
+# --------------------- lxml.html.diff Tokenization --------------------------
```
danielballan: Not reviewing any of this section. Let me know if you want me to look at it.

web_monitoring/html_diff_render.py
Mr0grog: http://lxml.de/api/lxml.html.diff-pysrc.html#InsensitiveSequenceMatcher.get_matching_blocks ¯\\\_(ツ)\_/¯

Mr0grog: To be clear, happy to change, but this was just a copy/paste job so we are running the same code we were running before.

danielballan: Odd. Seems safe to change it, in any case.

web_monitoring/html_diff_render.py
Mr0grog: The parts I modified (for undiffable stuff) are [564-569](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/152/files#diff-f483e5fdf85585c0850051651929f289R564) and [608-611](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/152/files#diff-f483e5fdf85585c0850051651929f289R564), if you want to look at that.

Mr0grog:
Updated to address review feedback.

Mr0grog:
:tada:

PR #155, opened 2018-02-05
html_token: Stop early with an error for non-HTML content

### LINKED ISSUE #147: html diffs: Should return early with a clear error if HTML does not parse
We should really easily be able to give better feedback than `type error` if a version does not parse as HTML (e.g. if someone fed the algorithm a PDF).

For example: https://api.monitoring.envirodatagov.org/api/v0/pages/4ca4bc6f-6dc2-4b01-8e7a-94d50baf276e/changes/e053c830-029a-4925-88c2-114b370c4fcc..8c5583c7-02f2-4ee2-a495-748865dc1330/diff/html_token?include=all

Mr0grog:
This is a simple attempt to resolve the big issues of asking the `html_token` differ to diff non-HTML content, which it simply can't do. It has minimal architectural changes, so it does its best to sniff the content its given to see if it it's almost certainly *not* HTML and stops early in that case. Note this is super lenient; it could get passed lots of types of invalid content and it would still try to diff it -- it works best against PDFs and images. Should we instead be more strict and look for things that actually appear to be HTML (as opposed to things that appear *not* to be HTML)?

It adds a custom `UndiffableContentError`, which is separated out because this would almost certainly be useful to other differs. I’m imagining we might benefit from other (or a more generic) errors for differs to return, but didn’t want to put in a whole bunch of speculative stuff yet.

I also have some big questions for you, @danielballan:

- Why does the diffing server [ignore text decoding errors](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/147-if-it-dont-look-like-html-you-best-not-treat-it-like-html/web_monitoring/diffing_server.py#L152-L153)? Binary content like images and PDFs could easily be stopped here. I’m assuming this was just quick-n-dirty code and now it would make sense to a) only decode if `a_text` or `b_text` is asked for and b) raise an error if decoding failed, but not sure if you did this intentionally to be more loose or with a different scenario in mind. (Maybe we should somehow pass the fact that a decoding error occurred to the differ?)

- It would be *really* useful to pass the diff routine the content type or just all the headers of the request. (Then we could check the content type header if one was specified.) Alternatively, there could be some secondary method (particular to each differ) the server might call to validate the input before calling the actual diff routine. Any thoughts on this?

Fixes #147


danielballan:
> now it would make sense to a) only decode if a_text or b_text is asked for and b) raise an error if decoding failed, but not sure if you did this intentionally to be more loose or with a different scenario in mind.

Yes.

> It would be really useful to pass the diff routine the content type or just all the headers of the request.

I think at the time I was trying to keep Headers separate from the diffing functions because I imagined that we would want to keep differs more generic/abstracted from web-related concerns. Now that seems completely misguided. This is a differ for content scraped from the web; it should be allowed to know about Headers.

danielballan:
> Should we instead be more strict and look for things that actually appear to be HTML (as opposed to things that appear not to be HTML)?

I wonder if we should expose through the REST API a flag that overrides the sniffer and/or forces a differ to attempt to diff something even if it thinks it is not applicable.

Anyway, I'm happy to merge/deploy this to resolve #147 and iterate on the more sweeping concerns separately.

Mr0grog:
> I'm happy to merge/deploy this to resolve #147 and iterate on the more sweeping concerns separately.

Ehh, I don’t think any of these are so complicated that I’m worried about the hold-up. Besides, there is also edgi-govdata-archiving/web-monitoring-ui#200, which helps on the other side. Happy to iterate here.

Re: decoding… I’m thinking only decide if `a_text` or `b_text` are in the signature. If it fails, should we:

- Ignore like now? (I’m not a huge fan of this)
- Immediately stop and return an error response? (My preferred.) (This means the diff routine can’t remediate, but if it really wanted to do special decoding tricks, it could just take the bytes and do it’s own decoding.)
- Pass `None`
- Pass `None` plus the exception as a new `a_decoding_error` argument.
- Pass the exception as the value instead of the decided string.

> I was trying to keep Headers separate from the diffing functions because I imagined that we would want to keep differs more generic/abstracted from web-related concerns. Now that seems completely misguided.

I definitely hear that. I’m thinking the headers could be optional (so you don’t *have* to pass them if using as a library function). Call them `a_headers` And `b_headers`?

> I wonder if we should expose through the REST API a flag that overrides the sniffer

I’m cool with that. I imagine it would override both checking headers and sniffing. What’s a good name?

- `force=true`
- `no_sniff=true`
- `sniff=false`
- `check_type=false`
- `content_type_options=nosniff` (like the response header)
- `content_type_options=nocheck`
- `content_type_options=ignore`
- Something else?

danielballan:
Per discussion we agreed on:

* In event of decoding errors, immediately stop and return an error response.
* Exposing `a_header` and `b_headers` as optional signature parameters sounds good.
* Add query param `ignore_decoding_errors=true`.
* `content_type_options`:
   * `normal` -- Check 'Context-Type'; sniff if it doesn't exit.
   * `nocheck` -- Check 'Context-Type; ignore if it doesn't exit.
   * `nosniff` -- Ignore 'Content-Type'; sniff.
   * `ignore` -- Diff no matter what.

Mr0grog:
Rebased to fix conflicts and implemented decoding error handling (and `?ignore_decoding_errors`).

@danielballan what do you think of the error-handling pattern that is developing here? Obviously, if it gets much bigger, we’ll need to have some kind of mapping or way for errors to provide their own status codes, but does the general idea seem good, or would you rather be explicitly handling those errors where they occur and writing responses in place? (e.g. how it’s done with [hash validation](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/diffing_server.py#L74-L77))

Mr0grog:
@danielballan I think this is good to go, so it could use a full review whenever you have time. See also the question above about error handling patterns.

web_monitoring/diffing_server.py
```diff
@@ -89,10 +92,22 @@ def get(self, differ):
 
     def write_error(self, status_code, **kwargs):
         response = {'code': status_code, 'error': self._reason}
+
+        # Handle errors that are allowed to be public
+        actual_error = 'exc_info' in kwargs and kwargs['exc_info'][1] or None
+        if (isinstance(actual_error, UndiffableContentError) or
```
danielballan: `isinstance(actual_error, (UndiffableContentError, UndecodableContentError))` would also work here

Mr0grog: OH! I didn’t even realize that could take a sequence as the second argument. 👍

Mr0grog:
Cleaned up review issues; will merge tomorrow if no concerns.

danielballan:
Go for it.

PR #161, opened 2018-02-22
Work harder to detect text encoding

Mr0grog:
AKA perhaps getting more strict on text encoding in #155 was not so wise as I thought.

Kendra ran into a text encoding issue today on https://monitoring.envirodatagov.org/page/bca19409-d4e5-40f3-9bd4-05c16bf88298/e5803dc5-957f-405e-bc8b-93ed95e7a1d2..38e96ee1-fafb-44d3-ae70-21337eb70ee6
API view: https://api.monitoring.envirodatagov.org/api/v0/pages/bca19409-d4e5-40f3-9bd4-05c16bf88298/changes/e5803dc5-957f-405e-bc8b-93ed95e7a1d2..38e96ee1-fafb-44d3-ae70-21337eb70ee6

No doubt there are *many* more of these waiting for people to trip over them.

I *suspect* this would have worked (but maybe with some oddities or mojibake in the output) before we made decoding super-strict in an effort to head off bad decoding of binary data like PDFs or images in #155. So I guess I’m walking that change back just a little: we should try to detect the encoding if the headers did not specify something. I think that’s better than just lazily ignoring errors like we did before.

In the change here, I added matching for encodings that can be specified in XML prologs and HTML `<meta>` tags. It only searches near the beginning of the document, so should not trip over that text if it happens to be specified in a PDF, for example.

That said, this might not be the right answer and this could be the starting point for something else.

- I don’t think staying strict and just updating the headers on our stored content is a good enough solution for the future.
- Should we use a library like chardet or UnicodeDammit (from BeautifulSoup)?
    - chardet doesn’t know how to read these XML/HTML indicators, so we’d probably want to use it in addition to the code here
    - UnicodeDammit does the stuff here, then uses chardet, then just falls back to a reasonable guess and forces things into place. That seems to me like it’s going *too* far for our purposes here—if we used it, we’d probably want to restrict it to certain content types. (Maybe diffs that want it should take bytes, do content detection sniffing, *then* use it to decode? Seems over-complicated, though.)
- Is this reasonable and general-purpose enough to do at the server level, or should the diff algorithms take bytes instead of text and do it there?

/cc @danielballan

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -2,6 +2,7 @@ import concurrent.futures
 from docopt import docopt
 import hashlib
 import inspect
+import re
 import tornado.gen
 import tornado.httpclient
 import tornado.ioloop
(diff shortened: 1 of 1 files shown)


danielballan:
To avoid more thrashing, I'm content to leave other dependencies out of it for now. Let's deploy this and see how it goes for awhile.

Mr0grog:
Sounds good to me. Merging.

PR #167, opened 2018-03-02
Replace site/agency with tags/maintainers

### LINKED ISSUE #166: Change `site`/`agency` to `tags`/`maintainers`
Update CLI args for the importer script and any related formatting functionality to send `tags` and `maintainers` instead of `site` and `agency`.

Not sure, but need to check if the DB client also needs updates (or anywhere else in the codebase).

Mr0grog:
Update all our import and listing methods to use `tags` and `maintainers` instead of `site` and `agency`. Also make them optional in all contexts where they are used.

A lot of the changes here are just refreshing the pre-recorded requests since we have had a lot of API changes.

Fixes #166.


web_monitoring/cli.py
```diff
@@ -27,24 +27,24 @@ def _add_and_monitor(versions):
         print("Errors: {}".format(errors))
 
 
-def import_ia(url, agency, site, from_date=None, to_date=None):
+def import_ia(url, from_date=None, to_date=None, maintainers=None, tags=None):
```
danielballan: I think we should make every argument except `url` a keyword-only argument to future-proof against adding/removing kwargs in the future.

Mr0grog: Sounds reasonable to me! Should be fixed in ddc27903362571893c17d6725873cd57c7bec0d0

Mr0grog:
@danielballan is this good now?

PR #170, opened 2018-03-14
Remove pf_edgi module and tests.

danielballan:
The PageFreezer deployment for EDGI has been taken down for now.
Future use of this API is expected to be one-off, in service
of migrating the data, so maintaining a Python RESTer for this API
no longer seems important. A backup has been pushed to the 'pf-edgi'
branch of the upstream fork.


Mr0grog:
👍

PR #171, opened 2018-03-14
Add a conda environment file.

danielballan:


Mr0grog:
Should this also modify the [deployment.md](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/deployment.md)?

deployment.md
```diff
@@ -24,10 +24,10 @@ $ curl <conda_url> > conda_installer.sh
 # e.g: curl https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > conda_installer.sh
 ```
 
-Then run the installer. Follow the prompts, but instead of the default install location, choose `/opt/conda`. (This gets us a system-wide install so the web server can use it.)
+Then run the installer. (These parameters get us a system-wide install so the web server can use it.)
 
 ```sh
-$ sudo bash conda_installer.sh
+$ sudo bash conda_installer.sh -b -p /opt/conda
```
Mr0grog: Ohhhh, this is much nicer!

Mr0grog:
Catching back up on this; is it good to go? (Can you add credentials to your Circle account so these tests can pass in the future, @danielballan?)

danielballan:
> Can you add credentials to your Circle account

Just did:

screen shot 2018-05-04 at 2 35 18 pm

And re-ran the build a couple times, but they didn't take. Maybe they only affect new PRs? But if that were true, it should would be difficult to debug issues with env vars....

My user experience with the Circle UI has been uniformly awful.

danielballan:
Oh -- saw your comment in the review too late. See #184. Thanks.

Issue #125, opened 2017-11-01
Create health check for Internet Archive data


Mr0grog:
The tool that auto-submits our URLs for crawls over at Internet Archive seems to crap out every so often. As we move towards regularly importing their data, we should have something that does a regular health check to get early alerts if it looks like something’s gone wrong on their end.

This probably looks like:

1. Grab a random sampling of 5 or 10 URLs we are following
2. Look for the latest snapshots of those URLs in IA
3. If none have snapshots within the last N hours (48? 72? probably should be adjustable), fire off an alert (e-mail, Slack, pagerduty, Sentry error? Any/all are viable).

weatherpattern:
Moved this to milestone v0.0.2 Stop Verisonista Imports.
This is still critical to have, but needed when we turn off Verisonista and move over to the Internet Archive.

weatherpattern:
This looks to me like script that can run independently from the rest of Scanner.
I'm going to pick this up, I have a friend who might be able to help me as well, if I get stuck. Where can I find the list of domain that we are following?

Mr0grog:
> Where can I find the list of domain that we are following?

You should query the API for this.

weatherpattern:
@Mr0grog @danielballan which folder should the script for this live? scripts? or web_monitoring?

weatherpattern:
For now, placed in the script in the edgi-scripts repo. It may move again later.

https://github.com/edgi-govdata-archiving/edgi-scripts/pull/28

Mr0grog:
Closing since #172 got merged.

PR #172, opened 2018-03-15
Check if Internet Archive is capturing our URLS

weatherpattern:
Addresses #125 

The original PR was in edgi-scripts [PR#28](https://github.com/edgi-govdata-archiving/edgi-scripts/pull/28), but got feedback to use existing db.py and internetarchive, which it does now. I adapted some of the code @Mr0grog suggested.

Currently, the output writes to a text file.

Still to do:
add method to send an alert to sentry.
add chron job to run daily


scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+MAX_CAPTURE_AGE = timedelta(hours=72)
+LINKS_TO_CHECK = 5
+
+
+def query_webmondb():
+    """
+    Query the web monitoring SCANNER API to get random links from the database
+    Interger -> List
```
Mr0grog: Seems like this is describing an integer input, but the function doesn’t actually take any args?

Also, we’re using the [Numpy docstring style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt), so this should be:

```
Returns
-------
list of string
```

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    try:
+        response = db.Client.from_env()
+    except Exception as e:
+        print('Could not access Web Monitoring API')
+        print(e)
+        return []
```
Mr0grog: Is there really any point in returning and going any further in this case? Maybe better to just let the exception crash the program.

Otherwise, if you want to suppress all the normal logging when the client is not configured, I’d suggest handling *specifically* the `MissingCredentials` error and letting all others crash, since they’re unexpected and you’ll want the full readout:

```py
except db.MissingCredentials as error:
    print(error)
    return []
```

Note I don’t think you really need a custom message; the `MissingCredentials` one is already pretty fancy: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/db.py#L154-L162

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        page = response.list_pages(chunk=1, chunk_size=1)
+        url_count = page['meta']['total_results']
+        return ([response.list_pages(chunk=number, chunk_size=1)
+            ['data'][0]['url']
+            for number in random.sample(range(url_count), LINKS_TO_CHECK)])
```
Mr0grog: This feels like a bit much to stuff into a comprehension. Maybe factor out the `response.list_pages(chunk=number, chunk_size=1)['data'][0]['url']` into its own function? Not sure; this just feels unwieldy.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def query_waybackCDX(url):
+    """
+    Query Wayback Machine using CDX API
+    Get a list of versions of a url, and check for ValueError.
+    String -> JSON
```
Mr0grog: Same note about docstring style as above.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+        return []
+    else:
+        page = response.list_pages(chunk=1, chunk_size=1)
+        url_count = page['meta']['total_results']
+        return ([response.list_pages(chunk=number, chunk_size=1)
+            ['data'][0]['url']
+            for number in random.sample(range(url_count), LINKS_TO_CHECK)])
+
+
+def query_waybackCDX(url):
```
Mr0grog: Should be `query_waback_cdx`

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Query Wayback Machine using CDX API
+    Get a list of versions of a url, and check for ValueError.
+    String -> JSON
+    """
+    try:
+        verisons = internetarchive.list_versions (url, from_date=datetime.now()
+            - MAX_CAPTURE_AGE)
```
Mr0grog: You should avoid splitting the line in the middle of a single argument value. Also, blank space between function name and `(` parameter list.

Also `verisons` → `versions`

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+        verision = next(verisons)
```
Mr0grog: Another, different typo in `verision`.

Also, no need to assign this to a variable, right?

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def query_webmondb():
+    """
+    Query the web monitoring SCANNER API to get random links from the database
+    Interger -> List
+    """
+    try:
+        response = db.Client.from_env()
+    except Exception as e:
```
Mr0grog: Please avoid single-letter variable names.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    Query Wayback Machine using CDX API
+    Get a list of versions of a url, and check for ValueError.
+    String -> JSON
+    """
+    try:
+        verisons = internetarchive.list_versions (url, from_date=datetime.now()
+            - MAX_CAPTURE_AGE)
+        verision = next(verisons)
+    except ValueError as e:
+        print(e)
```
Mr0grog: Please avoid single-letter variable names.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+        status = False
+        return status
```
Mr0grog: Unnecessary assignment.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    except:
+        print("Could not access Wayback CDX API")
```
Mr0grog: Might be useful to re-reaise the error so you get a stack trace in this case. This error would necessarily be something unexpected, so you’d probably want that, right?

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        status = True
+        return status
```
Mr0grog: Unnecessary assignment.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def output_file(responses):
+    """
+    Write Output to a Text file
+    List -> None
```
Mr0grog: Same issue about docstring style.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    fileoutput = open("ia_healthcheck.txt", "w")
```
Mr0grog: You should either use a context manager:

```py
with open('ia_healthcheck.txt', 'w') as fileoutput:
    # code here
```

Or make sure any possible errors here are handled so you don’t leave the file handle open.

---

Also, maybe useful to make this path a CLI arg or at least an argument to this function so it can be configured.

OR! Skip worrying about a file altogether here and write to stdout; then someone calling this program can just `ia_healthcheck > somefile.txt` or `ia_healthcheck | grep 'False:'` or whatever. You’ll have to make sure all your other `print()` calls go to `stderr`, though.

Finally, maybe consider a more machine-readable format for the output here, e.g. YAML, or JSON? Dunno, this is simple enough it might be fine, too.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    fileoutput.close()
+    return
+
+
+def output_email():
+    """
+    Send Output to and Email
+    List -> None
+    """
+    return
```
Mr0grog: There is a built-in `NotImplementedError` designed just to be raised in these circumstances.

Also, docstring format.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Get the random list of links from the Web Monitoring DB
+# Get the responses of the links from the Wayback URL
+# Check to see if the responses are within the time limit and write the output
+
+
+if __name__ == "__main__":
+
+    links = query_webmondb()
+    links = ['sdfs', 'http://epa.gov']
```
Mr0grog: Please remove debug code.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+    responses = [(url, query_waybackCDX(url)) for url in links if url]
```
Mr0grog: Is it possible to get `None` or an empty string for `url` here? If I’m reading right, it doesn’t seem like it should be.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
+from datetime import datetime, timedelta
```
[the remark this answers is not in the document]
weatherpattern: fixed.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+import random
+from ..web_monitoring import db
```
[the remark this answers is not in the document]
weatherpattern: Fixed. Works without the .. now.

scripts/ia_healthcheck.py
```diff
@@ -0,0 +1,92 @@
...
+from ..web_monitoring import internetarchive
+
+
+MAX_CAPTURE_AGE = timedelta(hours=72)
+LINKS_TO_CHECK = 5
```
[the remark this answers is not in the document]
weatherpattern: Not sure how to make these CLI options.

scripts/ia_healthcheck.py
weatherpattern: Corrected parameter mistake, and added Numpy docstring style.

scripts/ia_healthcheck.py
weatherpattern: Removed custom message and added MissingCredentials.

scripts/ia_healthcheck.py
weatherpattern: Added Numpy docstring style.

scripts/ia_healthcheck.py
weatherpattern: Went with the 2nd suggestion, and writes to stdout.

scripts/ia_healthcheck.py
weatherpattern: Added NotImplementedError.
Added Numpy docstring style.

scripts/ia_healthcheck.py
weatherpattern: fixed.

scripts/ia_healthcheck.py
weatherpattern: I think you are correct, and it shouldn't even be Non or an empty string.

Mr0grog:
Huh, those are all *replies* in existing threads. Looks pretty confusing out of context above :\

Mr0grog:
Tried to clean up the commit history here and rebase things on top of current master as a first step to moving this towards done.

Mr0grog:
Previous version of this branch is now at [`ia-healthcheck-old`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/tree/ia-healthcheck-old); need to remember to delete it when this is merged.

Mr0grog:
That deployment doc is specifically about the diff server; I don’t think it really makes sense to cover this there.

jsnshrmn:
Okay, so deployment.md is not the correct doc. More generally, a note still needs to go somewhere about executing this on a recurring basis. IA healthceck code that's never fired won't do us much good.

Mr0grog:
Yes. We’ve talked about this before, and it probably belongs with all the other ETL-related stuff on the scraper server, and should be documented along with all the other deployments in the services inventory doc.

PR #173, opened 2018-03-18
Remove remaining refs to removed pf_edgi module

Mr0grog:
We removed `pf_edgi` in #170, but it's still referenced in a bunch of places, included the `cli` module, which means all `cli` commands are broken!

We should probably add some kind of tests for CLI commands, but I just want to get them working again first.


danielballan:
Oops. Merge at will once Circle-CI is happy. Looks like you are iterating as I type this.

Mr0grog:
D’oh, I blindly went in and did that thinking the importing functionality was a different model for who-knows-what reason. Added another commit to remove the command that uses the removed code, which should clear us up.

PR #174, opened 2018-03-18
Add command to import known page URLs from IA

### LINKED ISSUE #86: Add way to import from IA only the pages/URLs we already have in the DB
This was originally going to be part of #82, but we’re splitting it off because it’s probably fairly complicated on its own and also requires updates to the DB API (e.g. edgi-govdata-archiving/web-monitoring-db#128).

Mr0grog:
Fixes #86, based on #173.

This adds the `import ia-known-pages` command, which imports only pages we already know about in the DB from Internet Archive. Use it like so:

```sh
> ./scripts/wm import ia-known-pages --from 2018-03-17
Importing 192 URLs from 2 Domains:
  yosemite.epa.gov
  www.usgs.gov
importing: 0 versions [00:00, ? versions/s]Submitting Versions to web-monitoring-db...
importing: 7 versions [00:18,  1.85s/ versions]
Import jobs IDs: (66,)
Polling web-monitoring-db until import jobs are finished...
```

If you set `LOG_LEVEL=INFO`, it’ll tell you how many URLs IA had that we skipped because we didn’t know them.

If you set `LOG_LEVEL=DEBUG`, it’ll print every skipped URL.

### DIFF
--- a/web_monitoring/__init__.py
+++ b/web_monitoring/__init__.py
@@ -7,8 +7,3 @@ del get_versions
 
 if os.environ.get('LOG_LEVEL'):
     logging.basicConfig(level=os.environ['LOG_LEVEL'].upper())
-    # Urllib3 logs a warning every time a retry happens. Way too noisy.
-    # TODO: is this the best way to configure? We should probably revisit this.
-    # Maybe the right answer is filtering when *reading* the output.
-    subcomponent_level = os.environ.get('LOG_LEVEL_SUBCOMPONENT', 'ERROR')
-    logging.getLogger('urllib3.connectionpool').setLevel(subcomponent_level)
--- a/web_monitoring/__init__.py
+++ b/web_monitoring/__init__.py
@@ -7,3 +7,8 @@ del get_versions
 
 if os.environ.get('LOG_LEVEL'):
     logging.basicConfig(level=os.environ['LOG_LEVEL'].upper())
+    # Urllib3 logs a warning every time a retry happens. Way too noisy.
+    # TODO: is this the best way to configure? We should probably revisit this.
+    # Maybe the right answer is filtering when *reading* the output.
+    subcomponent_level = os.environ.get('LOG_LEVEL_SUBCOMPONENT', 'ERROR')
+    logging.getLogger('urllib3.connectionpool').setLevel(subcomponent_level)
--- a/web_monitoring/cli.py
+++ b/web_monitoring/cli.py
@@ -1,51 +1,436 @@
 # Command Line Interface
 # See scripts/ directory for associated executable(s). All of the interesting
 # functionality is implemented in this module to make it easier to test.
+from collections import defaultdict
+from datetime import datetime, timedelta
 from docopt import docopt
+import json
+import logging
+from os.path import splitext
 import pandas
+from pathlib import Path
+import re
+import requests
+import signal
+import sys
+import threading
 from tqdm import tqdm
+from urllib.parse import urlparse
 from web_monitoring import db
 from web_monitoring import internetarchive as ia
+from web_monitoring import utils
+
+import queue
+import asyncio
+import concurrent
+
+
+logger = logging.getLogger(__name__)
+
+# Number of memento requests to make at once. Can be overridden via CLI args.
+PARALLEL_REQUESTS = 10
+
+# Matches the host segment of a URL.
+HOST_EXPRESSION = re.compile(r'^[^:]+://([^/]+)')
+# Matches URLs for "index" pages that are likely to be the same as a URL ending
+# with a slash. e.g. matches the `index.html` in `https://epa.gov/index.html`.
+# Used to group URLs representing the same logical page.
+INDEX_PAGE_EXPRESSION = re.compile(r'index(\.\w+)?$')
+# MIME types that we always consider to be subresources and never "pages".
+SUBRESOURCE_MIME_TYPES = (
+    'text/css',
+    'text/javascript',
+    'application/javascript',
+    'image/jpeg',
+    'image/webp',
+    'image/png',
+    'image/gif',
+    'image/bmp',
+    'image/tiff',
+    'image/x-icon',
+)
+# Extensions that we always consider to be subresources and never "pages".
+SUBRESOURCE_EXTENSIONS = (
+    '.css',
+    '.js',
+    '.es',
+    '.es6',
+    '.jsm',
+    '.jpg',
+    '.jpeg',
+    '.webp',
+    '.png',
+    '.gif',
+    '.bmp',
+    '.tif',
+    '.ico',
+)
+# Never query CDX for *all* snapshots at any of these domains (instead, always
+# query for each specific URL we want). This is usually because we assume these
+# domains are HUGE and completely impractical to query all pages on.
+NEVER_QUERY_DOMAINS = (
+    'instagram.com',
+    'youtube.com',
+    'amazon.com'
+)
+# Query an entire domain for snapshots if we are interested in more than this
+# many URLs in the domain (NEVER_QUERY_DOMAINS above overrides this).
+MAX_QUERY_URLS_PER_DOMAIN = 30
 
 
 # These functions lump together library code into monolithic operations for the
 # CLI. They also print. To access this functionality programmatically, it is
 # better to use the underlying library code.
 
+def _get_progress_meter(iterable):
+    # Use TQDM in all environments, but don't update very often if not a TTY.
+    # Basically, the idea here is to keep TQDM in our logs so we get stats, but
+    # not to waste a huge amount of space in the logs with it.
+    # NOTE: This is cribbed from TQDM's `disable=None` logic:
+    # https://github.com/tqdm/tqdm/blob/f2a60d1fb9e8a15baf926b4a67c02f90e0033eba/tqdm/_tqdm.py#L817-L830
+    file = sys.stderr
+    intervals = {}
+    if hasattr(file, "isatty") and not file.isatty():
+        intervals = dict(mininterval=10, maxinterval=60)
 
-def _add_and_monitor(versions):
+    return tqdm(iterable, desc='importing', unit=' versions', **intervals)
+
+
+def _add_and_monitor(versions, create_pages=True, skip_unchanged_versions=True, stop_event=None):
     cli = db.Client.from_env()  # will raise if env vars not set
     # Wrap verions in a progress bar.
-    versions = tqdm(versions, desc='importing', unit=' versions')
-    print('Submitting Versions to web-monitoring-db...')
-    import_ids = cli.add_versions(versions)
+    # TODO: create this on the main thread so we can update totals when we
+    # discover them in CDX, but update progress here as we import.
+    versions = _get_progress_meter(versions)
+    import_ids = cli.add_versions(versions, create_pages=create_pages,
+                                  skip_unchanged_versions=skip_unchanged_versions)
     print('Import jobs IDs: {}'.format(import_ids))
     print('Polling web-monitoring-db until import jobs are finished...')
-    errors = cli.monitor_import_statuses(import_ids)
+    errors = cli.monitor_import_statuses(import_ids, stop_event)
     if errors:
         print("Errors: {}".format(errors))
 
 
-def import_ia(url, *, from_date=None, to_date=None, maintainers=None,
-              tags=None, skip_unchanged='resolved-response'):
+def _log_adds(versions):
+    versions = _get_progress_meter(versions)
+    for version in versions:
+        print(json.dumps(version))
+
+
+class WaybackRecordsWorker(threading.Thread):
+    def __init__(self, records, results_queue, maintainers, tags, cancel,
+                 failure_queue=None, session_options=None,
+                 unplaybackable=None):
+        super().__init__()
+        self.summary = self.create_summary()
+        self.results_queue = results_queue
+        self.failure_queue = failure_queue
+        self.cancel = cancel
+        self.records = records
+        self.maintainers = maintainers
+        self.tags = tags
+        self.unplaybackable = unplaybackable
+        session_options = session_options or dict(retries=3, backoff=2,
+                                                  timeout=(30.5, 2))
+        session = ia.WaybackSession(**session_options)
+        self.wayback = ia.WaybackClient(session=session)
+
+    def reset_client(self):
+        self.wayback.session.reset()
+
+    def is_active(self):
+        return not self.cancel.is_set()
+
+    def run(self):
+        """
+        Work through the queue of CDX records to load them from Wayback,
+        transform them to Web Monitoring DB import entries, and queue them for
+        importing.
+        """
+        self.reset_client()
+
+        while self.is_active():
+            try:
+                record = next(self.records)
+                self.summary['total'] += 1
+            except StopIteration:
+                break
+
+            self.handle_record(record, retry_connection_failures=True)
+
+        self.wayback.close()
+        return self.summary
+
+    def handle_record(self, record, retry_connection_failures=False):
+        """
+        Handle a single CDX record.
+        """
+        # Check for whether we already know this can't be played and bail out.
+        if self.unplaybackable is not None and record.raw_url in self.unplaybackable:
+            self.summary['playback'] += 1
+            return
+
+        try:
+            version = self.process_record(record, retry_connection_failures=True)
+            self.results_queue.put(version)
+            self.summary['success'] += 1
+        except ia.MementoPlaybackError as error:
+            self.summary['playback'] += 1
+            if self.unplaybackable is not None:
+                self.unplaybackable[record.raw_url] = datetime.utcnow()
+            logger.info(f'  {error}')
+        except requests.exceptions.HTTPError as error:
+            if error.response.status_code == 404:
+                logger.info(f'  Missing memento: {record.raw_url}')
+                self.summary['missing'] += 1
+            else:
+                # TODO: consider not logging this at a lower level, like debug
+                # unless failure_queue does not exist. Unsure how big a deal
+                # this error is to log if we are retrying.
+                logger.info(f'  (HTTPError) {error}')
+                # TODO: definitely don't count it if we are going to retry it
+                self.summary['unknown'] += 1
+                if self.failure_queue:
+                    self.failure_queue.put(record)
+        except ia.WaybackRetryError as error:
+            # TODO: don't count or log (well, maybe DEBUG log) if failure_queue
+            # is present and we are ultimately going to retry.
+            self.summary['unknown'] += 1
+            logger.info(f'  {error}; URL: {record.raw_url}')
+
+            if self.failure_queue:
+                self.failure_queue.put(record)
+        except Exception as error:
+            # FIXME: getting read timed out connection errors here...
+            # requests.exceptions.ConnectionError: HTTPConnectionPool(host='web.archive.org', port=80): Read timed out.
+            # TODO: don't count or log (well, maybe DEBUG log) if failure_queue
+            # is present and we are ultimately going to retry.
+            self.summary['unknown'] += 1
+            logger.exception(f'  ({type(error)}) {error}; URL: {record.raw_url}')
+
+            if self.failure_queue:
+                self.failure_queue.put(record)
+
+    def process_record(self, record, retry_connection_failures=False):
+        """
+        Load the actual Wayback memento for a CDX record and transform it to
+        a Web Monitoring DB import record.
+        """
+        try:
+            return self.wayback.timestamped_uri_to_version(record.date,
+                                                           record.raw_url,
+                                                           url=record.url,
+                                                           maintainers=self.maintainers,
+                                                           tags=self.tags,
+                                                           view_url=record.view_url)
+        except Exception as error:
+            # On connection failures, reset the session and try again. If we
+            # don't do this, the connection pool for this thread is pretty much
+            # dead. It's not clear to me whether there is a problem in urllib3
+            # or Wayback's servers that requires this.
+            # This unfortunately requires string checking because the error can
+            # get wrapped up into multiple kinds of higher-level errors :(
+            if retry_connection_failures and ('failed to establish a new connection' in str(error).lower()):
+                self.reset_client()
+                return self.process_record(record)
+
+            # Otherwise, re-raise the error.
+            raise error
+
+    @classmethod
+    def create_summary(cls):
+        return {'total': 0, 'success': 0, 'playback': 0, 'missing': 0,
+                'unknown': 0}
+
+    @classmethod
+    def summarize(cls, workers):
+        return cls.merge_summaries((w.summary for w in workers))
+
+    @classmethod
+    def merge_summaries(cls, summaries):
+        merged = cls.create_summary()
+        for summary in summaries:
+            for key in merged.keys():
+                merged[key] += summary[key]
+
+        # Add percentage calculations
+        if merged['total']:
+            merged.update({f'{k}_pct': 100 * v / merged['total']
+                           for k, v in merged.items()
+                           if k != 'total' and not k.endswith('_pct')})
+        else:
+            merged.update({f'{k}_pct': 0.0
+                           for k, v in merged.items()
+                           if k != 'total' and not k.endswith('_pct')})
+
+        return merged
+
+
+def iterate_into_queue(iterable, queue):
+    for item in iterable:
+        queue.put(item)
+    queue.put(None)
+
+
+async def import_ia_db_urls(*, from_date=None, to_date=None, maintainers=None,
+                            tags=None, skip_unchanged='resolved-response',
+                            url_pattern=None, worker_count=0,
+                            unplaybackable_path=None, dry_run=False):
+    client = db.Client.from_env()
+    logger.info('Loading known pages from web-monitoring-db instance...')
+    urls, version_filter = _get_db_page_url_info(client, url_pattern)
+
+    # Wayback search treats URLs as SURT, so dedupe obvious repeats first.
+    www_subdomain = re.compile(r'^https?://www\d*\.')
+    urls = set((www_subdomain.sub('http://', url) for url in urls))
+
+    logger.info(f'Found {len(urls)} CDX-queryable URLs')
+    logger.debug('\n  '.join(urls))
+
+    return await import_ia_urls(
+        urls=urls,
+        from_date=from_date,
+        to_date=to_date,
+        maintainers=maintainers,
+        tags=tags,
+        skip_unchanged=skip_unchanged,
+        version_filter=version_filter,
+        worker_count=worker_count,
+        create_pages=False,
+        unplaybackable_path=unplaybackable_path,
+        dry_run=dry_run)
+
+
+def load_from_wayback(records, versions_queue, maintainers, tags, cancel, unplaybackable, worker_count, tries=None):
+    if tries is None or len(tries) == 0:
+        tries = (None,)
+    if isinstance(records, queue.Queue):
+        records = utils.ThreadSafeIterator(utils.queue_iterator(records))
+
+    summary = WaybackRecordsWorker.create_summary()
+
+    total_tries = len(tries)
+    retry_queue = None
+    for index, try_setting in enumerate(tries):
+        if retry_queue and not retry_queue.empty():
+            print(f'\nRetrying about {retry_queue.qsize()} failed records...', flush=True)
+            retry_queue.put(None)
+            records = utils.ThreadSafeIterator(utils.queue_iterator(retry_queue))
+
+        if index == total_tries - 1:
+            retry_queue = None
+        else:
+            retry_queue = queue.Queue()
+
+        workers = []
+        for i in range(worker_count):
+            worker = WaybackRecordsWorker(records, versions_queue,
+                                          maintainers, tags, cancel,
+                                          retry_queue, try_setting,
+                                          unplaybackable)
+            workers.append(worker)
+            worker.start()
+
+        for worker in workers:
+            worker.join()
+
+        try_summary = WaybackRecordsWorker.summarize(workers)
+        if index == 0:
+            summary = try_summary
+        else:
+            summary['success'] += try_summary['success']
+            summary['playback'] += try_summary['playback']
+            summary['missing'] += try_summary['missing']
+            summary['unknown'] -= (try_summary['success'] +
+                                   try_summary['playback'] +
+                                   try_summary['missing'])
+
+    # Recalculate percentages
+    summary = WaybackRecordsWorker.merge_summaries([summary])
+    return summary
+
+
+# TODO: this function probably be split apart so `dry_run` doesn't need to
+# exist as an argument.
+async def import_ia_urls(urls, *, from_date=None, to_date=None,
+                         maintainers=None, tags=None,
+                         skip_unchanged='resolved-response',
+                         version_filter=None, worker_count=0,
+                         create_pages=True, unplaybackable_path=None,
+                         dry_run=False):
     skip_responses = skip_unchanged == 'response'
-    with ia.WaybackClient() as wayback:
-        # Pulling on this generator does the work.
-        versions = (wayback.timestamped_uri_to_version(version.date,
-                                                       version.raw_url,
-                                                       url=version.url,
-                                                       maintainers=maintainers,
-                                                       tags=tags,
-                                                       view_url=version.view_url)
-                    for version in wayback.list_versions(url,
-                                                         from_date=from_date,
-                                                         to_date=to_date,
-                                                         skip_repeats=skip_responses))
+    worker_count = worker_count if worker_count > 0 else PARALLEL_REQUESTS
+    unplaybackable = load_unplaybackable_mementos(unplaybackable_path)
 
-        if skip_unchanged == 'resolved-response':
-            versions = _filter_unchanged_versions(versions)
+    # Use a custom session to make sure CDX calls are extra robust.
+    session = ia.WaybackSession(retries=10, backoff=4)
 
-        _add_and_monitor(versions)
+    with utils.QuitSignal((signal.SIGINT, signal.SIGTERM)) as stop_event:
+        with ia.WaybackClient(session) as wayback:
+            # wayback_records = utils.ThreadSafeIterator(
+            #     _list_ia_versions_for_urls(
+            #         urls,
+            #         from_date,
+            #         to_date,
+            #         skip_responses,
+            #         version_filter,
+            #         client=wayback))
+
+            executor = concurrent.futures.ThreadPoolExecutor(max_workers=worker_count + 1)
+            loop = asyncio.get_event_loop()
+            versions_queue = queue.Queue()
+            versions = utils.queue_iterator(versions_queue)
+            if skip_unchanged == 'resolved-response':
+                versions = _filter_unchanged_versions(versions)
+
+            if dry_run:
+                uploader = loop.run_in_executor(executor, _log_adds, versions)
+            else:
+                uploader = loop.run_in_executor(executor, _add_and_monitor, versions, create_pages, stop_event)
+
+            # summary = merge_worker_summaries([worker_summary()])
+            summary = WaybackRecordsWorker.create_summary()
+            all_records = _list_ia_versions_for_urls(
+                urls,
+                from_date,
+                to_date,
+                skip_responses,
+                version_filter,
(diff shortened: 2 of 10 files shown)


web_monitoring/cli.py
```diff
@@ -41,14 +47,72 @@ def import_ia(url, *, from_date=None, to_date=None, maintainers=None,
...
+    for page in _list_all_db_pages(client):
+        url_keys.add(page['url_key'])
+        domains.add(HOST_EXPRESSION.match(page['url']).group(1))
+
+    return domains, url_keys
+
+
+# TODO: this should probably be a method on db.Client, but db.Client could also
+# do well to transform the `links` into callables, e.g:
+#     more_pages = pages['links']['next']()
```
Mr0grog: I was wary of actually doing this here because it:

- Probably has a lot of room for discussion over different APIs/implementations
- Ought to be applied across all `list_*` methods, which is not small
- Probably entails big changes because we ought to break apart the URL composition/request from the response parsing bits.

Mr0grog:
Also unsure if we should be using Tornado or Asyncio or something for this, since it’s probably a lot of HTTP traffic.

Mr0grog:
Need to rebase and incorporate work from #179.

Mr0grog:
Rebased on master and added `--skip-unchanged` support that got merged in #179.

Mr0grog:
Ugh, this got way more complicated, but I think it’s probably doing the job better now.

web_monitoring/cli.py
danielballan: Agreed to all of the above.

Mr0grog:
I think I need to re-working things here based on discussions I had with IA folks recently about revisit records and other kinds of un-playback-able records. We’ve made some pretty incorrect assumptions here about what we’re getting when following redirects and looking up CDX results.

- Many mementos are not playback-able (for a variety of reasons). Requests to URLs corresponding to a non-playback-able memento are treated just like requests to non-existent mementos: you get redirected to the nearest-in-time playback-able one. You can tell whether you are getting a playback by checking for the presence of the `Memento-Datetime` header. It is only present in playback responses (included playbacks that are redirects).

- “Self-redirects,” or redirects to the same URL at a different protocol are not playback-able.

- Revisits *may* be playback-able. If there are very few captures of the URL, the revisit will play back. Otherwise, it will not. This threshold is arbitrary and potentially malleable, so we can’t meaningfully know when a revisit will be playback-able.

- You can find the actual response data corresponding to a revisit by searching *the whole domain* (since the revisit may be a response from a totally different page) for the revisit’s hash, using the `?filter=digest:{hash goes here}` query arg, but that query is very slow. **This only gets a matching response body, though!** If the response was a redirect with an empty body, this information is not going to help you because it doesn’t include headers — and you can’t tell from the CDX record what the revisit’s status code was (i.e. you can’t tell if it was a 3xx redirect).

The first major thing to note here is that this means not every CDX result can be transformed into an importable snapshot/version. We’re also going to encounter lots of revisits in our pages because they scrape so often, and knowing how to best handle those feels like it needs some thought.

Mr0grog:
Rebased on master.

Also worth noting, re: the last comment about IA import issues, we talked about it in [today’s meeting](https://docs.google.com/document/d/1PbDML8uHAWFjo2YEJvG0sCjnJWDdSNNu7UdvLL19wFo/edit) and I don’t think there are any clever solutions that will work well (mainly because of redirects). So the best we can do here is to fail saving a memento in [`timestamped_uri_to_version`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/7be25c89759387e58a7bd8f35fa34e16264a8fe0/web_monitoring/internetarchive.py#L312-L371) if the memento isn’t playback-able.

Mr0grog:
Once we are happy with #255, I’ll rebase this on it. For now, the rebased and improved version is on [`86-import-known-db-pages-from-ia-2`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/tree/86-import-known-db-pages-from-ia-2).

Mr0grog:
Rebased this insane mess, now that #255 is merged.

web_monitoring/cli.py
```diff
@@ -60,6 +143,99 @@ def _filter_unchanged_versions(versions):
...
+                                           skip_repeats=skip_repeats)
+        try:
+            for version in ia_versions:
+                if version_filter(version):
+                    yield version
+                else:
+                    skipped += 1
+                    logger.debug('Skipping URL "%s"', version.url)
+        except ValueError as error:
+            logger.warn(error)
```
danielballan: Might as well use [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.exception) for this.

```py
except ValueError:
    logger.exception('Error while applying version_filter to URL %r', url)
```

The Traceback will be automatically captured by `logger.exception` and shown below the log message.

web_monitoring/cli.py
```diff
@@ -26,21 +63,67 @@ def _add_and_monitor(versions):
...
+                try:
+                    yield wayback.timestamped_uri_to_version(version.date,
+                                                             version.raw_url,
+                                                             url=version.url,
+                                                             maintainers=maintainers,
+                                                             tags=tags,
+                                                             view_url=version.view_url)
+                except ia.MementoPlaybackError as error:
+                    wayback_errors['playback'].append(error)
+                    logger.info(f'  {error}')
```
danielballan: In contrast to my comment about `logger.exception`, _these_ messages make sense as INFO messages to me because they are tracking expected errors coming from a service that issues them frequently.

Mr0grog:
Ahhhh, sorry, wasn’t clear; I wasn’t really feeling like this was super mergeable at this point. Was planning to try and do the async bit and clean all this up (e.g. they flow between things has gotten super wonky; this needs to at least be extracted out of `cli.py` and re-organized a bit, maybe into a class or classes).

web_monitoring/cli.py
Mr0grog: Oh, this is really here to handle the error if there were no versions to list: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/4bb08f886a14e686f7eb0c68f2d853673aa6bd62/web_monitoring/internetarchive.py#L377-L379

…which, well, probably shouldn’t actually be a `ValueError` at all, since it’s not really a problem with your input. Actually, I’m not really sure why we have an exception for this at all. Should we get rid of that?

Mr0grog: (Anyway, implicit point there is that this isn’t really an *exception* or unexpected scenario, so I don’t think `logger.exception` makes sense for it. But the code and the error classes involved really don’t make that clear.)

danielballan:
Gotcha. This seemed either ready or _not at all_ ready, depending on scope.

Mr0grog:
Yeah, I guess we could merge it in an ugly state. My *main* concern is the async bit, because a full run of this across all 90-ish domain names we are monitoring can take a *while*. I think we need parallel downloading from IA to make this feasible.

danielballan:
I agree and I don't think we need to rush it. Was just trying to figure out if your "Rebased" comment implied "So can we merged this please?" or not. :-D Let's wait for async.

Mr0grog:
OK, this is, uh, still super messy, but is parallel and does function and I just started a full overnight test run against the staging API. 😬

I would like to rename and at least slightly refactor a lot of things, but feel free to take a look and comment on any broad-strokes issues.

web_monitoring/internetarchive.py
```diff
@@ -421,9 +451,16 @@ def timestamped_uri_to_version(self, dt, uri, *, url,
 
             # If the playback includes a redirect, continue on.
             if res.status_code >= 300 and res.status_code < 400:
+                target = res.headers.get('location')
+                # Wayback sometimes has circular memento redirects ¯\_(ツ)_/¯
+                # TODO: probably need to detect longer loops than just URLs
+                # that redirect to themselves. Need to interrupt every step
+                # along a series of redirects with all this error-handling :(
+                if target == uri:
```
Mr0grog: Update: since trying to run again with all the latest changes I’ve made here, I’ve already seen several examples where the original URL is a memento, but somewhere along the redirect chain we stop getting playbacks. Here’s one example: 
http://web.archive.org/web/20180316103545id_/http://www.nws.noaa.gov/sp/

So I guess I’d better fix this now :\

Should look into [`Session.resolve_redirects()`](http://docs.python-requests.org/en/master/api/?highlight=redirect#requests.Session.resolve_redirects) for this.

Mr0grog: There is also [`Response.next()`](http://docs.python-requests.org/en/master/api/?highlight=redirect#requests.Response.next).

web_monitoring/internetarchive.py
```diff
@@ -381,6 +438,107 @@ def list_versions(self, url, *,
...
+            response = utils.retryable_request('GET', url, allow_redirects=False, session=self.session)
+            while True:
+                is_memento = 'Memento-Datetime' in response.headers
+
+                if not is_memento:
+                    # If handling a reponse the original memento redirected to,
+                    # a non-memento redirect may be ok. The target URL will
+                    # rarely have been captured at the same time. (See 2b)
+                    playable = False
+                    if previous_was_memento and response.is_redirect:
```
Mr0grog: Ugh, we can actually have multiple redirects following from a memento and it’s *probably* ok. See this one, for example: http://web.archive.org/web/20180918070531id_/https://www.epa.gov/cleanpowerplan/clean-power-plan-existing-power-plants

```
http://web.archive.org/web/20180918070531id_/https://www.epa.gov/cleanpowerplan/clean-power-plan-existing-power-plants (memento)
http://web.archive.org/web/20180918070531id_/https://www.epa.gov/Energy-Independence (not a memento)
http://web.archive.org/web/20180918073149id_/https://www.epa.gov/energy-independence (not a memento)
http://web.archive.org/web/20180918032922id_/https://www.epa.gov/energy-independence (memento)
```

I’m reasonably certain the `/energy-independence` memento URL from 20180918073149 (3rd in chain) is redirecting because it’s a non-playback-able revisit (it is listed as a revisit in CDX).

So the question is: is a redirect like this acceptable? We don’t allow it for the initial request (in the end is just means the snapshot you asked for couldn’t be played back), but exactly what does this mean when it’s somewhere down the chain of redirects, and you ultimately are sort of looking up a different snapshot? In the strictest sense, we are currently correct in disallowing it — it means the correct capture couldn’t be played back, just like we treat the initial response. It’s iffy whether the URL you get redirected to at this point is likely to still represent the same content, especially if you don’t have a list of CDX results with hashes to guide you (and even then, problematic for redirect mementos.

Mr0grog: Side note: notice the URL casing changes above. Because everything in wayback is by url_key, which is always lower-case, that should probably be OK. The comparison below on the next line should be case-insensitive.

Mr0grog:
Fixed the URL case issue, and running this experimentally against staging for a three-day period. We’ll see how long *that* takes.

There is a lot to be fixed here, but it’s functional, so maybe we can at least start it running in production (and maybe merge and fix these later):

- Rewrite retry/backoff as a urllib3/requests-native `Retry` class
- Lots of obvious cleaning and refactoring for both style and sanity
- Correct signal handling for the threads (so `ctrl+c` works right)
- More correct logging (and log levels) vs. printing
    - Don't use tqdm if not a TTY? (noisy in logs)
    - Maybe add times to normal print-style logs if not a TTY?
- Decide whether we need to change to address the [redirect playback issue above](
https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/174#discussion_r218604815)
- Collect failed mementos and retry them on a second pass
- Add rate limiting as a simple lock object that is either configured per client or at the thread level, rather than having arbitrary limits built into individual WaybackClient methods.
- Track mementos that are not playback-able so we don’t try them on subsequent runs
    - I’m basically thinking a text file with one URL per line. When starting, we load the file into a `set` that we check before downloading a URL. When stopping, we dump the set back out to the file, but filter out mementos that are more than, say, a week old.

web_monitoring/internetarchive.py
Mr0grog: Note: I fixed the case issue, but have not made changes related to the non-playback-that-redirects-to-a-playback-that-is-still-within-redirect_target_window issue.

Mr0grog:
OK, I wound up implementing better retry/backoff logic (it’s somewhat tightly bound to `WaybackSession`, but we don’t actually need it anywhere else yet and it’s pretty generic/easy to extract) and implementing the second pass for retrying all the unknown failures (sort of a large-scale retry with no special backoff semantics).

In doing so, I made a pretty important discovery about how Wayback recalls mementos: the real problem is that many mementos are just *extraordinarily* slow to load — so much so that parts of Wayback appear to timeout with gateway errors internally and 60 seconds of backing off and retrying is not enough. HOWEVER, those loads also appear to eventually succeed in the background and get cached somewhere in the Wayback infrastructure, such that a request a bit later *will* succeed. (In fact, this is similar to two proposals I made a while back for handling diffs in web-monitoring-db — edgi-govdata-archiving/web-monitoring-db#197 & edgi-govdata-archiving/web-monitoring-db#318 ;) I also wouldn’t be surprised if there are multiple layers of internal caching happening there.

At any rate, all that means we can get a pretty massive speed-boost with a two-pass (or more) system where we fail first on the first pass and retry those failures with a lot more generous settings on the second pass. In a lot of cases, the second pass doesn’t even need those more generous settings, because enough time has gone by since the failure on the first pass and the retry on the second pass. AND we eliminate almost all our unknown failures. 🎉

Mr0grog:
Longer term, the above makes me feel like a better design might be one where a worker load failure simply pops the CDX record back onto the end of the queue. Queue items would probably need a wrapper envelope with info like how many times they’ve been tried or a time not to try before or something in that case. But that seems cleaner than two explicit passes.

Mr0grog:
Short term issues we’ve seen so far that really need resolving:

- When we extract the content of the `<title>` element, we shouldn’t be treating it as plain text (see edgi-govdata-archiving/web-monitoring-db#430). We need to `strip()` it and replace consecutive whitespace with single spaces, as HTML normally works. ([`utils.extract_title()`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/1330bf63157d6831d3308d0d05c27e30e72bb020/web_monitoring/utils.py#L11-L23))

- Default playback window for `get_memento()` needs to be larger (~24 hours, probably) and looser on playback errors in the middle. See this Slack convo: https://edgi.slack.com/archives/G46QNQMHQ/p1539884102000100

- When determining what domains to query from CDX, only query the domain if there are > x pages in it, otherwise query the page URLs directly (where x is probably something like `10`). It turns out we have just a few pages from major third-party sites (for reasons I don’t fully understand), and querying those domains for all the changes Wayback might have seen on them is a huge waste of time and resources.

- Must handle this gzip problem: #309 :(

Mr0grog:
Rebased to fix conflicts, since this branch is getting pretty long in the tooth.

Mr0grog:
Since the above fixes, production runs seem to be a *little* faster (just over 2 hours) and much more stable (I’m not seeing daily scrape runs still hanging around the next day or several days later).

That said…

- There’s still a stupid amount of unplaybackable mementos (~25%-35%), even with the widening of our acceptable timeframes. I guess the reality is that there’s just not much to do about that ¯\\\_(ツ)\_/¯

- We still get a lot of connection errors, so I’m not sure our general strategy around retries is quite right (which is fine, because it’s also just ridiculous code right now). I think we may also need a way to stop/slow down across all threads when we encounter certain errors (or encounter them N times in a row, or something).

- It seems critical to make sure I fix the signal handling problems across threads so that `ctrl+c` does the right thing. I had figured this was part of our stability problem, but we seem to be  stable in practice now, so maybe not.

And of course, the code is a terrifying mess. But this is starting to feel like a thing that at least *works* (as well as it can given limitations with Wayback’s systems and APIs) rather than a not-really-functional mess.

Mr0grog:
Rebased to get “active” page support (#330), so we only import the active pages.

Mr0grog:
The last set of changes around [resetting sessions](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/174/commits/2fa4066a346f99a1c91dff0b25dbaab0a5310add) appears to have dropped the error rate from a consistent 5%-8% down to 1%-2%, so I guess I should keep that and clean it up. It also means this logic around what errors are retry-able within a session is wrong:

https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/24c9fcd8aee78715a5ce75dcb515e4009fb46182/web_monitoring/internetarchive.py#L307-L315

We shouldn’t retry within the session on errors mentioning “failed to establish a new connection.” I should also see if that actual exception object is attached somewhere to the exception type we see in practice — we get a `requests.exceptions.ConnectionError` error that was “caused by” a variety of underlying issues:

```
HTTPConnectionPool(host='web.archive.org', port=
80): Max retries exceeded with url: /web/20190209220025id_/https://www.nsf.gov/news/special_reports/climate/intro_video_degree2.jsp (Caused by NewConnectionError('<ur
llib3.connection.HTTPConnection object at 0x7fc749c68e80>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

Mr0grog:
Looks like I should be able to get urllib3’s underlying error from any requests exception type by doing:

```py
# Because requests_exceptions are OSError/IOError instances where the
# first arg is the underlying error (can’t see another way to get this)
requests_exception.args[0]
```

And that will be a `urllib3.exceptions.MaxRetryError` that wraps a `urllib3.NewConnectionError` — the latter will be stored in the `reason` attribute. So:

```py
requests_exception.args[0].reason
```

Should give us the actual urllib3 error object in this scenario (we should probably be careful to test that each of those things along the chain is present, as this seems a little janky and prone to change). I *think* if we turn off the low-level retry functionality in urllib3 by [providing a custom HTTPAdapter](http://docs.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter) we won’t have the `MaxRetryError` wrapping the `NewConnectionError`, so the above would just be:

```py
requests_exception.args[0]
```

Quick note on customizing low-level retries: the default for requests is to create a Retry object with `0` retries. It turns out that the urllib3 `Retry` class treats total retries of `0` differently than `False`. (I think because it manage counting down by creating new retry instances which decrement the total, so it wants to be able to treat `0` as the end of the countdown, not as “no retries.”) With `0`, it will still wrap underlying exceptions. With `False`, it will not. I think most other behavior is essentially the same, though [it looks like it could possible affect the history](https://github.com/urllib3/urllib3/blob/1.24.1/src/urllib3/util/retry.py#L390).

Mr0grog:
Also of note: we might be able to perform the session resetting directly within the session by closing all the adapters and then calling `self.mount()` with new ones. Need to dig into requests source to make sure: https://github.com/kennethreitz/requests/blob/master/requests/sessions.py

Mr0grog:
Anyway, with the current state of affairs, I am typically seeing:

- ~66% successes
- ~32% un-playback-able mementos
- < 1% CDX entires that don’t actually even have a memento in the first place ¯\\\_(ツ)\_/¯
- ~2% other, unknown HTTP/network errors communicating with Wayback

So that’s a pretty big improvement! I don’t know that I can feasibly shoot for that last number being 0% anymore, so I guess I’m feeling relatively good about where it is now. So I think what’s left here is:

1. Clean up the error type checking logic during retry as noted in the comments above
2. Refactor `cli.py` to at least be slightly less horrifying code
3. Handle signals properly across threads so `ctrl+c` works right

Mr0grog:
Also needs fixing here: we are still failing to capture URLs that should probably be playbackable, like: http://web.archive.org/web/20190222170654id_/https://climate.nasa.gov/vital_signs/

I think the problem above is that there is a redirect that crosses both time and URLs:

```
  http://web.archive.org/web/20190222170654id_/https://climate.nasa.gov/vital_signs/  (is a memento that redirects, 👌)
→ http://web.archive.org/web/20190222170654id_/https://climate.nasa.gov/vital-signs/carbon-dioxide (is a non-memento redirect to both a different time and URL, 👎)
→ http://web.archive.org/web/20190222170518id_/https://climate.nasa.gov/vital-signs/carbon-dioxide/ 
 (Fails to capture; seen as invalid memento)
```

And I believe the logic that is failing us is exactly what I laid out in this comment here: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/24c9fcd8aee78715a5ce75dcb515e4009fb46182/web_monitoring/internetarchive.py#L693-L700

In this case, we added a `/` to the end, which is the same as far as SURT is concerned, but doesn’t pass our very-poor regex. I don’t want to pull in the Archive’s SURT package here because it may do network calls (to do reverse-lookups when there is an IP instead of a domain) and so has possibly high overhead. Maybe I should just trust that whatever URL these redirects are going to will be OK and only check the time component? I’ve been wary in the past about targets staying correct, but I guess maybe I shouldn’t be :\

Mr0grog:
(If following along via e-mail, I edited the above comment somewhat heavily after a little more digging revealed exactly what’s going on here: Wayback really did give us the wrong capture, so we did what was technically the right thing [rejecting it], but still probably the wrong thing in practice.)

Mr0grog:
We’re getting stuck with imports hanging because we aren’t handling exceptions form CDX. Here’s what I got when killing a hanging import:

```
  File "/opt/conda/envs/web-monitoring-etl/bin/wm", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/ubuntu/web-monitoring-processing/scripts/wm", line 6, in <module>
    main()
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/cli.py", line 636, in main
    loop.run_until_complete(command)
  File "/opt/conda/envs/web-monitoring-etl/lib/python3.7/asyncio/base_events.py", line 568, in run_until_complete
    return future.result()
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/cli.py", line 206, in import_ia_db_urls
    dry_run=dry_run)
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/cli.py", line 283, in import_ia_urls
    for wayback_records in toolz.partition_all(2000, all_records):
  File "/opt/conda/envs/web-monitoring-etl/lib/python3.7/site-packages/toolz-0.9.0-py3.7.egg/toolz/itertoolz.py", line 719, in partition_all
    for item in it:
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/cli.py", line 380, in _list_ia_versions_for_urls
    for version in ia_versions:
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/internetarchive.py", line 580, in list_versions
    for version in self.search(**params):
  File "/home/ubuntu/web-monitoring-processing/web_monitoring/internetarchive.py", line 491, in search
    raise UnexpectedResponseFormat(f'Could not parse CDX output: "{text}" (parameters: {final_query})')
web_monitoring.internetarchive.UnexpectedResponseFormat: Could not parse CDX output: "java.lang.IllegalArgumentException: fromKey > toKey" (parameters: {'url': 'http://140.90.23.116/cgi-bin/spg/spgstatus.cgi', 'from': '20190221015153', 'collapse': 'digest', 'showResumeKey': 'true', 'resolveRevisits': 'true'})
```

(This is for page record `4f2d9715-f5a1-4e50-bbae-af0862c1ba5d`, which we should deactivate because the page is dead/moved anyway, but that doesn’t change that we aren’t handling CDX errors well here.)

web_monitoring/cli.py
```diff
@@ -359,6 +381,11 @@ def merge_worker_summaries(summaries):
 
         await uploader
 
+    # Our signal handler is specific to the context of this function, so
+    # reinstate the previous handlers before leaving.
+    signal.signal(signal.SIGINT, sigint_handler)
```
danielballan: Did you consider using a context manager or a `finally` block to ensure the original handlers are reinstated? Here's [an example of doing this with a context](https://github.com/NSLS-II/bluesky/blob/603ef4d8f713bd66ba3bf42d281f4c7903a6b7c5/bluesky/utils.py#L139-L175) but in this case `try/finally` is likely more expedient.

Mr0grog: Oh, good point! I should definitely at least do `finally`, but a context manager (or maybe a decorator for the whole function?) would wrap it up nicely.

```python
with SignalCancel(signal.SIGINT, signal.SIGTERM) as stop:
    ...
```

Mr0grog:
@danielballan I might come back and try and tidy things up slightly over the next few days, but I don’t think I’m going to make any more significant architectural changes at this point, so it’s probably ready for a high-level review.

You might be able to see the places where I got stuck trying to architect a super fancy data pipeline system. I tried a few different approaches and think I came up with some neat stuff, but realized in the end that it was a lot of abstraction for something that would only be used once or twice in the codebase. It didn’t save enough work to be worthwhile.

Anyway, the big pieces here are probably:

- The main work is in `cli.import_ia_urls`. It spawns threads for searching CDX, loading mementos, and sending import requests to web-monitoring-db. The all get connected with `FiniteQueue` instances (see below): https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/3e0ab53f3bf05177caa9e251ba5b8288cb3e9c3a/web_monitoring/cli.py#L388-L452

- `cli.WaybackRecordsImporter` is a kind of crazy megaclass. We spawn a bunch of them to read CDX records off a queue. A class method handles running N of them in parallel and another class method handles re-queueing them for retries. https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/3e0ab53f3bf05177caa9e251ba5b8288cb3e9c3a/web_monitoring/cli.py#L116-L354

    In the super fancy stuff I came up with, I wanted to extract the logic for all the different kinds of errors, but again, it turned into a lot abstraction for little gain.

- `utils.FiniteQueue` is my super-simple take at trying to make a queue that has a defined end and is iterable. It's useful for linking all the threads above together since they can just read to the end. There's something similar in PyPI, but for multiprocessing instead of threading, so it didn't seem like a fit here. https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/3e0ab53f3bf05177caa9e251ba5b8288cb3e9c3a/web_monitoring/utils.py#L109-L152

- Some other useful classes in `utils` for handling signals (similar to what you linked before) and for reference counting.

- `internetarchive.WaybackClient` now depends on a new `WaybackSession` object that handles all the networking/connection/retry/reset logic. It turns out retrying properly needs some care, and while we'd like to keep a session open so we can re-use connections, Wayback frequently craps out on a connection (still a mystery to me), so having `WaybackSession.reset()` is a huge requirement. That single feature made the biggest difference in our error rate by far.

- CDX searches now have some maddening correctness checking, since it turns out the CDX index is chock full of nonsense entries like `http://<<mailto:first.last@pnnl.gov>>/` and even `data:` URLs. 😳

Mr0grog:
(Also, as a bonus, `WaybackSession` now does a lot of retrying by default, and so should be *way* less failure-prone when building docs.)

web_monitoring/__init__.py
```diff
@@ -7,3 +7,8 @@
 
 if os.environ.get('LOG_LEVEL'):
     logging.basicConfig(level=os.environ['LOG_LEVEL'].upper())
+    # Urllib3 logs a warning every time a retry happens. Way too noisy.
+    # TODO: is this the best way to configure? We should probably revisit this.
```
danielballan: You may already be aware that there are `logging.Filter` classes which have access to the complete `Record`. More simply, in recent versions of Python you can pass _any callable_ with the interface `f(record) -> bool` to `log_handler.add_filter`.

web_monitoring/utils.py
```diff
@@ -122,3 +88,197 @@ def get_color_palette():
...
+    def __enter__(self):
+        for signal_type in self.signals:
+            self.old_handlers[signal_type] = signal.getsignal(signal_type)
+            signal.signal(signal_type, self.handler)
+
+        return self
+
+    def __exit__(self, type, value, traceback):
+        for signal_type in self.signals:
+            self.old_handlers[signal_type] = signal.getsignal(signal_type)
```
danielballan: Is the intent to restore the signal handlers to the state they were before? If so, I think this line should be removed. The effect here seems to be stashing the _current_ state of the handlers and then setting them to be exactly the same as they are---a no-op.

web_monitoring/cli.py
```diff
@@ -1,51 +1,455 @@
...
+            if self.failure_queue:
+                self.failure_queue.put(record)
+            else:
+                self.summary['unknown'] += 1
+        except Exception as error:
+            # FIXME: getting read timed out connection errors here...
+            # requests.exceptions.ConnectionError: HTTPConnectionPool(host='web.archive.org', port=80): Read timed out.
+            # TODO: don't count or log (well, maybe DEBUG log) if failure_queue
+            # is present and we are ultimately going to retry.
+            logger.exception(f'  ({type(error)}) {error}; URL: {record.raw_url}')
```
danielballan: A minor thing---using `repr` instead of `str` will give you the type and the args in one:
```suggestion
            logger.exception(f'  {error!r}; URL: {record.raw_url}')
```

web_monitoring/cli.py
```diff
@@ -1,51 +1,455 @@
...
+        Returns
+        -------
+        list of WaybackRecordsWorker
+        """
+        workers = []
+        for i in range(count):
+            worker = cls(*args, **kwargs)
+            workers.append(worker)
+            worker.start()
+
```
danielballan: Random thought: Should this be a `contextlib.contextmanager` that yields `workers` and joins them on exit? If that's not useful in practice anyway, forget about it.

web_monitoring/cli.py
```diff
@@ -1,51 +1,455 @@
...
-                                                       tags=tags,
-                                                       view_url=version.view_url)
-                    for version in wayback.list_versions(url,
-                                                         from_date=from_date,
-                                                         to_date=to_date,
-                                                         skip_repeats=skip_responses))
+    worker_count = worker_count if worker_count > 0 else PARALLEL_REQUESTS
+    unplaybackable = load_unplaybackable_mementos(unplaybackable_path)
 
+    with utils.QuitSignal((signal.SIGINT, signal.SIGTERM)) as stop_event:
```
danielballan: This is clever.

web_monitoring/cli.py
```diff
@@ -60,11 +464,185 @@ def _filter_unchanged_versions(versions):
...
+                for version in ia_versions:
+                    if stop and stop.is_set():
+                        break
+                    if version_filter(version):
+                        yield version
+                    else:
+                        skipped += 1
+                        logger.debug('Skipping URL "%s"', version.url)
+            except ia.BlockedByRobotsError as error:
+                logger.warn(str(error))
```
danielballan: Include the exception type.
```suggestion
                logger.warn(repr(error))
```

web_monitoring/cli.py
```diff
@@ -60,11 +464,185 @@ def _filter_unchanged_versions(versions):
...
+            except ValueError as error:
+                # NOTE: this isn't really an exceptional case; list_versions()
+                # raises ValueError when Wayback has no matching records.
+                # TODO: there should probably be no exception in this case.
+                if 'does not have archived versions' not in str(error):
+                    logger.warn(error)
```
danielballan:
```suggestion
                    logger.warn(repr(error))
```

web_monitoring/cli.py
```diff
@@ -60,11 +464,185 @@ def _filter_unchanged_versions(versions):
...
+            except ia.WaybackException as error:
+                logger.error(f'Error getting CDX data for {url}: {error}')
```
danielballan:
```suggestion
                logger.error(f'Error getting CDX data for {url}: {error!r}')
```

web_monitoring/internetarchive.py
```diff
@@ -97,30 +184,151 @@ def cdx_hash(content):
...
+        maximum = self.retries
+        retries = 0
+        while True:
+            try:
+                result = super().send(*args, **kwargs)
+                if retries >= maximum or not self.should_retry(result):
+                    return result
+            except WaybackSession.handleable_errors as error:
+                if retries >= maximum:
+                    raise WaybackRetryError(retries, total_time, error)
```
danielballan: To retain the full traceback
```suggestion
                    raise WaybackRetryError(retries, total_time, error) from error
```

web_monitoring/__init__.py
Mr0grog: Oh, good point, I guess I should have used that. Will try and rework it using that tonight.

When I wrote this note, though, I was thinking more of the `LOG_LEVEL_SUBCOMPONENT` concept. It doesn’t feel to me like a good abstraction. Maybe better to drop it and do the more targeted filtering regardless of log level settings.

web_monitoring/utils.py
Mr0grog: Huhhhhhhhhhh yeah… not sure how this wound up here! You are right about the intent, and that this wouldn’t work. I think I just need to drop this line and it will be correct. And unset `self.old_handlers` after so it can be garbage collected.

web_monitoring/cli.py
Mr0grog: For this use case, probably not. There’s nothing useful to be done with these workers while they’re running, but we need access to their summaries after they’re done. It seemed usefully flexible to return the completed workers instead of just their summaries in case we added other properties or something.

web_monitoring/cli.py
Mr0grog: In this case, I didn’t do that because this exception always has nice descriptive text: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/98daf6360dea0bec4db4f8a4e3242e07740a490b/web_monitoring/internetarchive.py#L501-L503

Do you think we should still use `repr` here? Should I cut down the error message if so?

Mr0grog: Side note: I guess I should move most of that to the actual error class instead of here where we raise it.

web_monitoring/internetarchive.py
Mr0grog: Ohh, cool, I did not know you could do that! 😲

web_monitoring/utils.py
```diff
@@ -122,3 +88,197 @@ def get_color_palette():
...
+            signal.signal(signal_type, self.old_handlers[signal_type])
+
+
+class QuitSignal(Signal):
+    """
+    A context manager that handles system signals by triggering a
+    `threading.Event` instance, giving your program an opportunity to clean up
+    and shut down gracefully. If the signal is repeated a second time, the
+    process quites immediately.
```
Mr0grog: Typo:

```suggestion
    process quits immediately.
```

web_monitoring/cli.py
danielballan: OK, doesn't seem like including the exception type in the warning would add much here then.

web_monitoring/__init__.py
danielballan: I see. Yes, I agree.

Mr0grog: Hahaha, I tried to reproduce the warning so I could make sure I was filtering properly, and realized this was no longer necessary at all — urllib3’s built-in retry functionality can’t distinguish between a Wayback failure (that should be retried) and a memento of a failure (that should not be retried), so I wound up implementing retry myself. This logging setup must have been from back when I was trying to get native urllib3 retries to work. https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/8c126709f1d4473fbf4a7c5e65f6c8554188e526/web_monitoring/internetarchive.py#L265-L275

Anyway, I just removed all this code. ¯\\\_(ツ)\_/¯

Mr0grog: See 58ea3d80e7ceb4b39b9894fac4e558b8115b3e32

web_monitoring/cli.py
Mr0grog: Argh, went to move the text and remembered why I didn’t originally do that: we could probably end up encountering a `robots.txt` issue in other API besides CDX as well (although I’ve only seen it and know how it’s formatted in CDX).

Any thoughts on how to better design this for that? Should we let the constructor take some extra info like:

```py
raise BlockedByRobotsError(query['url'], f'In CDX search {final_query}')
```

Should we leave it to where the error is handled?

```py
try:
    raise ia.BlockedByRobotsError(query['url'])
except ia.BlockedByRobotsError as error:
    logger.warn(f'CDX search error: {error!r}')
```

Or something else? Leave it as-is?

danielballan: I see. I like the snippet above because it’s the job of the caller to explain what it was doing that caused the error. I don’t think exception’s constructor need be extended to carry arbitrary contextual information.

Mr0grog: OK, I took a crack at that in 91e48ce.

Mr0grog:
@danielballan I think I addressed all the issues you brought up. Want to give it another once-over?

lightandluck:
Huzzah! 🎉

PR #178, opened 2018-03-28
Add rate limiting & backoff for IA HTTP requests

### LINKED ISSUE #177: IA import scripts need rate limiting and retry/backoff semantics
As I’ve started to import a bunch of historical data from the Internet Archive, I’ve found that occasionally we hit gateway timeout errors (504 errors) when getting content. I’m not sure if we’re requesting too fast or if they just have these errors intermittently (but often enough that I am seeing them regularly).

To cover both cases, we should probably add rate limiting options as well as retry capability when resolving a URL.

Mr0grog:
This adds some very simplistic tools to `utils` and makes use of them when loading raw content from Internet Archive:

- `rate_limited()` is a context manager that limits the rate at which the code inside it is executed. The rate is specified as a maximum number of calls per second. It can be a float, so you can do something like 0.5 for 1 call every 2 seconds. A `group` keyword allows you to create non-interfering rate-limited contexts.

- `retryable_request()` is a wrapper around `requests.request()` that will automatically retry failed requests up to a given number of times and optionally back off by progressively greater amounts of time with each retry.

Fixes #177.

We may also want to expose CLI options to tweak these values, but that's not essential and not handled here.

If you aren’t able to provide a good review of the code itself, I’d appreciate feedback on whether this approach makes sense and whether the default values seem good:

- The default rate is 2 requests per second
- The default retry amount is 3 retries, with a backoff of 20 seconds (that’s 6.7 seconds, 10 seconds, and finally 20 seconds of pause before each retry, respectively).

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,5 +6,6 @@ doctr
 ipython
 matplotlib
 numpydoc
+requests-mock
 sphinx
 sphinx_rtd_theme
--- a/web_monitoring/internetarchive.py
+++ b/web_monitoring/internetarchive.py
@@ -334,7 +334,8 @@ def timestamped_uri_to_version(dt, uri, *, url, maintainers=None, tags=None,
     dict : Version
         suitable for passing to :class:`Client.add_versions`
     """
-    res = requests.get(uri)
+    with utils.rate_limited(group='timestamped_uri_to_version'):
+        res = utils.retryable_request('GET', uri)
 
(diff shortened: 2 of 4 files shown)


weatherpattern:
Adding a rate limit and the gradually increasing back off rates seems like a good and reasonable approach. There are fancier ways to do calculate the rates, but I think simpler is better, until we find out we need something more sophisticated (for whatever reason.) @lightandluck  Agree?

PR #185, opened 2018-05-07
Diffing server exception handling

ftsalamp:
Hello everyone! I am the GSoC student selected by the Internet Archive to work on the project of bringing together the Wayback Machine and the scanner software. I wrote some code trying to improve the exception handling in the diffing_server. I also wrote some tests on the code I added.

I'm very excited to work on this project and looking forward to your comments.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -95,7 +95,7 @@ class DiffHandler(tornado.web.RequestHandler):
                         400, reason=str(res_a.error))
                 else:
                     self.send_error(
-                        400, reason=str(res_a.error))
+                        res_a.code, reason=str(res_a.error))
                 return
 
         if res_b.error is not None:
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -61,16 +61,6 @@ XML_PROLOG_PATTERN = re.compile(
 
 client = tornado.httpclient.AsyncHTTPClient()
 
-def handle_request(response):
-    if response.error is not None:
-        try:
-            response.rethrow()
-        except ValueError:
-            print("This is a value error")
-            
-    else:
-         print('Handle request')
-
 class DiffHandler(tornado.web.RequestHandler):
     # subclass must define `differs` attribute
 
@@ -97,26 +87,17 @@ class DiffHandler(tornado.web.RequestHandler):
             try:
                 res_a.rethrow()
             except (ValueError, IOError):
-                pdb.set_trace()
-                data = {}
-                data['code'] = res_a.code
-                data['error'] = str(res_a.error)
-                data['parameter'] = 'a'
-                json_data = json.dumps(data)
-                self.write(json_data)
+                self.send_error(
+                        res_a.code, reason=str(res_a.error))
+                
                 return
 
         if res_b.error is not None:
             try:
                 res_b.rethrow()
             except (ValueError, IOError):
-                pdb.set_trace()
-                data = {}
-                data['code'] = res_b.code
-                data['error'] = str(res_b.error)
-                data['parameter'] = 'b'
-                json_data = json.dumps(data)
-                self.write(json_data)
+                self.send_error(
+                        res_b.code, reason=str(res_b.error))
                 return
 
 
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -2,7 +2,9 @@ import concurrent.futures
 from docopt import docopt
 import hashlib
 import inspect
+import json
 import re
+import sys
 import tornado.gen
 import tornado.httpclient
 import tornado.ioloop
@@ -16,8 +18,6 @@ from web_monitoring.diff_errors import (
 )
 import web_monitoring.html_diff_render
 import web_monitoring.links_diff
-import json
-import sys
 
 # Map tokens in the REST API to functions in modules.
 # The modules do not have to be part of the web_monitoring package.
@@ -69,7 +69,7 @@ class DiffHandler(tornado.web.RequestHandler):
         try:
             func = self.differs[differ]
         except KeyError:
-            self.send_error(404, reason="Diffing method not found.")
+            self.send_error(404, reason = f'Unknown diffing method: `{differ}`. You can get a list of supported differs from the `/` endpoint.')
             return
 
         # If params repeat, take last one. Decode bytes into unicode strings.
@@ -80,36 +80,16 @@ class DiffHandler(tornado.web.RequestHandler):
             b = query_params.pop('b')
         except KeyError:
             self.send_error(
-                        400, reason="Malformed request. A URL parameter is missing.")
+                        400, reason = 'Malformed request. You must provide a URL as the value for both `a` and `b` query parameters.')
             return
         # Fetch server response for URLs a and b.
         res_a, res_b = yield [client.fetch(a, raise_error=False), client.fetch(b, raise_error=False)]
-        
-        #Check if the HTTP requests were successfull and handle exeptions
-        if res_a.error is not None:
-            try:
-                res_a.rethrow()
-            except (ValueError, IOError):
-                if res_a.code == 599:
-                    self.send_error(
-                        400, reason=str(res_a.error))
-                else:
-                    self.send_error(
-                        res_a.code, reason=str(res_a.error))
-                return
 
-        if res_b.error is not None:
-            try:
-                res_b.rethrow()
-            except (ValueError, IOError):
-                if res_b.code == 599:
-                    self.send_error(
-                        400, reason=str(res_b.error))
-                else:
-                    self.send_error(
-                        res_b.code, reason=str(res_b.error))
-                return
-
+        try:
+            self.check_response_for_error(res_a)
+            self.check_response_for_error(res_b)
+        except tornado.httpclient.HTTPError:
+            return
 
         # Validate response bytes against hash, if provided.
         for query_param, res in zip(('a_hash', 'b_hash'), (res_a, res_b)):
@@ -122,7 +102,7 @@ class DiffHandler(tornado.web.RequestHandler):
                 actual_hash = hashlib.sha256(res.body).hexdigest()
                 if actual_hash != expected_hash:
                     self.send_error(
-                        500, reason="Fetched content does not match hash.")
+                        500, reason = 'Fetched content does not match hash.')
                     return
 
         # TODO Add caching of fetched URIs.
@@ -134,11 +114,11 @@ class DiffHandler(tornado.web.RequestHandler):
         except KeyError:
             exceptionMessage = str(sys.exc_info()[1])
             exceptionMessage = exceptionMessage[1:-1]
-            self.send_error(400, reason="Malformed request. "+exceptionMessage)
+            self.send_error(400, reason = 'Malformed request. {}'.format(exceptionMessage))
             return
         except UndecodableContentError:
             exceptionMessage = str(sys.exc_info()[1])
-            self.send_error(422, reason=exceptionMessage)
+            self.send_error(422, reason = exceptionMessage)
             return
         res['version'] = web_monitoring.__version__
         # Echo the client's request unless the differ func has specified
@@ -165,6 +145,23 @@ class DiffHandler(tornado.web.RequestHandler):
         if response['code'] != status_code:
             self.set_status(response['code'])
         self.finish(response)
+
+    def check_response_for_error(self, response):
+        #Check if the HTTP requests were successfull and handle exeptions
+        if response.error is not None:
+            try:
+                response.rethrow()
+            except (ValueError, IOError):
+                #Response code == 599 means that no HTTP response was received.
+                #In this case the error code should become 400 indicating that the error was
+                #raised because of a bad request parameter.
+                if response.code == 599:
+                    self.send_error( 
+                        400, reason = str(response.error))
+                else:
+                    self.send_error(
+                        response.code, reason = str(response.error))
+                raise tornado.httpclient.HTTPError(0)
 
 
 def _extract_encoding(headers, content):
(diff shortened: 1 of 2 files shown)


web_monitoring/diffing_server.py
```diff
@@ -15,6 +16,8 @@
 )
 import web_monitoring.html_diff_render
 import web_monitoring.links_diff
+import json
+import sys
```
Mr0grog: Please alphabetize the imports!

web_monitoring/diffing_server.py
```diff
@@ -67,17 +69,47 @@ def get(self, differ):
         try:
             func = self.differs[differ]
         except KeyError:
-            self.send_error(404)
+            self.send_error(404, reason="Diffing method not found.")
```
Mr0grog: If we’re improving these messages, one important practice we should follow is to, wherever reasonable, include information about what value was wrong and how it could be fixed.

This message would be even better if were something like:

```py
self.send_error(404, reason=f'Unknown diffing method: `{differ}`. You can get a list of supported differs from the `/` endpoint.')
```

web_monitoring/diffing_server.py
```diff
@@ -67,17 +69,47 @@ def get(self, differ):
...
                         self.request.arguments.items()}
-        a = query_params.pop('a')
-        b = query_params.pop('b')
-
+        try:
+            a = query_params.pop('a')
+            b = query_params.pop('b')
+        except KeyError:
+            self.send_error(
+                        400, reason="Malformed request. A URL parameter is missing.")
```
Mr0grog: As above, it’s *really* helpful to be able to say *what* parameter was missing. So maybe something like:

```py
self.send_error(
    400,
    reason='Malformed request. You must provide a URL as the value for both `a` and `b` query parameters.')
```

Even better to be able say specifically whether it was `a`, `b`, or both that were missing, but that’s a lot more code and I think the above is probably good enough :)

web_monitoring/diffing_server.py
```diff
@@ -67,17 +69,47 @@ def get(self, differ):
...
+        res_a, res_b = yield [client.fetch(a, raise_error=False), client.fetch(b, raise_error=False)]
+        
+        #Check if the HTTP requests were successfull and handle exeptions
+        if res_a.error is not None:
+            try:
+                res_a.rethrow()
+            except (ValueError, IOError):
+                if res_a.code == 599:
+                    self.send_error(
+                        400, reason=str(res_a.error))
```
Mr0grog: Could you please add a comment describing why you are re-mapping the `599` code to `400` here? I think I understand why, but this is the sort of thing I think benefits from a clear explanation for others who are working on this code later :)

web_monitoring/diffing_server.py
```diff
@@ -67,17 +69,47 @@ def get(self, differ):
...
+                else:
+                    self.send_error(
+                        res_a.code, reason=str(res_a.error))
+                return
+
+        if res_b.error is not None:
```
Mr0grog: These two error handling blocks (for `res_a` and `res_b`) are exactly the same, so it would be btter to encapsulate this error logic into a function that you can call once for each response.

web_monitoring/diffing_server.py
```diff
@@ -97,7 +129,17 @@ def get(self, differ):
...
+            res = yield executor.submit(caller, func, res_a, res_b, **query_params)
+        except KeyError:
+            exceptionMessage = str(sys.exc_info()[1])
+            exceptionMessage = exceptionMessage[1:-1]
+            self.send_error(400, reason="Malformed request. "+exceptionMessage)
+            return
+        except UndecodableContentError:
+            exceptionMessage = str(sys.exc_info()[1])
+            self.send_error(422, reason=exceptionMessage)
+            return
```
Mr0grog: These errors should already be handled in [`write_error()`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/diffing_server.py#L107-L125). Was that not working for your case?

`write_error` does some important extra work, like:

- differentiating between errors that should be publicly visible and errors that should not (because that could leak security-sensitive information),
- including stack traces _only_ when the server is configured to do so (again, this is a big security concern), and
- formatting the errors in a JSON structure (all output should be JSON here, though we are not perfect at that yet).

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -0,0 +1,55 @@
...
+        response = self.fetch('https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt')
+        with self.assertRaises(UndecodableContentError):
+            df._decode_body(response,'a',False)
+
+    def testFetchUndecodableContent(self):
+        response = self.fetch('http://localhost:'+str(self.get_http_port())+'/html_token?format=json&include=all&a=https://example.org&b=https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt')
+        self.assertEqual(response.code, 422)
+
+def mockDiffingMethod(c_body):
+        return
```
Mr0grog: Please make sure all files end with a blank line!

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -0,0 +1,55 @@
...
+
+class DiffingServerExceptionHandlingTest(AsyncHTTPTestCase):
+
+    def get_app(self):
+        app = df.make_app()
+        app.listen(str(self.get_http_port()))
+        return
+
+    def testInvalidURLaFormat(self):
+        response = self.fetch('http://localhost:'+str(self.get_http_port())+'/html_token?format=json&include=all&a=example.org&b=https://example.org', method='GET')
```
Mr0grog: Can you please use a format string for this? e.g:

```py
port = self.get_http_port()
response = self.fetch(
    f'http://localhost:{port}/html_token?format=json&include=all&a=example.org&b=https://example.org',
    method='GET')
```

Same for all the other tests here.

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertEqual(response.code, 400)
```
Mr0grog: We haven’t done a good job of this everywhere, but we are trying to make sure that this server always responds with JSON. It would be great if these tests didn’t only check the response code, but also checked that they:

1. Have the `application/json` content-type
2. Are parse-able JSON
3. The JSON data is an object with at least the fields `code` (a number) and `error` (a string) present

Same for all the other tests here. You could probably write a method that does those assertions and use it in each of the tests.

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -0,0 +1,55 @@
...
+from web_monitoring.diff_errors import UndecodableContentError
+
+class DiffingServerExceptionHandlingTest(AsyncHTTPTestCase):
+
+    def get_app(self):
+        app = df.make_app()
+        app.listen(str(self.get_http_port()))
+        return
+
+    def testInvalidURLaFormat(self):
```
Mr0grog: Please use underscore case for method names: `def test_invalid_url_a_format`. Same for the rest of these tests.

lightandluck:
Woot! Welcome @ftsalamp! 🎉 That's awesome that you've taken this on and glad that you're here 😀

ftsalamp:
I think I have implemented all the requested changes. Thank you for your help and input @Mr0grog !

web_monitoring/diffing_server.py
ftsalamp: You are absolutely right about this. I think I was just too eager to handle everything, saw this in the stack trace and went ahead and re-handled it without checking the code first. I'm sorry about this.

weatherpattern:
Hello @ftsalamp ! Thank you for your contribution!
Don't forget to add yourself to as a [contributor](https://github.com/edgi-govdata-archiving/web-monitoring).

web_monitoring/diffing_server.py
```diff
@@ -2,11 +2,14 @@
 from docopt import docopt
 import hashlib
 import inspect
+import json
```
Mr0grog: This isn’t used anymore, so you should remove it.

web_monitoring/diffing_server.py
```diff
@@ -2,11 +2,14 @@
...
 import re
+import sys
```
Mr0grog: This isn’t used anymore, so you should remove it.

web_monitoring/diffing_server.py
```diff
@@ -2,11 +2,14 @@
...
 import tornado.gen
 import tornado.httpclient
 import tornado.ioloop
 import tornado.web
+from tornado.stack_context import ExceptionStackContext
```
Mr0grog: This isn’t used anymore, so you should remove it.

web_monitoring/diffing_server.py
```diff
@@ -67,17 +69,27 @@ def get(self, differ):
...
+            return
         # Fetch server response for URLs a and b.
-        res_a, res_b = yield [client.fetch(a), client.fetch(b)]
+        res_a, res_b = yield [client.fetch(a, raise_error=False), client.fetch(b, raise_error=False)]
+
+        try:
+            self.check_response_for_error(res_a)
+            self.check_response_for_error(res_b)
+        except tornado.httpclient.HTTPError:
+            return
```
Mr0grog: I’m pretty sure this is a logic error. In `check_response_for_error()`, we have:

```py
if response.error is not None:
    try:
        response.rethrow()
    except (ValueError, IOError):
        # send back the right error response
        raise tornado.httpclient.HTTPError(0)
    # But we didn’t handle tornado.httpclient.HTTPError errors!
```

If `response.rethrow()` raised a `tornado.httpclient.HTTPError`, it won’t ever get an error response because we didn’t handle it inside `check_response_for_error()`, but we still stopped here in `DiffHandler.get()` as if we did. It basically stops here, but never sends the error message. You can try this by adding a test:

```py
def test_a_is_404(self):
        response = self.fetch(f'/html_token?format=json&include=all&a=http://httpstat.us/404&b=https://example.org')
        print(f'STATUS:: {response.code}')
        print('BODY:::: "%s"' % response.body.decode('utf-8'))
        self.assertEqual(response.code, 400)
        self.json_check(response)
```

…which fails and prints out:

```
STATUS:: 200
BODY:::: ""
```

web_monitoring/diffing_server.py
```diff
@@ -124,6 +136,23 @@ def write_error(self, status_code, **kwargs):
             self.set_status(response['code'])
         self.finish(response)
 
+    def check_response_for_error(self, response):
+        #Check if the HTTP requests were successfull and handle exeptions
+        if response.error is not None:
+            try:
+                response.rethrow()
+            except (ValueError, IOError):
```
Mr0grog: This works, but [`IOError` got merged into `OSError` in Python 3.3](https://docs.python.org/3/library/exceptions.html#IOError), so it might be better to use that instead.

web_monitoring/diffing_server.py
```diff
@@ -124,6 +136,23 @@ def write_error(self, status_code, **kwargs):
...
+                #Response code == 599 means that no HTTP response was received.
+                #In this case the error code should become 400 indicating that the error was
+                #raised because of a bad request parameter.
```
Mr0grog: There is actually a specific HTTP error for this case: `504 GATEWAY TIMEOUT` https://httpstatuses.com/504

Mr0grog:
Sorry I wasn’t able to get back to this yesterday, @ftsalamp!

web_monitoring/diffing_server.py
```diff
@@ -90,7 +102,7 @@ def get(self, differ):
                 actual_hash = hashlib.sha256(res.body).hexdigest()
                 if actual_hash != expected_hash:
                     self.send_error(
-                        500, reason="Fetched content does not match hash.")
+                        500, reason = 'Fetched content does not match hash.')
```
danielballan: Elaborating on @Mr0grog's request for spaces around operators, Python's recommended style does not put spaces around `=` when it's used in a keyword argument like

```py
def f(x=3):
    ...

f(x=2)
```

only when it's used in a assignment like

```py
y = 6
```

You can use a tool like [flake8](http://flake8.pycqa.org/en/latest/) to check the code style and catch these things.

web_monitoring/diffing_server.py
ftsalamp: If no HTTP response is received tornado seems to return 599, according to the docs. http://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.HTTPError
How should we go about this? :)

web_monitoring/diffing_server.py
ftsalamp: Thank you @danielballan ! Having worked for 6 months mostly with Swift, the tool you suggested is a great way to remember Python's coding style.

web_monitoring/diffing_server.py
Mr0grog: Right, Tornado’s doing the correct thing here. Sorry if I was unclear — I didn’t mean we shouldn’t be checking for a 599; I meant our server should return a 504 when it receives a 599. 504 specifically means “this server timed out when asking another server for the data you requested.” It’s a more specific code than 400 in this case.

That said, it would likely be good to make sure here that we really are dealing with a timeout and not something else, like a malformed address that was never attempted (that should probably still be a 400 or maybe a 422). For example, I *think* your `test_invalid_url_a_format` test is triggering a `socket.gaierror` here (subclass of `OSError`), which would be a 400/422 because it’s a bad address, not a timeout.

Mr0grog: Either way, I’m not too worried about sending exactly the right code here; just handling this case better than we previously were is probably an improvement. We can do a separate PR later for more specific and correct codes here.

danielballan:
The tests are failing on a problem that was removed in #186, so I will rebase this on master to incorporate that fix in this PR.

Mr0grog:
Sorry I was so slow to get back to this! Thanks, @ftsalamp.

PR #186, opened 2018-05-10
Run tests in CI without secret credentials

Mr0grog:
This makes CI actually useful for PRs from forks.

1. Treat the versions that we need for tests as fixtures and store them *in the repo* instead of in an untracked cache on disk. This still preserves the ability to automagically grab new ones, but the credentials are only required if you’re doing that, which is extremely rare and should now never happen in CI.

2. Skip the PageFreezer diff tests. We don’t actually use that diff, so requiring everyone to have PF credentials and to run these tests seems unnecessary. (Still keeping the test around so folks with credentials *can* run it if needed.)


Mr0grog:
…and it passed from a fork 🎉

lightandluck:
Nice! 🍾

danielballan:
Smart.

PR #187, opened 2018-05-14
Add tornado debug env value

ftsalamp:
Tornado offers a debug mode that could prove useful during development. In this branch I have added an environment variable in the configuration file that sets the value of tornado's debug mode and use it   when starting the diffing_server.


web_monitoring/diffing_server.py
```diff
@@ -233,7 +234,7 @@ class BoundDiffHandler(DiffHandler):
     return tornado.web.Application([
         (r"/([A-Za-z0-9_]+)", BoundDiffHandler),
         (r"/", IndexHandler),
-    ])
+    ], debug=os.environ.get('TORNADO_DEBUG_MODE', 'False'))
```
Mr0grog: I think it’s a bad idea to read the environment here—instead, `make_app()` should take debug mode as an argument and you should read the value from the environment either in `cli()` or read it into a constant at the top of the file.

It’s a small thing, but that helps:

1. Keep environment reading confined to a single place
2. Make it possible to always instantiate a copy of the server in code with different settings from whatever’s global.

Also, does `'False'` actually work correctly here? I would have expected a non-empty string to be truthy, since this is supposed to take a boolean value.

.env.example
```diff
@@ -16,3 +16,6 @@ PAGE_FREEZER_API_KEY="API_KEY"
 export WEB_MONITORING_DB_STAGING_URL="https://api-staging.monitoring.envirodatagov.org"
 export WEB_MONITORING_DB_STAGING_EMAIL=""
 export WEB_MONITORING_DB_STAGING_PASSWORD=""
+
+# This is used in diffing_server
+export TORNADO_DEBUG_MODE="False"
```
Mr0grog: I think this might be better without “Tornado” in the name—maybe `DIFFING_SERVER_DEBUG` or something. I think it would be good in the future (no need to do this now) to have it affect logger levels in other places, too, and a broader name makes that more reasonable to do.

web_monitoring/diffing_server.py
ftsalamp: Moved it to the beginning of the file and I am now checking better to see if it is true or false. Thank you!

web_monitoring/diffing_server.py
```diff
@@ -56,6 +57,7 @@
 
 
 client = tornado.httpclient.AsyncHTTPClient()
+debug_mode = (os.environ['DIFFING_SERVER_DEBUG'] == 'True')
```
Mr0grog: Three notes here:

- `os.environ['DIFFING_SERVER_DEBUG']` will throw a `KeyError` if `DIFFING_SERVER_DEBUG` is not set. I think this value should be optional, so you should do `os.environ.get('DIFFING_SERVER_DEBUG', 'False')`.

- With environment-provided booleans like this, I usually try and make them case-insensitive—otherwise, it’s really easy for someone to mis-configure the program and not know it. After all, how many times have you spent 15 minutes trying to figure out why something isn’t working only to discover it was a small typo? ;)

- Since this is a constant and we *generally* stick to [PEP 8](https://www.python.org/dev/peps/pep-0008/), please make it upper-case: `DEBUG_MODE`.

So:

```py
DEBUG_MODE = os.environ.get('DIFFING_SERVER_DEBUG', 'False').strip().lower() == 'true'
```

ftsalamp: Thank you for your help!

PR #188, opened 2018-05-15
Cors

ftsalamp:
Hello again! In our project we needed the diffing_server to support CORS. In this branch if the .env file has an ACCESS_CONTROL_ALLOW_ORIGIN_HEADER entry the required headers will be added to the response. If you don't need CORS, remove the entry from the env file and the extra headers will not be added.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,4 +22,3 @@ export DIFFING_SERVER_DEBUG="False"
 
 # This is used in the web_monitoring.diffing_server module.
 export ACCESS_CONTROL_ALLOW_ORIGIN_HEADER="*"
-


.env.example
```diff
@@ -16,3 +16,6 @@ PAGE_FREEZER_API_KEY="API_KEY"
 export WEB_MONITORING_DB_STAGING_URL="https://api-staging.monitoring.envirodatagov.org"
 export WEB_MONITORING_DB_STAGING_EMAIL=""
 export WEB_MONITORING_DB_STAGING_PASSWORD=""
+
+# This is used in the web_monitoring.diffing_server module.
+export ACCESS_CONTROL_ALLOW_ORIGIN_HEADER="*"
```
Mr0grog: Please keep the newline at the end of the file!

web_monitoring/diffing_server.py
```diff
@@ -57,10 +58,24 @@
 
 client = tornado.httpclient.AsyncHTTPClient()
 
+acao_env_value = \
+    os.environ.get('ACCESS_CONTROL_ALLOW_ORIGIN_HEADER', '-1')
```
Mr0grog: Why the string `'-1'` insted of `None` here? That would simplify the code here (`os.environ.get('ACCESS_CONTROL_ALLOW_ORIGIN_HEADER')` — no default needed) and simplify the check in `set_default_headers()`.

Also, please don’t use `acao` in the name here — we try to [only use abbreviations for words that are *primarily* referred to by their abbreviation (e.g. URL, FAQ)](https://github.com/edgi-govdata-archiving/web-monitoring/blob/master/CONTRIBUTING.md#naming). Doing so makes it a lot easier to understand for someone else reading the code later.

web_monitoring/diffing_server.py
```diff
@@ -57,10 +58,24 @@
...
+
 
 class DiffHandler(tornado.web.RequestHandler):
     # subclass must define `differs` attribute
 
+    def set_default_headers(self):
```
Mr0grog: I think this should also be added to `IndexHandler`, right? Maybe it would be better to create a separate base class that includes these CORS features and have `DiffHandler` and `IndexHandler` inherit from that.

web_monitoring/diffing_server.py
```diff
@@ -57,10 +58,24 @@
...
+        if acao_env_value != '-1':
+            self.set_header("Access-Control-Allow-Origin", acao_env_value)
+            self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+            self.set_header('Access-Control-Allow-Methods', 'POST, GET, OPTIONS')
+
+    def options(self, other):
```
Mr0grog: Please remove `other`, since it’s not used.

vbanos:
Thank you for your feedback. Yes, we consider abuse very seriously. It will definitely be protected behind a firewall.

.env.example
ftsalamp: Intellij removes the newline at the end of the file and I have to remember to add two of them. Sorry about this!

danielballan:
Once we get _ours_ behind a firewall, we should also document this concern prominently.

Mr0grog:
@ftsalamp this looks good to me. Would you please rebase and fix the conflicts? Then I’ll go ahead and merge.

weatherpattern:
@ftsalamp nice job on your first merged pull requests!

PR #196, opened 2018-06-20
Really clean up that new links diff algorithm

Mr0grog:
This is a follow-on to #153. This is a complicated last-minute addition and I don’t feel 100% confident the code here is right, though, so I’m making it a separate PR. If anybody has time and is a masochist, I’d _love_ a review of whether this seems correct and sensible. (Code cleanliness and quality review is welcome, too, but it's definitely second to algorithmic sanity and correctness here.)

This "fixes" the last issue I've been able to identify with the overall algorithm in the links diff, and unfortunately makes it really complicated. This last part makes me seriously wonder whether the way I'm abusing SequenceMatcher here is really the right approach. That said, it does seem to work.

The last major hurdle here is the fact that the loose equality checking for Links, which was meant to make SequenceMatcher classify sets of *similar* links as equal, whereupon we could do a second pass with more complicated checking, could fail if the number of similar links in the before or after version was different. Take this example list of links:

|   | A                             | B                                |
| - | ----------------------------- | -------------------------------- |
| 1 | "Pony time!" `ponytime.com/a` | "Pony time!" `ponytime.com/b`    |
| 2 | "Pony time!" `ponytime.com/b` | "Donkey time." `not-ponies.com/` |

What we *want* is to wind up with:

1. Removed "Pony time!" ponytime.com/a
2. Equal   "Pony time!" ponytime.com/b  =  "Pony time!" ponytime.com/b
3. Added   "Donkey time." not-ponies.com/

But because items A1, A2, and B1 are all "equal", SequenceMatcher groups A1 + B1 together as unchanged (good), but then makes A2 a removal. i.e. we got:

1. Equal "Pony time!" ponytime.com/a  = "Pony time!" ponytime.com/b (which we then diff the components of)
2. Removed "Pony time!" ponytime.com/b
3. Added "Donkey time." not-ponies.com/

That's *ok,* but ideally the removal should be A1, while A2 + B1 are the equal group (because A2 + B2 are *exactly* equal). So! This change basically does a first pass after SequenceMatcher to look at all the added/removed items and, if they are roughly equal to the items in the equal section immediately preceeding them (or following), *move* them from the added/removed to the equal section. That then means the equal sections will be disbalanced (i.e. they have more items in the A than B side or vice-versa), so the handling of the pass where we do deeper comparisons on the equal group also gets more complex.

I suspect there might be a better overall solution in finding sequences of roughly equal Links and bundling them up in some other kind of wrapper token *before* sending them to SequenceMatcher. It sounds more manageable, at least. But I am tired, that would be a lot of work to prototype, and what I've already done here seems to work, so pushing this for now.

### DIFF
--- a/web_monitoring/links_diff.py
+++ b/web_monitoring/links_diff.py
@@ -214,8 +214,6 @@ def _assemble_diff(a, b, opcodes):
         List of opcodes from SequenceMatcher that defines what to do with each
         item in the lists of links.
     """
-    # TODO: fix the following situation.
-    #
     # If we have the lists:
     #    A                             B
     #    --------------------------- | ------------------------------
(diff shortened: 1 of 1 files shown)


Mr0grog:
Here’s a real-world example of this problem:

screen shot 2018-06-19 at 5 26 24 pm

Mr0grog:
Note: this has been a nice page to test with: http://localhost:3001/page/b09293a0-0824-45f6-ab0e-3af5a4c30b48/47e04dd6-85a0-4680-9dbd-394fa5026c29..01bf71a3-4855-4228-ba74-31f59f6f39aa

Mr0grog:
Great, will merge this, then.

danielballan:
Any I reason I shouldn't click the big green button on this, @Mr0grog?

Mr0grog:
Nope, I think I just lost track of this later in the day :\

Issue #198, opened 2018-06-25
Text of link diff can include non-visible content (e.g. the source of a <script>)


Mr0grog:
The new links diff uses [`element.text` to get the text of a link](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/links_diff.py#L185-L195), but it turns out that can include non-visible content, like the source of a `<script>` or `<style>` element. It’s possible Beautiful Soup’s [`get_text()`](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#get-text) will do the right thing for us, but if not, we should probably just strip non-visible elements from the DOM when gathering links.

A good example is https://monitoring.envirodatagov.org/page/aaf79858-6cd9-469e-a5ca-a9929afb3283/0f0b4588-fd03-480a-b9b7-47a438ff4090..c254d7e8-cd58-4cf4-acf0-8e17199ea493

HTML:

```html
<div class="site-name">
  <a href="/">
    <script data-pagespeed-no-defer>//<![CDATA[
[...]

PR #200, opened 2018-06-25
Ensure links diff columns are resonable widths

### LINKED ISSUE #199: New links diff allows columns to get too wide/narrow
If the text of a link is huge compared to its URL (or vice-versa) the new table-based links diff display can result in making one column unusably narrow. We should probably make text and URL equal width or at least place some minimum percentage width on each.

See also #198, which can really easily trigger this issue. The same URL (https://monitoring.envirodatagov.org/page/aaf79858-6cd9-469e-a5ca-a9929afb3283/0f0b4588-fd03-480a-b9b7-47a438ff4090..c254d7e8-cd58-4cf4-acf0-8e17199ea493) might be good to test with.

screen shot 2018-06-24 at 8 46 49 pm

While we’re at it, we should probably also make sure the links table spans the full the width of the window (in the opposite case, where the combined text and URL width is too narrow): https://monitoring.envirodatagov.org/page/aaf79858-6cd9-469e-a5ca-a9929afb3283/0cd52f13-0de0-4933-9406-777ac7a795c9..d0132182-8d74-4b8a-8a68-677d08da027f

screen shot 2018-06-24 at 8 48 45 pm

Mr0grog:
The table of links always fills the full width and the columns are split:
- 1.5 em for the change type (+/-/o)
- 50/50 of remaining space for text and URL

Fixes #199.

In the normal case:

screen shot 2018-06-24 at 9 30 38 pm

In the ugly case of #198 (still to be fixed):

screen shot 2018-06-24 at 9 30 21 pm

### DIFF
--- a/web_monitoring/links_diff.py
+++ b/web_monitoring/links_diff.py
@@ -85,6 +85,8 @@ def links_diff_html(a_text, b_text, a_headers=None, b_headers=None,
         }
         .links-list {
             border-collapse: collapse;
+            table-layout: fixed;
+            width: 100%;
         }
         .links-list th {
             background: #f6f6f6;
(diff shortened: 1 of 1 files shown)


Mr0grog:
Note the test failure is IA, which has recently become flaky.

Mr0grog:
Gonna go ahead and merge this since there’s no feedback and it’s reasonably low risk/high benefit.

PR #229, opened 2018-08-08
Use a Python gitignore template, not a node one.

danielballan:
For historical reasons, this repo has a `.gitignore` based on a template for node projects rather than Python ones. This uses a Python template + some additions for our `.env` file with secrets, sphinx's build products and pytest's cache, which is a relatively new artifact of running pytest.


.gitignore
```diff
@@ -1,110 +1,136 @@
...
-# Gradle:
-.idea/gradle.xml
-.idea/libraries
 
-# Mongo Explorer plugin:
-.idea/mongoSettings.xml
 
-## File-based project format:
-*.iws
+### Python template
```
Mr0grog: Looks like you got partially doubled content in here. Looks good otherwise!

Mr0grog:
Also the oddball thing I often add is `scratch.*`, so I can keep a `scratch.md` or `scratch.py` file or whatever in the repo with a stash of random code bits and ideas and samples. Dunno if you’d want to include that or not.

.gitignore
danielballan: Fixed

Mr0grog:
👍 Merging!

PR #232, opened 2018-08-11
Return to being totally lenient on text decoding

### LINKED ISSUE #203: Can/should we be more lenient about decoding errors?
One of our analysts is getting a unicode decoding error with this diff: 
https://monitoring.envirodatagov.org/page/a867fab2-52a2-4057-959d-9e24f6af5d0f/e8559c94-4a18-4637-ade2-e1f4828b6477..61cafbb0-9ba8-49d9-8a6b-5d28f6063ee0

Specifically, that’s a problem in this version: 
https://api.monitoring.envirodatagov.org/api/v0/versions/e8559c94-4a18-4637-ade2-e1f4828b6477

More specifically, the page declares itself to be UTF-8, but does in fact contain an encoding error. The diffing server currently does strict decoding by default (so it just throws an error in this case), but you can change that with the [`ignore_decoding_errors` query param](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/diffing_server.py#L251-L253)… except the UI never actually uses that param.

1. Should the UI always set this to `true`? (Or should we instead default it to `true` instead of `false` here on the processing side? Or something else?)

2. When we ignore errors, we totally ignore them. Should we instead do like most browsers do and replace them with `U+FFFD` (�)? That *should* mean just setting `errors='replace'` instead of `errors='ignore'` [when we actually decode the bytes](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/diffing_server.py#L205-L206)
    - Should we *also* include information about the fact that there was an error decoding in the response envelope in this case, even if it wasn’t a fatal error? e.g. something like:

        ```py
        # In `_decode_body`:
        try:
            return response.body.decode(encoding, errors='strict'), None
        except UnicodeError as error:
            # In the calling function, we'd add the second return value to the response envelope
            return response.body.decode(encoding, errors='replace'), error
        ```

/cc @danielballan

Mr0grog:
This removes the `?ignore_decoding_errors` query param and changes the behavior to never raise exceptions when there are text decoding problems in the diffing server (instead, we just replace the error with the unicode "REPLACEMENT CHARACTER" character, like a browser generally would).

This winds up partially reverting #155 and removing one of our protections against incorrectly trying to diff large binary blobs, but the other content sniffing logic added in #155 is still there and should hopefully be a good enough solution to that problem. The reality is that there's lots of content on the web with actual encoding errors, and we still need to be able to diff it. To do that, the UI would have turned on `ignore_decoding_errors` all the time anyway, so it was probably a pointless switch to have.

Fixes #203.

@danielballan I left the `UndecodableContentError` type in there, but it’s totally vestigial now. Since we’ve waffled back and forth on this, I thought it might be useful to keep, but if you think it’s just confusing cruft, I’ll pull it out.

### DIFF
--- a/web_monitoring/diff_errors.py
+++ b/web_monitoring/diff_errors.py
@@ -3,9 +3,3 @@ class UndiffableContentError(ValueError):
     Raised when the content provided to a differ is incompatible with the
     diff algorithm. For example, if a PDF was provided to an HTML differ.
     """
-
-
-class UndecodableContentError(ValueError):
-    """
-    Raised when the content downloaded for diffing could not be decoded.
-    """
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -11,9 +11,7 @@ import tornado.web
 import traceback
 import web_monitoring
 import web_monitoring.differs
-from web_monitoring.diff_errors import (
-    UndiffableContentError, UndecodableContentError
-)
+from web_monitoring.diff_errors import UndiffableContentError
 import web_monitoring.html_diff_render
 import web_monitoring.links_diff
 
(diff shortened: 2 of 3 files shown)


Mr0grog:
> I left the UndecodableContentError type in there

Nevermind, the linter yelled at me, so I just went ahead and removed it.

PR #233, opened 2018-08-12
Improve element lookups using sets instead of lists

vbanos:
We have many lists of tags like `block_level_tags` and we check if an element is inside them using `name in block_level_tags`.

When we do that and we use a list, Python iterates ALL elements and make an equal number of comparisons. If we use `set()` instead, Python performs just one lookup.

Since this code is used a lot to check all page elements when comparing two pages, there
will be a considerable performance improvement using `sets()`.

Even the `test_html_diff.py` unit tests were a few seconds faster for me.

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -37,7 +37,7 @@ logger = logging.getLogger(__name__)
 # insertion/deletion elements. Instead, break the insertions/deletions in two.
 # TODO: custom elements are iffy here. Maybe include them? (any tag with a `-`
 # in the name)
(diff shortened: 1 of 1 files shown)


vbanos:
circleci seems to fail for a reason irrelevant to my PR but I cannot find a way to rerun it... any help would be appreciated.

danielballan:
I re-ran the tests and they passed.

Mr0grog:
> circleci seems to fail for a reason irrelevant to my PR

Ah, it looks like a third-party service was being flaky. I wrote this up in #236. Sorry for the confusion :(

PR #234, opened 2018-08-12
Optimise web_monitoring.differs._is_invisible_text

vbanos:
This is a busy function used for each element we analyse. Search in `INVISIBLE_TAGS` is done with 1 lookup using a `set()` instead of comparing with every element in the list.

The regex was not compiled.


vbanos:
@Mr0grog yes please include me :)

Mr0grog:
Done! 👍

Issue #220, opened 2018-08-03
Package all the dependencies for conda.


danielballan:
Nothing is particularly hard to compile, but conda packages are convenient and fast, especially as these projects don't ship wheels.

Mr0grog:
This was an outgrowth of #219, right? I was deploying the last few PRs and thought I would switch over the live servers to your newer deploy instructions using `environment.yml`, but realized two things:

1. `environment.yml` is missing version specifiers, which I think makes it a non-starter for a production service. I was going to add them, but…

2. I ran into the same cython issue as @jsnshrmn. Should we add it to the environment file for now, or should I just keep the servers using the old-style deployment for now?

Mr0grog:
(Basically, what should we do for now until you are able to address this issue?)

danielballan:
Stick to the old-style deployment.

We don't want to maintain the pins in two places. I think my original post was from before we pinned versions in requirements.txt. Now that we do that, I think the best way forward is **deploy** from `master` and use pip/venv/requirements.txt. For **interactive use** by analysts, who will get the best benefit from conda, roll actual releases of the package, and produce a proper conda package, which can reflect the version pinning from requirements.txt.

Mr0grog:
👍 I will work up a PR to remove use of the `environment.yml` file from `deploy.md`, then.

PR #246, opened 2018-08-22
Remove environment.yml from deploy instructions

Mr0grog:
Per #220, we don't want to suggest using the Conda environment file in production for now, because it doesn't have pinned dependencies (which can be dangerous).

@danielballan since we never set this up in production using Conda in the first place, I’m curious… in the supervisor config, we have:

```ini
[program:wm-diffing-server-8000]
command=/opt/conda/envs/web-monitoring-processing/bin/wm-diffing-server --port 8000
# env vars, logs, etc
```

Do we need to *activate* the environment here first, or is Conda actually putting a shim that sets up import paths in `$CONDA_PATH/envs/$ENV/bin/wm-diffing-server`?

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -119,29 +119,10 @@ The content of this file should look like:
 ; BTW, Tornado processes are single threaded.
 ; To take advantage of multiple cores, you'll need multiple processes.
 
-[program:wm-diffing-server-8000]
-command=/opt/conda/envs/web-monitoring-processing/bin/wm-diffing-server --port 8000
-stderr_logfile = /var/log/supervisor/tornado-stderr.log
-stdout_logfile = /var/log/supervisor/tornado-stdout.log
-environment=PAGE_FREEZER_API_KEY=<page_freezer_key>
-stopasgroup=true
-
-[program:wm-diffing-server-8001]
-command=/opt/conda/envs/web-monitoring-processing/bin/wm-diffing-server --port 8001
-stderr_logfile = /var/log/supervisor/tornado-stderr.log
-stdout_logfile = /var/log/supervisor/tornado-stdout.log
-environment=PAGE_FREEZER_API_KEY=<page_freezer_key>
-stopasgroup=true
-
-[program:wm-diffing-server-8002]
-command=/opt/conda/envs/web-monitoring-processing/bin/wm-diffing-server --port 8002
-stderr_logfile = /var/log/supervisor/tornado-stderr.log
-stdout_logfile = /var/log/supervisor/tornado-stdout.log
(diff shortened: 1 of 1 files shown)


danielballan:
> Do we need to activate the environment here first, or is Conda actually putting a shim that sets up import paths in `$CONDA_PATH/envs/$ENV/bin/wm-diffing-server`?

Oh, you are correct. That definitely does not work as written. Since `activate && ...` can be brittle, a better approach IMO is to implement an `if __name__ == '__main__'` block in `web_monitorg/diffing_server.py` and then execute it like  `$CONDA_PATH/envs/$ENV/bin/python -m web_monitoring.diffing_server [...]`. This is the approach I have taken in more recent work. See #251.

Mr0grog:
@danielballan if you don’t have time for feedback on #251 in the next couple days, I think I’ll just amend this PR to remove the custom env entirely (so it acts like our current deploy today). Then we can revisit once #251 is worked out.

danielballan:
Finally commented on #251.

Mr0grog:
OK, I set up a new machine from scratch to make sure these instructions were all good. It turns out that if you run the `wm-diffing-server` command from the env, all the paths and python versions get set correctly. Pretty nice, Conda 👍

BUT I found a few other things that needed updating:

- We showed how to install Conda, but not how to set it up in your shell, so you'd never be able to run `conda`. I added a bit about setting it up in `.bashrc`.
- Noted the fact that you probably have to log out and log back in for your session to have the `conda` group you added yourself to.
- The crazy line where we `conda install` all the packages no longer worked because Conda only supports the `==` version specifier. It also doesn’t like spaces. Use a little bash magic to make it happy.
- `python setup.py install` doesn't actually install the two packages we include via git, so we *also* need to `pip install` them! Dunno how this was missed before. Also, this makes installation look pretty ridiculous.
- Finally make use of process groups to reduce the repetition in the supervisor config file. Been meaning to do that since forever.

Look good to you, @danielballan?

danielballan:
>  Also, this makes installation look pretty ridiculous.

Yeah, when I get the bandwidth we should just switch to `venv` at this point. My view on that shifted since we started doing this. As I think I've written elsewhere, the conda installation will be good for analysts using the tools interactively (presumably alongside other tools from the PyData ecosystem) but more trouble than it's worth for deploying the server.

> Look good to you, @danielballan?

Yes. Merging.

danielballan:
But deployment, I'm a fan of the built-in solution. Especially given my bad experience getting rbenv (from which pipenv is based) to work inside ansible.

Mr0grog:
Ok. I strongly suspect it might be worthwhile, but haven’t had a chance to put it through its paces myself, either. Definitely a separate discussion.

PR #251, opened 2018-08-29
Make modules that implement CLIs runnable.

danielballan:
This makes it easier to ensure that a CLI is running against a specific Python
binary.

```
$ python -m web_monitoring.diffing_server --port 9999
Starting server on port 9999
^C
$ python -m web_monitoring.cli -h
Command Line Interface to the web_monitoring Python package

Usage:
wm import ia <url> [--from <from_date>] [--to <to_date>] [options]

<snipped>
```

### DIFF
--- a/web_monitoring/cli.py
+++ b/web_monitoring/cli.py
@@ -108,3 +108,7 @@ Options:
                       from_date=_parse_date_argument(arguments['<from_date>']),
                       to_date=_parse_date_argument(arguments['<to_date>']),
                       skip_unchanged=skip_unchanged)
+
+
+if __name__ == '__main__':
+    main()
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -334,3 +334,7 @@ Options:
     arguments = docopt(doc, version='0.0.1')
     port = int(arguments['<port>'] or 8888)
     start_app(port)
+
+
+if __name__ == '__main__':
+    cli()


Mr0grog:
Actually, isn’t this why we *have* [`scripts/wm`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/scripts/wm) and [`scripts/wm-diffing-server`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/scripts/wm-diffing-server)? They pretty much just do what this code is. I guess I’m struggling to understand what this is for above and beyond that?

e.g. in #246, why not:

```sh
> $CONDA_PATH/envs/$ENV/bin/python -m $CONDA_PATH/envs/$ENV/bin/wm-diffing-server [...]
```

danielballan:
> Does the stuff in [`__init__`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/__init__.py) run when executing a module directly this way?

Yes.

> Actually, isn’t this why we *have* [`scripts/wm`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/scripts/wm) and [`scripts/wm-diffing-server`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/scripts/wm-diffing-server)? They pretty much just do what this code is. I guess I’m struggling to understand what this is for above and beyond that?

That would work too. It is common to do both, effectively making the `scripts` just a convenience. As a user, it is not immediately obvious that some executable I find in `*/bin` is a thing I can run with `python ...` whereas it is clear that `python -m path.to.module` should work, and will definitely be run using _that `python`. I think I picked up this pattern from `jupyterhub`, which does it this way.

So --- not strictly necessary at all, and your proposed solution would have been technically equivalent. I just find this more natural, and there's no harm in supporting both.

Mr0grog:
> It is common to do both

Ah, ok 👍 

Same as before; I have no concerns :)

PR #255, opened 2018-09-06
Be nicer to Internet Archive and also be more strict when loading mementos

### LINKED ISSUE #202: Incorporate Suggestions from IA folks on CDX API methods/tests
The [tests that utilize Internet Archive’s CDX API](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/tests/test_ia.py#L6-L21) have recently become a bit flaky. @vbanos suggested a few improvements we should make here:

1. Add retries. [We already did this for some other Wayback APIs](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/internetarchive.py#L337-L338), but I guess not CDX, so we should incorporate those 😛  (Waiting to hear back on ideal retry/backoff/rate-limit values)

2. Use requests’s [`sessions` feature](http://docs.python-requests.org/en/master/user/advanced/#session-objects).

3. Use the [new?] [`pageSize` parameter](https://github.com/internetarchive/wayback/tree/master/wayback-cdx-server#pagination-api) to test paging over a much smaller query on a URL that doesn’t have a bajillion captures (maybe we could use a date range to further limit it so it doesn't grow over time).

Mr0grog:
Fixes #202, subtask of #174.

The part of this dealing with being more strict on memento loading is required for #174, but it seemed like it would be good to split out into a separate PR so it would be easier to evaluate. Then I decided to go ahead and do the rest of #202.

A few notes and questions on this:

- Does it make sense to raise a `MementoPlaybackError` in `timestamped_uri_to_version()`? Should we be raising something else? Should we not be raising and instead returning `None`?

- The way sessions get passed down the stack feels like it’s more ugly than it needs to be, but I was struggling with finding something better and needing to not close a session that got passed in. Maybe it’s just too late and I’m tired. Would appreciate feedback on this approach.

- I didn’t try to add the [backoff package](https://pypi.org/project/backoff/) here as I suggested in #202 — that seemed like adding more complexity and another big question here. I think we should look at it in a separate PR.

- I added VCR to the IA tests, and in doing so added a subdirectory for each test module to the `cassettes` directory. That seemed like a good idea to avoid any possible collision from test names, but let me know if you think I should just let them all comingle in one place instead.

@danielballan would really appreciate a look from you on this.


danielballan:
> Does it make sense to raise a `MementoPlaybackError` in `timestamped_uri_to_version()`? Should we be raising something else? Should we not be raising and instead returning None?

I definitely prefer an error to `None`. That error seems fine.

> The way sessions get passed down the stack feels like it’s more ugly than it needs to be, 

Per our discussion on Slack, I have pushed some commits to use client classes to hold session state. Feels a bit silly now, but less annoying that passing sessions around manually. And I am guessing it will feel less silly if we grow more methods in the future.

> I think we should look at [backoff] in a separate PR.

:+1:

> That seemed like a good idea to avoid any possible collision from test names, but let me know if you think I should just let them all comingle in one place instead.

Seems like a good practice to follow. Thank you.

Handing back to you @Mr0grog.

web_monitoring/internetarchive.py
```diff
@@ -71,9 +71,8 @@ def original_url_for_memento(memento_url):
     >>> original_url_for_memento('http://web.archive.org/web/20170813195036/https://arpa-e.energy.gov/?q=engage/events-workshops')
     'https://arpa-e.energy.gov/?q=engage/events-workshops'
     """
-    try:
-        url = MEMENTO_URL_PATTERN.match(memento_url).group(1)
-    except:
+    url = MEMENTO_URL_PATTERN.match(memento_url).group(1)
+    if url is None:
```
Mr0grog: Hmmm, this was my lazy way of handling both non-matches (because calling `.group(1)` would raise an `AttributeError`) and bad input types (because calling `.match(not_a_string)` would raise a `TypeError`). Also, I’m not sure this condition can ever be true (if there was a match but nothing in group 1, we’d get `url == ''`).

I think I’m going to change this to split the difference — let the `.match()` call raise a `TypeError` without catching it and recasting as a `ValueError`, then check whether it matched before getting the capture group so we don’t have to have a `try...except` at all.

Seem like a good approach, @danielballan? Otherwise I can flip back to the `try...except` approach but without a *bar* `except`.

Mr0grog: See 4551a9be7c3daa77a9145134bc1415f9ce616305

web_monitoring/internetarchive.py
```diff
@@ -90,162 +90,171 @@ def cdx_hash(content):
...
-
-    Raises
-    ------
-    UnexpectedResponseFormat
-        If the CDX response was not parseable.
-
-    References
-    ----------
-    * https://github.com/internetarchive/wayback/tree/master/wayback-cdx-server
+    session : :class:`requests.Session`, optional
```
Mr0grog: @danielballan totally out of scope here, but this brought things back to mind… what do you think about adding type annotations? (I’ve never used them in Python) Can our doc tools leverage them?

Mr0grog:
@danielballan OK, so the big thing here that I did was add session closing and a context manager API to make that task easier. HOWEVER, when I went to use it in the CLI script, I realized it was really easy to screw up. Take this example:

```py
def get_somewhere_versions():
    with CDXClient() as cdx:
        return cdx.list_versions('http://somewhere.com')

for versions in get_somewhere_versions():
    # do something
```

That will not throw an error, but it will fail to take advantage of the session (because the session gets closed before the iterator even starts iterating). I'm also not sure what ways this might silently *fail*. The session continues to function even though most of its internal state and its pool has been savaged by the `close()` call (and the pool does not get auto repaired; I read through the source and checked).

So I went down a crazy path of trying to make this defer the closing of the session until all the things using it are done. (See the madness in the [`crazy-delayed-session-closing-this-was-a-bad-idea` branch (diff)](https://github.com/edgi-govdata-archiving/web-monitoring-processing/compare/202-be-nicer-to-internet-archive-and-also-be-more-strict-when-loading-mementos...crazy-delayed-session-closing-this-was-a-bad-idea)) But that got complicated, as you can see in that diff. I'm not sure it was worthwhile. Another approach to take might be to subclass session and make it raise an exception if you call `send` after calling `close`, just to make it obvious that you are hitting this subtly confusing aspect of the API. Or we just leave it be.

web_monitoring/internetarchive.py
danielballan: :+1:

web_monitoring/internetarchive.py
danielballan: I have started using them in situation where I want extra code introspection, such as annotating a function's parameters so that I can automatically generate a GUI with the right input types (text box, select box, etc.). I'm not yet won over that they are worth the verbosity for their own sake on small- to medium-sized projects.

danielballan:
> See the madness 

Impressive!

> Another approach to take might be to subclass session and make it raise an exception if you call send after calling close, just to make it obvious that you are hitting this subtly confusing aspect of the API. Or we just leave it be.

I like this. (Why don't Sessions work this way by default?) I am used to this kind of failure, for example:

```py
with h5py.File('data.h5') as f:
    f['dataset'] = ...
# some time later...
f['dataset2'] = ... # BOOM
```

If we find people tripping over it too often, _then_ I would revisit the idea of something a little more magical, along the lines of `LockableSession` to help them avoid having to think. Reasonable?

web_monitoring/internetarchive.py
```diff
@@ -87,149 +92,187 @@ def cdx_hash(content):
...
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.close()
+
+    def close(self):
+        "Close the client's session."
+        self.session.close()
+
+    def search(self, params):
```
danielballan: Now that we have our headers around what the important parameters are, I'd like to explode this function signature to expose them:

```py
def search(resolve_revisits, show_resume_key, **kwargs):
    ...
```

The `**kwargs` provides an escape hatch to pass through any params that we miss or may be added later before our library is updated. My goal here is to make it much more obvious to the caller what the allowed and important parameters are.

Mr0grog: Should we just add all the known options as explicit keywords, then?

Mr0grog:
How dis, @danielballan?

web_monitoring/internetarchive.py
danielballan: Yes, that's my preference. I even have [a rant about why](https://nsls-ii.github.io/scientific-python-cookiecutter/guiding-design-principles.html#complexity-is-always-conserved).

danielballan:
Looks good. Do you mind [adding known options as keyboard args](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/255#discussion_r217473114) before we merge, while we're in the process of thrashing this API? If you'd rather just get it and handle that separately, this is good to merge.

Mr0grog:
Feels like a bit of a pain to decompose those into explicit arguments only to recompose them into a dict for processing, but I guess I did cause me to add type support (so not everything has to be a string). Is 96b7817 about what you were thinking?

Mr0grog:
I do have some misgivings about `showResumeKey` and `resumeKey`, which I kind of want to rename so they make a little more sense in the context of an iterator, but ¯\\\_(ツ)\_/¯

And I left the casing alone except for the ones that are Python keywords and had to have new names. Not sure if that’s very nice, though :\

danielballan:
> Is 96b7817 about what you were thinking?

Yes thank you.

> And I left the casing alone except for the ones that are Python keywords and had to have new names. Not sure if that’s very nice, though :\

Yeah, that's a big ugly, but it's also good to be able to do `search(**params)` where params are valid query parameters (case and all).

PR #269, opened 2018-09-14
Add some examples using the WaybackClient.

danielballan:
Here are two examples for using the `WaybackClient` to do basic analysis. I'm sure we can think up more interesting examples, but these seem like a simple place to start. Hopefully others can build on this.

For those unfamiliar, the ipython sphinx directive:

```rst
.. ipython:: python

   record.date
```

causes actual code execution and renders the output like this:

image


docs/source/cdx_api.rst
```diff
@@ -7,17 +7,127 @@ Python API to Internet Archive Wayback Machine
 Search for historical snapshots of a URL. Download metadata about the snapshots
 and/or the snapshot content itself.
 
-We implement Python clients for the CDX and Memento APIs provided by Wayback
```
danielballan: Note that I didn't _delete_ this; I just moved it way down the pay and rephrased a bit.

Mr0grog:
Haven’t really looked at the rest yet, but why format this in style (or notebook style, I guess), while formatting the API doc examples as:

```py
>>> some_input
Some output
```

Should we change them (or this) to reflect the same example style?

danielballan:
Hmm. Once again, this collaboration has made me pause to reconsider conventions or habits that I have taken as given for years.

Examples in docstrings are commonly read via `help(obj)` or `obj?` interactively. We could use the IPython sphinx directive in a docstring, but then users reading that content interactively would not have the benefit of the output, only the inputs, because docstrings are un-built, i.e. not mediated by sphinx.

Examples in narrative-style tutorial documentation, by contrast, are virtually only read in their published, built form. There, relying on the IPython sphinx directly to generate the output and doc-building time gives us one more layer of defense against unnoticed bit rot by ensuring that the examples have to actually _run_ against a current version of the library on every PR. I picked up this practice from contributing to pandas documentation ([example](https://pandas.pydata.org/pandas-docs/stable/advanced.html)) and have transferred it into day-job projects effectively.

One big reason _not_ to actually run the example code at build time is those example require resources not available in the build context, including API tokens, a cluster of CPUs, or a simply a lot of processing time. Perhaps that's why the examples in [the dask documentation](https://dask.pydata.org/en/latest/dataframe.html), for example, use plain `code-block` examples and include the (expected) output. It seems likely that a lot of the interesting examples we might want to show would require a long time to run and also impose requests on third-party services like IA. In these initial examples, I have intentionally avoided those problems by choosing questions that require only one or external requests to run. But if we carry on with more interesting examples, this strategy will be quite limiting.

So...I think using a mixture of `.. ipython:: python` and `.. code-block:: python` is not inherently a bad approach, but the utility of `.. ipython:: python` goes down when examples are costly to run. In our case, it might make sense to set the standard of using `.. code-block:: python` exclusively. If you agree that that argument seems reasonable, then I'll make the changes.

danielballan:
OTOH: https://twitter.com/danielballan/status/1041676546640162817

Mr0grog:
👍 that all sounds reasonable to me (good reasons for using and not using IPython in the narrative docs and good reasons for always not using it in the API docs examples).

docs/source/cdx_api.rst
```diff
@@ -7,17 +7,128 @@ Python API to Internet Archive Wayback Machine
...
+   record
+
+we can find our answer: Wayback's first snapshot of nasa.gov was in 1996. We
+can use dot access on ``record`` to access the date specifically.
+
+.. ipython:: python
+
+   record.date
+
+Example: How many times does the word 'space' appear on nasa.gov?
```
lightandluck: Change 'space' -> 'mars'

docs/source/cdx_api.rst
```diff
@@ -7,17 +7,128 @@ Python API to Internet Archive Wayback Machine
...
+   content.count('mars')
+
+This is is case-sensitive, so to be more accurate we should convert the content
+to lowercase first.
+
+.. ipython:: python
+
+   content.lower().count('mars')
+
+We picked up a couple occurrences of ``'Mars'`` that the original count missed.
```
lightandluck: Should we make 'Mars' here lowercase to be consistent?

docs/source/cdx_api.rst
```diff
@@ -7,17 +7,128 @@ Python API to Internet Archive Wayback Machine
...
+   visible.lower().count('mars')
+
+Our count of ``visible`` excludes appearances of ``'mars'`` that aren't in the
+visible text, such as in link URLs. Both analyses could be interesting,
+depending on the context.
 
 API Documentation
 =================
 
+The Wayback Machines exposes its data through two different mechanisms,
```
Mr0grog: Typo: “Machines” → “Machine”

danielballan:
Let's stick with the IPython sphinx directive for now, then, but consider trying something else when we start adding heavier examples.

PR #281, opened 2018-10-03
Support list of origin URLs for CORS

vbanos:
ENV var `ACCESS_CONTROL_ALLOW_ORIGIN_HEADER` can be a list of comma
separated URLs like:
```
ACCESS_CONTROL_ALLOW_ORIGIN_HEADER="https://web-beta.archive.org,https://web.archive.org"
```
If an AJAX CORS request has HTTP header `Origin:
https://web.archive.org` the response will include:
`Access-Control-Allow-Origin: https://web.archive.org`.

This is required when passing credentials with a CORS request. Using
just `*` raises a browser error.

In case users keep using `ACCESS_CONTROL_ALLOW_ORIGIN_HEADER="*"`,
nothing changes, the server response still has:
`Access-Control-Allow-Origin: *` and CORS works provided you don't pass
credentials with that.

Added also a relevent unit test.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -77,16 +77,16 @@ access_control_allow_origin_header = \
 
 class BaseHandler(tornado.web.RequestHandler):
 
-    _ALLOWED_ORIGINS = None
-
     def set_default_headers(self):
         if access_control_allow_origin_header is not None:
-            if not self._ALLOWED_ORIGINS:
-                self._ALLOWED_ORIGINS = set([origin.strip() for origin in
-                                             access_control_allow_origin_header.split(',')])
+            if 'allowed_origins' not in self.settings:
+                self.settings['allowed_origins'] = \
+                    set([origin.strip() for origin
+                         in access_control_allow_origin_header.split(',')])
             req_origin = self.request.headers.get('Origin')
             if req_origin:
-                if req_origin in self._ALLOWED_ORIGINS or '*' in self._ALLOWED_ORIGINS:
+                allowed = self.settings.get('allowed_origins')
+                if allowed and (req_origin in allowed or '*' in allowed):
                     self.set_header('Access-Control-Allow-Origin', req_origin)
             self.set_header('Access-Control-Allow-Credentials', 'true')
             self.set_header('Access-Control-Allow-Headers', 'x-requested-with')
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -74,21 +74,20 @@ DEBUG_MODE = os.environ.get('DIFFING_SERVER_DEBUG', 'False').strip().lower() == 
 
 access_control_allow_origin_header = \
     os.environ.get('ACCESS_CONTROL_ALLOW_ORIGIN_HEADER')
-ALLOWED_ORIGINS = set()
-if access_control_allow_origin_header:
-    ALLOWED_ORIGINS = set([orig.strip() for orig in
-                           access_control_allow_origin_header.split(',')])
 
 class BaseHandler(tornado.web.RequestHandler):
 
+    _ALLOWED_ORIGINS = None
+
     def set_default_headers(self):
         if access_control_allow_origin_header is not None:
+            if not self._ALLOWED_ORIGINS:
+                self._ALLOWED_ORIGINS = set([origin.strip() for origin in
+                                             access_control_allow_origin_header.split(',')])
             req_origin = self.request.headers.get('Origin')
-            if req_origin and req_origin in ALLOWED_ORIGINS:
-                self.set_header('Access-Control-Allow-Origin', req_origin)
-            else:
-                self.set_header('Access-Control-Allow-Origin',
-                                access_control_allow_origin_header)
+            if req_origin:
+                if req_origin in self._ALLOWED_ORIGINS or '*' in self._ALLOWED_ORIGINS:
+                    self.set_header('Access-Control-Allow-Origin', req_origin)
             self.set_header('Access-Control-Allow-Credentials', 'true')
             self.set_header('Access-Control-Allow-Headers', 'x-requested-with')
             self.set_header('Access-Control-Allow-Methods', 'GET, OPTIONS')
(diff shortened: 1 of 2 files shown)


web_monitoring/diffing_server.py
```diff
@@ -74,16 +74,23 @@ def __init__(self, url, body, headers):
 
 access_control_allow_origin_header = \
     os.environ.get('ACCESS_CONTROL_ALLOW_ORIGIN_HEADER')
-
+ALLOWED_ORIGINS = set()
+if access_control_allow_origin_header:
+    ALLOWED_ORIGINS = set([orig.strip() for orig in
```
Mr0grog: Please avoid abbreviations, especially when they can be ambiguous for many people (i.e. `orig` → `origin`).

web_monitoring/diffing_server.py
```diff
@@ -74,16 +74,23 @@ def __init__(self, url, body, headers):
 
 access_control_allow_origin_header = \
     os.environ.get('ACCESS_CONTROL_ALLOW_ORIGIN_HEADER')
-
+ALLOWED_ORIGINS = set()
+if access_control_allow_origin_header:
```
Mr0grog: This feels like a lot of logic to just have sitting bare in the module. It might be nicer if this was packaged into a function and that function was called either in `make_app()` or lazily in `BaseHandler.set_default_headers()` (in both cases, you could cache the resulting set in `app.settings`).

web_monitoring/diffing_server.py
```diff
@@ -74,16 +74,23 @@ def __init__(self, url, body, headers):
...
-            self.set_header("Access-Control-Allow-Origin",
-                            access_control_allow_origin_header)
-            self.set_header("Access-Control-Allow-Credentials", "true")
-            self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+            req_origin = self.request.headers.get('Origin')
+            if req_origin and req_origin in ALLOWED_ORIGINS:
+                self.set_header('Access-Control-Allow-Origin', req_origin)
+            else:
+                self.set_header('Access-Control-Allow-Origin',
+                                access_control_allow_origin_header)
```
Mr0grog: If we know `*` won’t actually work in practice with most good browsers, I don’t think it makes any sense for us to continue supporting it here. We should either reflect the request’s origin or not set this header at all (take your pick).

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -166,6 +166,26 @@ def test_check_cors_headers(self):
         assert response.headers.get('Access-Control-Allow-Headers') == 'x-requested-with'
         assert response.headers.get('Access-Control-Allow-Methods') == 'GET, OPTIONS'
 
+    # The processing of the access_control_allow_origin_header to produce
+    # ALLOWED_ORIGINS happens inside the diffing_server module and not on web
+    # app init. As a result, we need to mock patch both of them here to make
+    # the unit test work.
```
Mr0grog: See my notes above about supporting `*` — I think it might make sense to not have the handler code ever look at `access_control_allow_origin_header` and instead only examine the parsed `ALLOWED_ORIGINS` set. If that’s the case, you’d only need to patch one thing here.

web_monitoring/diffing_server.py
vbanos: `*` will work in latest FF and Chrome if you don't pass credentials in your CORS request. That's why I wanted to support it.

If you insist, I will only support request's origin, provided its into the allowed origins conf.
Thanks.

Mr0grog: Ah, I guess that makes some sense as a reason to keep around.

I guess part of me feels like the fact that `*` will only work sometimes is likely to be surprising/confusing for a lot of folks. What if we do this?

```py
req_origin = self.request.headers.get('Origin')
if access_control_allow_origin_header is not None and req_origin is not None:
    if req_origin in ALLOWED_ORIGINS or '*' in ALLOWED_ORIGINS:
        self.set_header('Access-Control-Allow-Origin', req_origin)

    # Set the other headers
```

Is that throwing out an important security feature?

vbanos: If that's OK with you, I've got no problem at all.

Mr0grog: 👍

web_monitoring/diffing_server.py
vbanos: Done.

web_monitoring/diffing_server.py
vbanos: Good point, I moved the logic inside `BaseHandler.set_default_headers()`.

web_monitoring/diffing_server.py
vbanos: Done.

web_monitoring/tests/test_diffing_server_exc_handling.py
vbanos: Updated unit test.

web_monitoring/diffing_server.py
```diff
@@ -75,15 +75,21 @@ def __init__(self, url, body, headers):
...
+    _ALLOWED_ORIGINS = None
+
     def set_default_headers(self):
         if access_control_allow_origin_header is not None:
-            self.set_header("Access-Control-Allow-Origin",
-                            access_control_allow_origin_header)
-            self.set_header("Access-Control-Allow-Credentials", "true")
-            self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+            if not self._ALLOWED_ORIGINS:
+                self._ALLOWED_ORIGINS = set([origin.strip() for origin in
```
Mr0grog: In Tornado, `RequestHandler` instances get constructed for each request, so this caching doesn’t really do any good. You should cache it in the application’s settings dict. You can get that either from `RequestHandler.application.settings` or just `RequestHandler.settings`:

```py
if not self.settings['allowed_origins']:
    self.settings['allowed_origins'] = set(...)
```

More in the docs:
- http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.settings
- http://www.tornadoweb.org/en/stable/web.html#tornado.web.Application.settings

vbanos: If you do `if not self.settings['allowed_origins']:` and `allowed_origins` is not set, we'll have an error.
Its better to do: `if 'allowed_origins' not in self.settings:`

vbanos: Thank you for explaining this, I'm fixing it now.

PR #285, opened 2018-10-08
Compressed responses

vbanos:
Service output wasn't compressed. Using `tornado.web.Application` option
`compress_response=True` enables that and reduces the result size
significantly, improving performance.

Gzip compression is supported from virtually every browser and HTTP
client library so I don't see a reason to make it optional.


vbanos:
https://www.tornadoweb.org/en/stable/_modules/tornado/web.html#Application

PR #286, opened 2018-10-08
Remove redundant JSON spaces

vbanos:
All diff results contain JSON with redundant spaces. That's because
`tornado.escape.json_encode` is used and it doesn't support passing
options to `json.dumps` to trim whitespaces.

We use exactly the same code as `tornado.escape.json_encode` and
`tornado.web.RequestHandler.write` but we trim JSON whitespace.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -1,5 +1,6 @@
 import concurrent.futures
 from docopt import docopt
+import json
 import hashlib
 import inspect
 import os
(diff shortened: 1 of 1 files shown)


danielballan:
I'm 50/50 on this. Do we gain that much compression? Is it worth the hit in readability?

Mr0grog:
I guess one think I can see this being useful for is actually doing the opposite — making output more readable in debug mode :P

```py
if self.settings.get('debug'):
    chunk = json.dumps(chunk, indent=2)
else:
    chunk = json.dumps(chunk, separators=',:')
chunk = chunk.replace('</', '<\\/')
```

vbanos:
From my experience, when a system is in production mode it shouldn't have spaces in JSON but when its in dev mode, its good to have spaces. That's how Flask does it:
https://github.com/pallets/flask/blob/master/flask/json/__init__.py#L257

We try to strip JSON spaces in our systems in general to save space.

vbanos:
What @Mr0grog suggested would be nice but the actual `json.dumps` tornado code lives here
https://github.com/tornadoweb/tornado/blob/master/tornado/escape.py#L67
so its not easy to access the `debug` setting or override the function.

I guess that we should close this PR and skip this feature.

PR #289, opened 2018-10-12
Add text decoding errors back in again, but looser this time

### LINKED ISSUE #287: Protect `html_text_dmp` and `html_source_dmp` with content-type checking and sniffing
The `html_token` diff type currently checks the content-type header and optionally sniffs to make sure it’s reasonably to try and diff its inputs: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/05a0b2550ac5a2982487087c29055600cdf8c214/web_monitoring/html_diff_render.py#L252-L257

The `html_text_dmp` and `html_source_dmp`, which can both go wild and consume all server resources, need similar checks and heuristics:

- For `html_text_dmp`, the semantics can be similar to `html_token`: we want to make sure what we’ve got seems like HTML.
- For `html_source_dmp`, the semantics are different — we just want to make sure it doesn’t seem like binary content (e.g. PDF, image, zip file, Excel file, etc.).

Side note: would this be a useful, generalized binary heuristic?

```py
probably_binary = decoded_string.count('\ufffd') / len(decoded_string) > 0.25
```

---

**Backstory:** This afternoon, after publishing edgi-govdata-archiving/web-monitoring-db#406, I queued up a day’s worth of changes to analyze. Unfortunately, this wound up causing the diffing server to do that terrible thing it does and totally lock up the VM it’s running on (see #154), only now that we are using Kubernetes, it took down every pod on its node :(

Happily, Sentry gets us some pretty good diagnostic info: https://sentry.io/environmental-data-governance-/db-prod/issues/722661459/

It turns out what happened was that some of the (earliest ↔︎ current) diffs involved version pairs where one was a PDF and the other was not. When the earliest version was a PDF that came from Versionista, and when that version was old enough not to have its content-type explicitly specified, it ran across the lame fall-back logic I wrote that [just assumes it might be HTML and goes for it](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/80187bf09ffb886cb241ce30607a8bd440390d6b/app/jobs/analyze_change_job.rb#L91-L105).

When the differ tries to diff a big binary blob, it’s a bad time. Asking it to do so used to be reasonably safe, though: if the content failed to decode as text, we’d raise an exception and stop. But we removed that and made decoding more lenient in #232. I thought we were still safe because we also do sniffing (added in #155), but that’s only on the `html_token` diff.

One important fix here is to make it less likely for these diffs to even try to do their work when they receive obviously binary data, hence this issue. (We *also* need resource limits and other things, but we should still sniff to avoid running amok in the first place.)

/cc @danielballan @jsnshrmn

Mr0grog:
In #287, I noted we need to solve the situation where a text-diffing function receives binary data and described two approaches: expanding the sniffing code we already have and a heuristic for binary data. For now, I wound up going with just the heuristic — it can easily cover all differs without them having to do anything special and it’s relatively painless to implement. The question is whether this heuristic seems sensible.

This essentially adds the decoding errors we removed in #232, but in a more loose form: instead of raising for any decoding error, we only raise if the text winds up being composed of > 25% errors. What we’re really looking for here is stuff that isn’t meant to be text (e.g. PDFs, images, etc.), but if it *was* meant to be text, it’s a good bet we can’t work reasonably with it anyway if it’s so full of errors. As before, you can opt out of this behavior with the `?ignore_decoding_errors` query param.

**Does this seem like a good heuristic? Is 25% the right amount of errors?**

I can also add the sniffing here, but as I started to think more about what it meant to sniff for text (as opposed to sniffing for HTML like we currently do), I kept struggling with whether to have allow lists or block lists of content-types and whether any of it would be better than the heuristic I’ve added here, so I thought I’d just do this heuristic first. Thoughts?

Fixes #287.


Mr0grog:
Rebased to fix conflicts with other PRs that landed on master.

Mr0grog:
@danielballan I’m going to go ahead and merge this so we can give auto-analysis a harder push as we agreed on Friday, but feel free to still review at some point and we can address it in a follow-up PR.

PR #291, opened 2018-10-17
Use cchardet to accelerate bs4

vbanos:
BeautifulSoup may use `chardet` if available to do better encoding
detection.
https://github.com/newvem/beautifulsoup/blob/master/bs4/dammit.py#L18
We don't have it currently in `requirements.txt`.

`cchardet` is a drop in replacement of `chardet` which supports way more
encodings and is far better and faster than `chardet`. Their names are
somehow similar but their technology is totally different. Please check
out their homepages to learn more:

https://pypi.org/project/chardet/

https://pypi.org/project/cchardet/

We just need to do `import cchardet as chardet` before importing `bs4`
to use it, no other code change is necessary.

Performance improves significantly using it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 beautifulsoup4 ==4.6.3
+cchardet ==2.1.4
 python-dateutil ~=2.7.3
 diff_match_patch_python ~=1.0.2
 docopt ~=0.6.2
--- a/web_monitoring/differs.py
+++ b/web_monitoring/differs.py
@@ -1,3 +1,4 @@
+import cchardet as chardet
 from bs4 import BeautifulSoup, Comment
 from diff_match_patch import diff, diff_bytes
 from web_monitoring.utils import get_color_palette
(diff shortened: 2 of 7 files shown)


vbanos:
Circle CI fails with errors like:
```
'cchardet as chardet' imported but unused
```
I understand this because `chardet` is not used inside the current python script but its not right because `chardet` is used inside `bs4` https://github.com/newvem/beautifulsoup/blob/master/bs4/dammit.py#L18 as I already mentioned and the way I do it is the only way to load it as a drop-in replacement for `chardet` AFAIK.

vbanos:
We use `cchardet` in production Wayback Machine instead of `chardet` and its better in every way, that's why I insist trying to load it. Else, I could just add `chardet` in `requirements.txt` and be ready to go.

vbanos:
Thats great @Mr0grog!!
I removed the redundant imports (and a redudant space as well).

PR #292, opened 2018-10-18
Disable SSL cert validation for target sites

vbanos:
Target sites with bad SSL configuration like the examples here:
https://badssl.com/ cannot be checked. Our HTTP client raises an error.

Using the option `validate_cert=False` allows us to access them without
an issue.

HTTP requests become a bit faster too as we don't run the SSL validation
code.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,11 +18,21 @@ export WEB_MONITORING_DB_STAGING_URL="https://api-staging.monitoring.envirodatag
 export WEB_MONITORING_DB_STAGING_EMAIL=""
 export WEB_MONITORING_DB_STAGING_PASSWORD=""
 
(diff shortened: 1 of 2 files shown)


vbanos:
From my experience, its better to disable target SSL certificate validation code in systems fetching data from the general web to be more compatible with target sites.

Mr0grog:
I’m ok with validation being off by default, but my gut says this should be configurable — it seems like there ought to be plenty of reasons to want it on. Any thoughts on this, @danielballan?

vbanos:
OK, I made it optional.

Mr0grog:
Ok, looks like GitHub is functional again, so I’ll merge this in.

Issue #154, opened 2018-01-25
Add resource limits for diffing server


Mr0grog:
During startup, the diffing server should be able  to read an argument or environment variable to set limits on resource usage (or working theory on lockups we’ve seen is that some diff requests cause the process to eat all a machine’s available memory). See https://docs.python.org/3/library/resource.html#resource.setrlimit

danielballan:
Potentially useful for this: https://pypi.org/project/superlance/

danielballan:
I think the right approach here is two layers:

1. Put a limit on the maximum number of bytes of content we will download. The `requests` supports this via the `stream=True` parameters and the `iter_content(chunk_size)` method on responses. This limit will ensure that one giant request can't knock over an otherwise happy server.
2. Use a second process to monitor the memory usage of our server process (a la superlance, above) and kill it if it gets too greedy. This will ensure that a server overwhelmed with many requests can't consume so much system memory that, say, `ssh` breaks.

Using `resource` is apparently not a standard way to do this. Its behavior is platform-dependent, and trusting an overburdened process to police _itself_ is obviously problematic.

jsnshrmn:
> 2. Use a second process to monitor the memory usage of our server process (a la superlance, above) and kill it if it gets too greedy. This will ensure that a server overwhelmed with many requests can't consume so much system memory that, say, ssh breaks.

Alternatively, you could just do a simple check of the system state before job execution. I actually have a super simple shell wrapper I've written to prevent cron jobs from burning up burst credits on aws t2 instances.  Let me see if I can turn that code up as an example.

danielballan:
On the dev call, we concluded that:

* The "single giant request" failure mode (1) is rare but has happened and is easy to address, so we should do it.
* The "overwhelmed by many reasonably-sized requests" failure mode is common, and we should address it putting requests in a queue with a bounded (and configurable) number of workers. Something along the lines of (2) is still worth doing as a second layer of defense.

Mr0grog:
> Alternatively (or additionally), you could just do a simple check of the system state before job execution.

FWIW, it think that’s a great approach for a job-based system. But this is a single long-running HTTP server process.

jsnshrmn:
ah, right, supervisor fires up the processes and they stay up as long as they can. I haven't poked around the code for the webservers yet, but it sounds like I should.

[...]

PR #293, opened 2018-10-18
Use CurlAsyncHTTPClient instead of the default AsyncHTTPClient

vbanos:
According to the docs: https://www.tornadoweb.org/en/stable/httpclient.html
`CurlAsyncHTTPClient` is better than the default `AsyncHTTPClient` in many ways:

* `curl_httpclient` is more likely to be compatible with sites that are
not-quite-compliant with the HTTP spec, or sites that use little-exercised
features of HTTP.
* `curl_httpclient` is faster.

The required code modification to use it is minimal.

`pycurl` is an additional requirement to use it. I also needed to install
system pkg `libcurl4-openssl-dev` to do that in my server (Ubuntu 14.04
LTS).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
 lxml ~=4.2.5
 pandas ~=0.23.4
+pycurl ~=7.43
 raven ~=6.9.0
 requests ~=2.19.1
 toolz ~=0.9.0
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -54,6 +54,7 @@ XML_PROLOG_PATTERN = re.compile(
     b'<?xml\\s[^>]*encoding=[\'"]([^\'"]+)[\'"].*\?>',
     re.IGNORECASE)
 
+tornado.httpclient.AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
 client = tornado.httpclient.AsyncHTTPClient()
 
 


Mr0grog:
The cURL-based client is definitely faster — in another project I worked on for a client earlier this year, we found it to be reliably an order of magnitude or more faster (the regular client commits the sin of reading each byte in pure Python code). I brought this up with @danielballan once many months ago.

That said, we have [“put a limit on the maximum number of bytes of content we will download”](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/154#issuecomment-410534201) as a high priority change we need to make, and it looks like the cURL client doesn’t have support for this like the regular client does. This part of the system has also never been the bottleneck in doing a diff. (in my testing, at least — have you seen differently, @vbanos?)

While this would be a speed improvement, I’m concerned if it prevents us from solving bigger problems that we are actively experiencing (limiting how much we will ever load). @danielballan where are you on this work?

vbanos:
There is a CURL option to limit the max number of bytes to download https://curl.haxx.se/libcurl/c/CURLOPT_MAXFILESIZE.html

vbanos:
@Mr0grog I agree that this part of the code is not a bottleneck but it seemed like a simple improvement to do.

Mr0grog:
> There is a CURL option to limit the max number of bytes to download https://curl.haxx.se/libcurl/c/CURLOPT_MAXFILESIZE.html

From those same docs, though:

> The file size is not always known prior to download, and for such files this option has no effect even if the file transfer ends up being larger than this given limit.

Mr0grog:
Alright, since it seems like @danielballan isn’t going to be able to give any feedback here, I’m going to go ahead and say that we can’t take this as-is. We have two issues that need addressing:

1. Big responses can crash the server
2. HTTP requests are slow

The first is a critical problem, while the second would be a huge improvement, but not nearly as important. We can’t have a solution for (2) that makes (1) hard or impossible to reasonably do.

I would love to switch to `CurlAsyncHTTPClient`, but we *absolutely must* show that we can solve the problem of cutting off big responses with it first. The docs make it pretty clear that `CURLOPT_MAXFILESIZE` is not going to do that for us. This PR needs to include a working, tested solution for that before we can merge it.

Mr0grog:
It looks like what probably has to happen here is:

- Subclass `CurlAsyncHTTPClient`
- Set `CURLOPT_MAXFILESIZE` to handle the ideal case (server reports an accurate response size in header)
- Override either [`_process_queue()`](https://github.com/tornadoweb/tornado/blob/v5.1.1/tornado/curl_httpclient.py#L201) or [`_curl_setup_request`](https://github.com/tornadoweb/tornado/blob/v5.1.1/tornado/curl_httpclient.py#L295) to make sure the buffer write function limits the number of bytes and returns a correct int indicating that not all bytes could be written when we hit that limit (`pycurl` *should* abort if that happens).
    - Could do this by subclassing `io.BytesIO` to have a `write()` function that behaves this way and ensuring the internal `curl.info['buffer']` field uses that kind of object.
    - Or could simply create a new write function that does this in [`_curl_setup_request`](https://github.com/tornadoweb/tornado/blob/v5.1.1/tornado/curl_httpclient.py#L295)
- Add tests for both the ideal case (server responds with correct size headers that are too large) and non-ideal (server doesn’t send that header or sends an acceptable size for that header but then actually writes too many bytes), both of which are extremely common.

danielballan:
Thanks, @Mr0grog. Apologies for low capacity for review the last two weeks.

Mr0grog:
Also bonus points if whatever we do for this can also be repurposed for a PR on Tornado to add the functionality natively ;)

danielballan:
@vbanos Do you have time or interest in addressing the issues enumerated above? If not, I would be happy to pick this back up.

PR #302, opened 2018-10-31
Css diff leak

ftsalamp:
Hello everyone!

We have found some cases where CSS from the webpages that are being compared with the wm-diffing-server conflict with the CSS added to highlight the changes between them.

With this pull request I am aiming to fix the "leaking css rules" by adding to the highlighting elements the wm-diff class.

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -299,10 +299,13 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
 
         color_palette = get_color_palette()
         change_styles.string = f'''
-            ins, ins > * {{text-decoration: none; background-color:
-                {color_palette['differ_insertion']};}}
-            del, del > * {{text-decoration: none; background-color:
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks! I wondered if we might encounter this issue. What do you think of adding a random postfix, so that we can be _sure_ we won't this again? If the random string is generated at import time (not hard-coded) it won't even be possible for a bad actor to generate pathological input that matches our CSS class.

I'm thinking something like:

```py
import uuid

WM_CSS_CLASS = f'wm-diff-{str(uuid.uuid4()[:6]}"
```

danielballan:
Cherry-picked a16d993 onto master and force-pushed to remove the merge commit.

Mr0grog:
FWIW, I do think the right solution at the end of the day is a custom element (like @ftsalamp originally proposed). I think we have to do it in order to fix the way we currently rip pre-existing `ins`/`del` tags out of the document before diffing and it would also neatly sidestep most of the styling class problem. Implementing that is a lot more complicated, though, so this is a hugely useful first step :)

Issue #303, opened 2018-11-05
Investigate why diffing server does not seem to be handling concurrent jobs well


Mr0grog:
We’ve seen lots of issues with the diffing server being unable to handle multiple concurrent jobs well. Specifically, when we have a one diff at a time constantly being performed on a particular server instance, it does not seem to respond very quickly to to more requests.

I know we use `ProcessPoolExecutor` for the diffs so that they don’t have to worry about multitasking cooperatively (many diffs are pure Python code and so won’t work particularly well with either evented code (because they run continuously) or threaded code (because of the GIL). However, a quick glance makes it appear we are probably not using it well:

https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/840ea803237c8aa49d457b48d2e6e7438157c858/web_monitoring/diffing_server.py#L191-L194

- We are created many pools (one per request). We should probably have one executor per server instance.
- We don’t set a number of workers, and it defaults to the number of cores on the machine. I believe on our current setup, that’s two, which is… not a lot of headroom. Ideally, we’d be executing these workloads as jobs when possible (e.g. when responding to the analysis worker instead of a diff requested by a user) and could dedicate a full CPU core to each, but we aren’t there yet (see edgi-govdata-archiving/web-monitoring#62 and edgi-govdata-archiving/web-monitoring-db#406). In the mean time, I think it would certainly be OK to let the system schedule multiple analysis processes per core. That’s half the value of a process pool instead of a thread pool anyhow.

But is that the whole story? Is there anything else that might be contributing to this?

/cc @danielballan

Mr0grog:
In the logs, I’m seeing a whole lot of `concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.` on diffs that are pretty trivial and simple. Definitely something going on here; not sure if that might be related to our creating a whole lot of process pools (as noted above) or something entirely different.

Mr0grog:
(One thought I had was that our resource limits in Kubernetes might be causing it to kill the processes, but I’m reasonably certain we’d see the actual pod get killed instead, not just a child process. We also aren’t seeing diffing pods with short uptime anyway. So that can’t be it.)

Mr0grog:
I’m going to close this now, since the process pool work seems to have more-or-less totally solved this for now.

PR #304, opened 2018-11-05
Use a single, long-lived process pool for diffs

Mr0grog:
The diffing server was constructing and using an independent process pool on every request. It seems unlikely that this is the main cause of our scalability issues (but maybe?), but it still seems like an obviously good thing to do. There's not much point in a pool if you aren't sharing it.

See also #303. I’m not marking this as a solution for it, since I don’t expect that it is. But we can put it in and see whether we start getting fewer errors.


Mr0grog:
Heh, figured that was the case. Can’t believe I had totally missed that we were doing this until now!

PR #308, opened 2018-11-07
Track diff server errors with Sentry

### LINKED ISSUE #306: Set up Sentry.io for the diffing server
I thought we had an issue for this, but clearly not! We are using Sentry in a very manual way to report results in the `ia_healthcheck` script: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/82b383a0796effe7bdc1a65530c7a7fed1ac949b/scripts/ia_healthcheck#L88-L96

…but we *really* need to be using it to track all our exceptions in the diffing server. Reference here: https://docs.sentry.io/platforms/python/

Mr0grog:
We should have done this long ago. Some caveats:

- This has forced me to go understand the various ways Tornado handles errors and how we’re using it only halfway correctly. I’m not trying to fix any of that here, though.

- Sentry automatically finds and captures things sent to Python’s logging framework at the `ERROR` level. Python happens to log out any response with a status code >= 400 that way, which results in spurious, duplicate logging, so I’ve ignored that particular logger.

- Through the same logging mechanism as above (but a different actual logger), Sentry automatically captures detailed exception info. However, if we manually call `self.send_error()`, that doesn’t happen (see point about understand Tornado error handling above).

    These kinds of errors are ones we’re generally expecting, though, so it’s debatable whether we should be capturing them with Sentry. I took the approach of capturing them if we didn’t explicitly mark them as `4xx` errors (i.e. we didn’t think they were the *user’s* fault).

Fixes #306.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
 lxml ~=4.2.5
 pandas ~=0.23.4
-raven ~=6.9.0
+sentry-sdk ~=0.5.3
 requests ~=2.20.0
 toolz ~=0.9.0
 tornado ~=5.1
--- a/scripts/ia_healthcheck
+++ b/scripts/ia_healthcheck
@@ -7,7 +7,7 @@
 
 from datetime import datetime, timedelta
 import random
-import raven
+import sentry_sdk
 import sys
 from web_monitoring import db, internetarchive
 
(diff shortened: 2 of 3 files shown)


Mr0grog:
There’s probably a nice TODO here in abstracting most of this out into a `RequestHandler` we can inherit from, but that would require us to do clean up our code and do things like manipulating errors in `send_error()` (and only do *formatting* in `write_error()`). I figured it was better to ship this soon that debate and polish that abstraction.

PR #325, opened 2018-11-29
Fix missing implementation of `include_latest`

Mr0grog:
We had an argument for this `include_latest`, but it didn't actually do anything. Now it does AND it has a test.


Mr0grog:
Merging, since this hasn’t had any feedback.

Issue #328, opened 2018-12-03
IA healthcheck script should only check active pages


Mr0grog:
Our healthcheck is checking whether all pages in the DB are being maintained in Wayback, but it should really only be checking our *active* pages (that’s a new concept we just added to -db). This probably depends on (and can be solved at the same time as) #327.

PR #330, opened 2018-12-05
Add `active` argument to `db.Client.list_pages()` and use it in the IA healthcheck

### LINKED ISSUE #327: DB API should support new `active` property in `list_pages`
As of edgi-govdata-archiving/web-monitoring-db#440, you can now query page based on whether they are active. We should add support for that to `db.client.list_pages(active=True)`.

Mr0grog:
This is pretty straightforward and adds support for the new `active` DB API argument to our Python wrapper, then uses that in the IA healthcheck script, since we should only really be checking active pages.

Fixes #327, #328.

### DIFF
--- a/scripts/ia_healthcheck
+++ b/scripts/ia_healthcheck
@@ -30,14 +30,14 @@ def sample_monitored_urls(sample_size):
     list of string
     """
     client = db.Client.from_env()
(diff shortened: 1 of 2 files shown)


Mr0grog:
I’m going ahead and merging this, since it’s been a few days.

PR #355, opened 2019-02-08
Support etag headers

### LINKED ISSUE #264: Support ETag headers
Split off from #256, where @Mr0grog suggested:

>  Support ETag headers (maybe a hash of path + query + version?)

This seems straightforward enough. Is this is as simple as, for example

```py
hashlib.sha256(web_monitoring.__version__ + path + str(query_params)).hexdigest()
```

?

jsnshrmn:
I believe this closes #264. This is my first time dealing with Tornado, and I was quite pleased to see that it can do most of the heavy lifting here.

I wasn't exactly sure of the client behavior we wanted to support, so I implemented it for GET and HEAD requests.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -115,14 +115,24 @@ class BaseHandler(tornado.web.RequestHandler):
 class DiffHandler(BaseHandler):
     # subclass must define `differs` attribute
 
-    # Compute our own etags.
+    # Compute our own ETag header values.
+    # We're not actually hashing content for this, since that is expensive.
     def compute_etag(self):
         query_params = {k: v[-1].decode() for k, v in
             self.request.arguments.items()}
-        etag = str('"' + hashlib.sha256(
+        etag = str('W/"' + hashlib.sha256(
             web_monitoring.__version__.encode('utf-8') + self.request.path.encode('utf-8') + str(query_params).encode('utf-8')
         ).hexdigest() + '"').encode('utf-8')
         return etag
+
+    def head(self, differ):
+
+        self.set_etag_header()
+        if self.check_etag_header():
+            self.set_status(304)
+            self.finish()
+            return
+
 
     @tornado.gen.coroutine
     def get(self, differ):
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -135,7 +135,7 @@ class DiffHandler(BaseHandler):
 
         # Uses the "weak validation" directive since we don't guarantee that future
         # responses for the same diff will be byte-for-byte identical.
-        etag = str('W/"' + web_monitoring.utils.hash_content(validation_bytes) + '"').encode('utf-8')
+        etag = f'W/"{web_monitoring.utils.hash_content(validation_bytes)}"'
         return etag
 
     def head(self, differ):
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -92,15 +92,6 @@ access_control_allow_origin_header = \
 class BaseHandler(tornado.web.RequestHandler):
 
     def set_default_headers(self):
-        # Set Etag header if possible.
-        try:
-            etag = hashlib.sha256(
-                web_monitoring.__version__.encode('utf-8') + self.request.path.encode('utf-8') + str(query_params).encode('utf-8')
-            ).hexdigest().encode('utf-8')
-            self.set_header('Etag', etag)
-        except NameError:
-            pass
-
         if access_control_allow_origin_header is not None:
             if 'allowed_origins' not in self.settings:
                 self.settings['allowed_origins'] = \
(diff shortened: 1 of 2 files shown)


jsnshrmn:
Note to self: check etag behavior for error responses.

web_monitoring/diffing_server.py
```diff
@@ -115,8 +115,35 @@ def options(self):
 class DiffHandler(BaseHandler):
     # subclass must define `differs` attribute
 
+    # Compute our own ETag header values.
+    # We're not actually hashing content for this, since that is expensive.
+    def compute_etag(self):
+        query_params = {k: v[-1].decode() for k, v in
+            self.request.arguments.items()}
+        etag = str('W/"' + hashlib.sha256(
```
Mr0grog: Can you note the meaning of `W/“` (weak, which seems like a good choice 👍) and why we chose it here for people who are not deeply familiar with ETag headers?

web_monitoring/diffing_server.py
```diff
@@ -115,8 +115,35 @@ def options(self):
...
+            web_monitoring.__version__.encode('utf-8') + self.request.path.encode('utf-8') + str(query_params).encode('utf-8')
+        ).hexdigest() + '"').encode('utf-8')
```
Mr0grog: This seems slightly unwieldy to read — maybe split up the hash composition and the string?

```py
hash = hashlib.sha256(
    web_monitoring.__version__.encode('utf-8') +
    self.request.path.encode('utf-8') +
    str(query_params).encode('utf-8'))
return f'W/"{hash.hexdigest()}"'.encode('utf-8')
```

(Side note: does this methods actually require us to return bytes instead of a string?)

Mr0grog:
Oh. This should have a test!

Mr0grog:
Also just recalled that we have `utils.hash_content(content_bytes)`.

jsnshrmn:
I added a super basic http get test against local files that just checks for appropriate response codes based on whether the client sends the `If-None-Match` and if the value matches our computed Etag.

web_monitoring/diffing_server.py
```diff
@@ -115,8 +116,47 @@ def options(self):
...
+        # We're not actually hashing content for this, since that is expensive.
+        validation_bytes = str(
+            web_monitoring.__version__
+            + self.request.path
+            + str(self.decode_query_params())
+        ).encode('utf-8')
+
+        # Uses the "weak validation" directive since we don't guarantee that future
+        # responses for the same diff will be byte-for-byte identical.
+        etag = str('W/"' + web_monitoring.utils.hash_content(validation_bytes) + '"').encode('utf-8')
```
Mr0grog: Please use a format string for this. It’s more concise and readable.

```suggestion
        etag = f'W/"{web_monitoring.utils.hash_content(validation_bytes)}"'.encode('utf-8')
```

web_monitoring/tests/test_diffing_server_exc_handling.py
```diff
@@ -52,6 +52,29 @@ def test_both_local(self):
...
+                        f'a=file://{a.name}&b=file://{b.name}',
+                                           headers={'If-None-Match': etag,
+                                           'Accept': 'application/json'})
+                self.assertEqual(warm_response.code, 304)
+
+                mismatch_response = self.fetch('/html_token?format=json&include=all&'
+                        f'a=file://{a.name}&b=file://{b.name}',
+                                           headers={'If-None-Match': 'Stale Value',
+                                           'Accept': 'application/json'})
+                self.assertEqual(mismatch_response.code, 200)
```
Mr0grog: We should also have a test to make sure no Etag header is sent if there was an error on the server, e.g. if the content of the file can’t be decoded, or the content of `b` doesn’t match `b_hash`, or the URL of `b` is unreachable, etc. I don’t know that we need to exhaustively test all these cases, but we should test at least one.

web_monitoring/diffing_server.py
Mr0grog: Also, now that we have tests, I ran them without encoding this return value as bytes and it seems to work fine. Is there any reason we are doing it?

```suggestion
        etag = f'W/"{web_monitoring.utils.hash_content(validation_bytes)}"'
```

jsnshrmn: No good reason at all. This is much nicer.

web_monitoring/tests/test_diffing_server_exc_handling.py
jsnshrmn: I added an assertion for the absence of the etag header in our tests where we assert a 4xx response code for diffs. Tornado appears to be doing the right thing for us since we're using its methods for computing + checking etags and sending errors.

jsnshrmn:
I did some manual testing of the head method, and it doesn't behave exactly has desired, since it doesn't have any of the validation that happens for get.  I should either drop head, or make sure it performs enough checks to throw errors for bad requests. If we're not planning on doing head requests, I'd just drop it, but I thought I'd check the room. @Mr0grog or @danielballan?

Mr0grog:
> [A HEAD request] doesn't have any of the validation that happens for get

We didn’t support HEAD requests before (I had missed the fact that it was new when reviewing originally), so I’m not sure why we would want to add support for them in this PR. If we *did* (they might be nice to verify that and endpoint URL corresponds to an actual diff, but I’m not sure we have a real use-case for that, since you can also just check `/`), that should be a separate PR.

web_monitoring/tests/test_diffing_server_exc_handling.py
Mr0grog: > Tornado appears to be doing the right thing for us since we're using its methods for computing + checking etags and sending errors.

It’s doing the right thing, but not because of that! *All* headers that have been set so far get thrown out when you call `send_error()` (which also gets called under the hood if you throw an exception). It’s more that this is good to have in the tests so that we know if we erroneously return error information by explicitly setting status code and calling `write()`.

PR #380, opened 2019-03-25
Upgrade CI from Python 3.6 to 3.7.

### LINKED ISSUE #341: Upgrade to Python 3.7
Python 3.7 has been out for a while, and offers some useful features that libraries we could be using depend on (e.g. Sentry’s Tornado integration requires it). It would be good to upgrade.

danielballan:
Closes #341 

Question: Is there any reason to continue supporting and testing against 3.6? We must either add a 3.6 job in addition to the 3.7 job or we must update the README to specify a minimum version of 3.7 before this is merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/web-monitoring-processing
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.7
     steps:
       - checkout
       - restore_cache:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Use an official Python runtime as a parent image
-FROM python:3.6-slim
+FROM python:3.7-slim
 MAINTAINER enviroDGI@gmail.com
 
 RUN apt-get update && apt-get install -y --no-install-recommends \


Mr0grog:
> Is there any reason to continue supporting and testing against 3.6?

@vbanos I know you’ve mentioned Python versioning as a concern before. Do you or Internet Archive have any concerns about this one?

> We must either add a 3.6 job in addition to the 3.7 job or we must update the README

Good point! I’d like to hear back from @vbanos above, but maybe the right, quick-ish solution is some YAML hackery to get two jobs?

```yaml
jobs:
  build: &build_tests
    # The same full job definition we currently have goes here

  build_python_3_6:
    <<: *build_tests
    docker:
      - image: circleci/python:3.6

workflows:
  version: 2
  build:
    jobs:
      - build:
          filters:
            branches:
              ignore: release
      - build_python_3_6:
          filters:
            branches:
              ignore: release
```

*Why yes I have been dealing with a lot of cray YAML stuff lately why do you ask

vbanos:
@Mr0grog Python 3.7 requirement won't be a problem.

Thank you for asking!

Mr0grog:
@danielballan should I just merge this, or did you want to implement builds in multiple versions as above?

Mr0grog:
Rebased to re-run tests. Will merge if everything passes.

PR #383, opened 2019-03-29
Create Docker image and k8s cron job for ia-healthcheck

ibuys:

PR #384, opened 2019-03-29
Create Docker image and k8s cron job for ia-archiver

ibuys:


Mr0grog:
Does this actually depend on the work in #383? If so, cool, but it would be really helpful to call that out and explain why this is based on that commit, since it’s not a PR that should be merged independently. Otherwise, for future work, it would be really helpful to isolate what’s actually relevant to the PR.

Otherwise, same basic feedback as in https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/383#issuecomment-478194065

ibuys:
It does not. I'll close this and open up another one to make it more clear.

PR #389, opened 2019-04-02
Create Docker image for k8s cron job to run ia-archiver and ia-health…

ibuys:
This commit adds two Dockerfiles to build images intended to be run in the EDGI Kubernetes cluster. These images replace two cron jobs currently being run on a static server. ia-archiver runs a job to import from archive.org, and ia-healthcheck runs a job to ensure that archive.org is working correctly. 

This commit also updates circle-ci integration to allow the new images to be published to Docker Hub so they can be pulled into the Kubernetes cluster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,14 +54,12 @@ jobs:
           docker login -u $DOCKER_USER -p $DOCKER_PASS
       - run: |
           docker build -t envirodgi/processing:$CIRCLE_SHA1 .
-          docker build -t envirodgi/processing:latest .
-          docker build -t envirodgi/ia-healthcheck:latest -f Dockerfile.ia-healthcheck .
-          docker build -t envirodgi/ia-archiver:latest -f Dockerfile.ia-archiver .
+          docker build -t envirodgi/ia-healthcheck:$CIRCLE_SHA1 -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-etl:$CIRCLE_SHA1 -f Dockerfile.ia-etl .
       - run: |
           docker push envirodgi/processing:$CIRCLE_SHA1
-          docker push envirodgi/processing:latest
-          docker push envirodgi/ia-healthcheck:latest
-          docker push envirodgi/ia-archiver:latest
+          docker push envirodgi/ia-healthcheck:$CIRCLE_SHA1
+          docker push envirodgi/ia-etl:$CIRCLE_SHA1
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,9 +55,13 @@ jobs:
       - run: |
           docker build -t envirodgi/processing:$CIRCLE_SHA1 .
           docker build -t envirodgi/processing:latest .
+          docker build -t envirodgi/ia-healthcheck:latest -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-archiver:latest -f Dockerfile.ia-archiver .
       - run: |
           docker push envirodgi/processing:$CIRCLE_SHA1
           docker push envirodgi/processing:latest
+          docker push envirodgi/ia-healthcheck:latest
+          docker push envirodgi/ia-archiver:latest
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,10 +56,17 @@ jobs:
           docker build -t envirodgi/processing:$CIRCLE_SHA1 .
           docker build -t envirodgi/ia-healthcheck:$CIRCLE_SHA1 -f Dockerfile.ia-healthcheck .
           docker build -t envirodgi/ia-etl:$CIRCLE_SHA1 -f Dockerfile.ia-etl .
+          docker build -t envirodgi/processing:latest .
+          docker build -t envirodgi/ia-healthcheck:latest -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-etl:latest -f Dockerfile.ia-etl .
+
       - run: |
           docker push envirodgi/processing:$CIRCLE_SHA1
           docker push envirodgi/ia-healthcheck:$CIRCLE_SHA1
           docker push envirodgi/ia-etl:$CIRCLE_SHA1
+          docker push envirodgi/processing:latest
+          docker push envirodgi/ia-healthcheck:latest
+          docker push envirodgi/ia-etl:latest
 
 workflows:
   version: 2
--- a/Dockerfile.ia-archiver
+++ b/Dockerfile.ia-archiver
@@ -1,0 +1,28 @@
+# Use an official Python runtime as a parent image
+FROM python:3.6-slim
+MAINTAINER enviroDGI@gmail.com
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git gcc g++
+
+# Set the working directory to /app
+WORKDIR /app
+ARG WEB_MONITORING_DB_URL
+ARG WEB_MONITORING_DB_PASSWORD
+ARG WEB_MONITORING_DB_EMAIL
+
+# Copy the requirements.txt alone into the container at /app
+# so that they can be cached more aggressively than the rest of the source.
+ADD requirements.txt /app
+
+# Install any needed packages specified in requirements.txt
+RUN pip install --trusted-host pypi.python.org -r requirements.txt
+
+# Copy the rest of the source.
+ADD . /app
+
+# Install package.
+RUN pip install .
+
+# Run server on port 80 when the container launches.
+CMD wm import ia-known-pages --from 120 --parallel 35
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -55,9 +55,13 @@ jobs:
       - run: |
           docker build -t envirodgi/processing:$CIRCLE_SHA1 .
           docker build -t envirodgi/processing:latest .
+          docker build -t envirodgi/ia-healthcheck:latest -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-archiver:latest -f Dockerfile.ia-archiver .
```
Mr0grog: Can you also re-tag these as `envirodgi/xyz:$CIRCLE_SHA1`? Not having unique tags makes rollbacks, reproduceability, etc. really hard. I’d go so far as saying the hash-based tag is more important than `latest` if we’re only going to have one.

.circleci/config.yml
```diff
@@ -55,9 +55,13 @@ jobs:
...
       - run: |
           docker push envirodgi/processing:$CIRCLE_SHA1
           docker push envirodgi/processing:latest
+          docker push envirodgi/ia-healthcheck:latest
+          docker push envirodgi/ia-archiver:latest
```
Mr0grog: Same here.

Dockerfile.ia-archiver
```diff
@@ -0,0 +1,28 @@
...
+MAINTAINER enviroDGI@gmail.com
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git gcc g++
+
+# Set the working directory to /app
+WORKDIR /app
+ARG WEB_MONITORING_DB_URL
+ARG WEB_MONITORING_DB_PASSWORD
+ARG WEB_MONITORING_DB_EMAIL
```
Mr0grog: It doesn’t look like these are used (also, I’m worried about somehow baking in credentials here). Can we remove them?

Dockerfile.ia-archiver
```diff
@@ -0,0 +1,28 @@
...
+RUN pip install --trusted-host pypi.python.org -r requirements.txt
+
+# Copy the rest of the source.
+ADD . /app
+
+# Install package.
+RUN pip install .
+
+# Run server on port 80 when the container launches.
+CMD wm import ia-known-pages --from 120 --parallel 35
```
Mr0grog: Could you make env vars for these so they can be tweaked without building a new image?

Dockerfile.ia-healthcheck
```diff
@@ -0,0 +1,31 @@
...
+MAINTAINER enviroDGI@gmail.com
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git gcc g++
+
+# Set the working directory to /app
+WORKDIR /app
+ARG WEB_MONITORING_DB_URL
+ARG WEB_MONITORING_DB_PASSWORD
+ARG WEB_MONITORING_DB_EMAIL
```
Mr0grog: Same notes as in the other dockerfile.

Dockerfile.ia-healthcheck
```diff
@@ -0,0 +1,31 @@
...
+RUN pip install --trusted-host pypi.python.org -r requirements.txt
+
+# Copy the rest of the source.
+ADD . /app
+
+# Install package.
+RUN pip install .
+
+# Make port 80 available to the world outside this container.
+EXPOSE 80
```
Mr0grog: Do we need an open port here?

Mr0grog:
Belated realization: can we rename `ia-archiver` to `ia-etl` or `ia-load` or something? I’m worried about this being confusing w/r/t the `ia-archiver` EC2 instance, which is running a script that causes Internet Archive to snapshot a page *itself* and doesn’t actually move any data into our systems (which is what this does). (Also fair if you think it’s the EC2 instance that needs renaming; or that both do ¯\\\_(ツ)\_/¯)

.circleci/config.yml
ibuys: Yep, good catch, no problem.

Dockerfile.ia-archiver
ibuys: I think those args are needed when running the image, so the application can grab them as environmental variables.

ibuys: I'll double-check.

Dockerfile.ia-healthcheck
ibuys: I don't think so.

.circleci/config.yml
ibuys: Actually, looking through this seems like the only thing we can do is what's already being done with envirodgi/processing, and building two images. One for archiving as an artifact and one labeled "latest", and the latest tag is what's actually pulled in  by web-monitoring-ops.

https://github.com/ibuys/web-monitoring-ops/blob/5f0efef84d0419185e6aabb86f59942b19d15db1/templates/staging/ia-archiver-cronjob.yaml#L13

Am I missing something, or does that sound good to you?

Dockerfile.ia-archiver
ibuys: We use those in the launch of the container from the cronjob definition. Although, after reading through the documentation I'm pretty sure that should be "ENV" and not "ARG". 

https://github.com/ibuys/web-monitoring-ops/blob/5f0efef84d0419185e6aabb86f59942b19d15db1/templates/staging/ia-archiver-cronjob.yaml#L15

.circleci/config.yml
Mr0grog: Right, part of what I was getting at here was that ops *should not use `:latest`.* But that’s a matter for edgi-govdata-archiving/web-monitoring-ops#22, and it looks like you’ve done what needs doing here :)

Dockerfile.ia-archiver
Mr0grog: You don’t need to define `ENV` in the dockerfile to use env vars later. For example, we do all the config for the diff server via env vars in Kubernetes, too (https://github.com/edgi-govdata-archiving/web-monitoring-ops/blob/2095044a5f2ba7d65cc5140f7d8fd01705103440/templates/production/diffing-deployment.yaml#L36-L45), but we don’t need to declare them in `Dockerfile` (https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/Dockerfile).

There are actually quite a few more env vars that can be used, which we document in `.env.example`. I think it’s better to leave them out the dockerfile because otherwise they are likely to get out of sync.

Dockerfile.ia-etl
```diff
@@ -0,0 +1,31 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git gcc g++
+
+# Set the working directory to /app
+WORKDIR /app
+ENV WEB_MONITORING_DB_URL
+ENV WEB_MONITORING_DB_PASSWORD
+ENV WEB_MONITORING_DB_EMAIL
+ENV WEB_MONITORING_FROM
+ENV WEB_MONITORING_PARALLEL
```
Mr0grog: I think these should both have defaults (and those defaults can be the same values you had before) — I just want the vars so that they are customizable without creating a whole new image.

I’m also not sure there’s any need to namespace these as `WEB_MONITORING_*`. The other ones have that in the name because they are referring to an instance of web-monitoring-db, not because they are being namespaced. For example, we also have `VALIDATE_TARGET_CERTIFICATES`, which is not namespaced: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/9a8430c6385d9d5b384417eaeb23d06d99332563/.env.example#L32-L35

.circleci/config.yml
```diff
@@ -54,10 +54,19 @@ jobs:
           docker login -u $DOCKER_USER -p $DOCKER_PASS
       - run: |
           docker build -t envirodgi/processing:$CIRCLE_SHA1 .
+          docker build -t envirodgi/ia-healthcheck:$CIRCLE_SHA1 -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-etl:$CIRCLE_SHA1 -f Dockerfile.ia-etl .
           docker build -t envirodgi/processing:latest .
+          docker build -t envirodgi/ia-healthcheck:latest -f Dockerfile.ia-healthcheck .
+          docker build -t envirodgi/ia-etl:latest -f Dockerfile.ia-etl .
```
Mr0grog: It’s a little more efficient to use `docker tag <source> <target>` here. Not sure why we weren’t doing that before. e.g:

```
docker tag envirodgi/ia-etl:$CIRCLE_SHA1 envirodgi/ia-etl:latest
```

Mr0grog:
OH! You should also feel free to add yourself to [the contributors list in the README](https://github.com/edgi-govdata-archiving/web-monitoring-processing#contributors) (unless you don’t want to be listed).

Mr0grog:
Reassigning this to me to update and bring across the finish line.

Mr0grog:
I started updating this on the [`k8s-cronjobs-v2` branch](https://github.com/edgi-govdata-archiving/web-monitoring-processing/compare/k8s-cronjobs-v2). I’ll eventually close this and create a new PR from that branch.

However, given that the current jobs are actually running based on the code in PR #174, and I intend to squash the history there, I think I’m going to wait and clean that PR up a bit before going further here.

PR #400, opened 2019-05-02
Remove `idea` from exempt labels

lightandluck:
Per https://github.com/edgi-govdata-archiving/overview/issues/231


Mr0grog:
Don’t worry about the docs build failure. I need to find time to figure out a better example to use, since Wayback seems to crap out frequently on that one :\

lightandluck:
Ok, thanks for the info about the error!

Issue #391, opened 2019-04-12
Links diff and HTML diff should ignore Java Servlet Session IDs


Mr0grog:
Java Servlets can keep your session ID in the URL instead of in cookies by tacking `;jsessionid=XYZ` onto the end of a URL. See NOAA NCEI’s Historical Observing Metadata Repository for an example: https://www.ncdc.noaa.gov/homr/

(Note: the behavior only occurs on that site for fresh sessions, so try with a private/incognito browser window.)

```
https://www.ncdc.noaa.gov/homr/api;jsessionid=A2DECB66D2648BFED11FC721FC3043A1
```

Since most captures of a page using this will necessarily have different sessions, we should ignore this part of URLs for links and subresources of a page as if it’s just noise. (This should be adjustable via an argument, but I think ignoring it is the right default.) *Ideally,* the full URL would still appear in the output; it just wouldn’t be highlighted by the differ.

I’m pretty sure there are other (mostly older) systems that do something similar, and we should treat them the same as we discover them.

PR #401, opened 2019-05-04
Add `url_rules` option to HTML diff

ftsalamp:
This PR aims to fix the false positive issue that appears when comparing webpages whose resources might have extra information. This information might be a timestamp in the path of a linked resource (occurring when diffing web archives of pages) or a jsessionid, etc.

In this PR I am implementing the __eq__ functions on the href_token and tag_token classes.
If the URL parameter "strict_urls" is set, and there is a matching value in the REGEX_RULES dictionary of the Strict_url_rule class, the __eq__ function of those objects will take into account the appropriate rule when deciding if the objects are equal or not.

This solves issue #391 

Looking forward to any comments/suggestions on my approach :)

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -256,20 +256,20 @@ class CompoundComparator:
         return False
 
 
-class StrictUrlRule:
+class UrlRules:
     """
-    The StrictUrlRule class represents the the mapping between the
+    The UrlComparisonRule class represents the the mapping between the
     various Comparator classes and the keywords used to match them. This
-    mapping is done inside the CLASS_RULES dictionary.
+    mapping is done inside the RULES dictionary.
     """
-    CLASS_RULES = {'WBM': WaybackUrlComparator,
-                   'jsessionid': ServletSessionUrlComparator,
-                   'UK_WBM': WaybackUkUrlComparator}
+    RULES = {'WBM': WaybackUrlComparator,
+             'jsessionid': ServletSessionUrlComparator,
+             'UK_WBM': WaybackUkUrlComparator}
 
     @classmethod
-    def compare_array(cls, element_array, other_array, comparator):
-        for url_a in element_array:
-            for url_b in other_array:
+    def compare_array(cls, url_list_a, url_list_b, comparator):
+        for url_a in url_list_a:
+            for url_b in url_list_b:
                 if comparator:
                     if comparator.compare(url_a, url_b):
                         return True
@@ -279,8 +279,11 @@ class StrictUrlRule:
 
     @classmethod
     def get_comparator(cls, mode):
+        if not mode:
+            return None
+
         try:
-            comparators = [cls.CLASS_RULES[name.strip()]()
+            comparators = [cls.RULES[name.strip()]()
                            for name in mode.split(',')]
             return CompoundComparator(comparators)
         except KeyError:
@@ -289,7 +292,7 @@ class StrictUrlRule:
 
 def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
                      include='combined', content_type_options='normal',
-                     strict_urls='jsessionid'):
+                     url_rules='jsessionid'):
     """
     HTML Diff for rendering. This is focused on visually highlighting portions
     of a page’s text that have been changed. It does not do much to show how
@@ -342,16 +345,16 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         - `nocheck` ignores the `Content-Type` header but still sniffs.
         - `nosniff` uses the `Content-Type` header but does not sniff.
         - `ignore` doesn’t do any checking at all.
-    strict_urls : string
-        The value of this parameter indicates whether the tag elements and href
-        elements should use special rules when checking their equality.
+    url_rules : string
+        Use specialized rules for comparing URLs in links, images, etc.
+        Possible values are:
         - `jsessionid` ignores Java Servlet session IDs in URLs.
         - `WBM` considers two Wayback Machine links as equivalent if they have
           the same original URL, regardless of each of their timestamps.
         - `UK_WBM` like `WBM`, but for the UK Web Archive (webarchive.org.uk)
-        You can also combine multiple comparison methods with a comma,
-        e.g. `jsessionid,WBM`.
-        Default: `jsessionid`
+        You can also combine multiple comparison rules with a comma,
+        e.g. `jsessionid,WBM`. Use None or an empty string for exact
+        comparisons. (Default: `jsessionid`)
 
     Example
     -------
@@ -365,6 +368,8 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         a_headers,
         b_headers,
         content_type_options)
+
+    comparator = UrlRules.get_comparator(url_rules)
 
     soup_old = html5_parser.parse(a_text.strip() or EMPTY_HTML,
                                   treebuilder='soup', return_root=False)
@@ -381,10 +386,6 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
 
     soup_old = _cleanup_document_structure(soup_old)
     soup_new = _cleanup_document_structure(soup_new)
-
-    comparator = None
-    if strict_urls:
-        comparator = StrictUrlRule.get_comparator(strict_urls)
 
     results, diff_bodies = diff_elements(soup_old.body, soup_new.body, comparator, include)
 
@@ -850,8 +851,7 @@ def flatten_el(el, include_hrefs, skip_tag=False):
                 src_array.append(el_src)
             srcset = el.get('srcset')
             if srcset is not None:
-                srcset_array = srcset.split(',')
-                for src in srcset_array:
+                for src in srcset.split(','):
                     src_array.append(src.split(' ', maxsplit=1)[0])
             yield (TokenType.img, src_array, start_tag(el))
         elif el.tag in undiffable_content_tags:
@@ -980,7 +980,7 @@ class ImgTagToken(tag_token):
 
     def __eq__(self, other):
         if isinstance(other, ImgTagToken):
-            return StrictUrlRule.compare_array(self.data, other.data, self.comparator)
+            return UrlRules.compare_array(self.data, other.data, self.comparator)
         return False
 
     def __hash__(self):
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -840,7 +840,10 @@ def flatten_el(el, include_hrefs, skip_tag=False):
     not returned (just its contents)."""
     if not skip_tag:
         if el.tag == 'img':
-            src_array = [el.get('src')]
+            src_array = []
+            el_src = el.get('src')
+            if el_src is not None:
+                src_array.append(el_src)
             srcset = el.get('srcset')
             if srcset is not None:
                 srcset_array = srcset.split(',')
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -371,8 +371,8 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         - `nosniff` uses the `Content-Type` header but does not sniff.
         - `ignore` doesn’t do any checking at all.
     strict_urls : string
-        The value of the URL parameter that indicates if the tag elements and href
-        elements should use special rules when checking equality,
+        The value of this parameter indicates whether the tag elements and href
+        elements should use special rules when checking their equality.
 
     Example
     -------
@@ -386,8 +386,6 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         a_headers,
         b_headers,
         content_type_options)
-
-    Strict_url_rule.mode = strict_urls
 
     soup_old = html5_parser.parse(a_text.strip() or EMPTY_HTML,
                                   treebuilder='soup', return_root=False)
@@ -678,16 +676,6 @@ class tag_token(DiffToken):
         obj.html_repr = html_repr
         obj.comparator = comparator
         return obj
-
-    def __eq__(self, other):
-        # This equality check aims to apply specific rules to the contents
-        # of the tag element solving false positive cases
-        if Strict_url_rule.mode:
-            return clean_resource(self, other)
-        return str.__eq__(self, other) or str.__eq__(self.lower(), other.lower())
-
-    def __hash__(self):
-        return hash(self.lower())
 
     def __repr__(self):
         return 'tag_token(%s, %s, html_repr=%s, post_tags=%r, pre_tags=%r, trailing_whitespace=%r)' % (
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -206,8 +206,10 @@ class WaybackUrlComparator:
             url_b = self.strip_url_schema(url_b[match_b.end():])
             return url_a == url_b
         else:
-            url_a.strict_urls = None
-            url_b.strict_urls = None
+            if hasattr(url_a, 'strict_urls'):
+                url_a.strict_urls = None
+            if hasattr(url_b, 'strict_urls'):
+                url_b.strict_urls = None
             return url_a == url_b
 
 
@@ -243,6 +245,7 @@ class StrictUrlRule:
                    'jsessionid': ServletSessionUrlComparator,
                    'UKWA': WaybackUkUrlComparator}
 
+    @staticmethod
     def clean_resource_url(element, other, mode):
         try:
             comparator = StrictUrlRule.CLASS_RULES[mode]()
@@ -640,8 +643,10 @@ class tag_token(DiffToken):
             self.pre_tags,
             self.post_tags,
             self.trailing_whitespace)
+
     def html(self):
         return self.html_repr
+
 
 class href_token(DiffToken):
     """ Represents the href in an anchor tag.  Unlike other words, we only
@@ -907,9 +912,34 @@ class SpacerToken(DiffToken):
 # I had some weird concern that I needed to make this token a single word with
 # no spaces, but now that I know this differ more deeply, this is pointless.
 class ImgTagToken(tag_token):
+
+    regex = re.compile(r"(src|srcset)=('|\")(.*?)('|\")")
+
+    def are_any_urls_equal(self, other):
+        strict_urls = self.strict_urls
+        self_array = self[7:-3]
+        self_array = self_array.split(',')
+        other_array = other[7:-3]
+        other_array = other_array.split(',')
+        for url_a in self_array:
+            for url_b in other_array:
+                if strict_urls:
+                    return StrictUrlRule.clean_resource_url(url_a[1:-1], url_b[1:-1], strict_urls)
+                elif url_a == url_b:
+                    return True
+        return False
+
     def __new__(cls, tag, data, html_repr, strict_urls, pre_tags=None,
                 post_tags=None, trailing_whitespace=""):
-        obj = DiffToken.__new__(cls, "\n\nImg:%s\n\n" % data,
+        found_urls = ImgTagToken.regex.findall(html_repr)
+        urls = []
+        for url in found_urls:
+            if url[0] == 'srcset':
+                if ' ' in url[2]:
+                    urls.append(url[2].split(' ')[0])
+                    continue
+            urls.append(url[2])
+        obj = DiffToken.__new__(cls, "\n\nImg:%s\n\n" % str(urls),
                             pre_tags=pre_tags,
                             post_tags=post_tags,
                             trailing_whitespace=trailing_whitespace)
@@ -918,6 +948,14 @@ class ImgTagToken(tag_token):
         obj.html_repr = html_repr
         obj.strict_urls = strict_urls
         return obj
+
+    def __eq__(self, other):
+        if isinstance(other, ImgTagToken):
+            return self.are_any_urls_equal(other)
+        return False
+
+    def __hash__(self):
+        return super.__hash__(self)
 
 
 def _customize_tokens(tokens):
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -191,26 +191,15 @@ MAX_SPACERS = 2500
 
 class WaybackUrlComparator:
     matcher = re.compile(r'web/\d{14}(im_)?/(https?://)?(www.)?')
-    url_schema_matcher = re.compile(r'(https?\:\/\/)?(www\.)?')
-
-    def strip_url_schema(self, url):
-        schema_match = self.url_schema_matcher.search(url)
-        if schema_match:
-            return url[schema_match.end():]
 
     def compare(self, url_a, url_b):
         match_a = self.matcher.search(url_a)
         match_b = self.matcher.search(url_b)
         if match_a and match_b:
-            url_a = self.strip_url_schema(url_a[match_a.end():])
-            url_b = self.strip_url_schema(url_b[match_b.end():])
+            url_a = url_a[match_a.end():]
+            url_b = url_b[match_b.end():]
             return url_a == url_b
-        else:
-            if hasattr(url_a, 'strict_urls'):
-                url_a.strict_urls = None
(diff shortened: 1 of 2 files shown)


web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,16 @@
 MAX_SPACERS = 2500
 
 
+class Strict_url_rule:
+    mode = False
+    REGEX_RULES = {'WBM': r"/web/\d{14}",
+                   'jsessionid': r";jsessionid[^;=]*=([^;]+)",
+                   'UKWA': r"https://www\.webarchive\.org\.uk/wayback/en/archive/\d{14}mp_/"}
```
vbanos: Performance improvement idea:

Use `re.compile(r"/web/\d{14}")` here to compile regex initially and store them in the `REGEX_RULES` dict.

Then, later do:
```
        match_element = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(element)
```
Instead of what you currently do:
```
        match_element = re.search(Strict_url_rule.REGEX_RULES[Strict_url_rule.mode], element)
```
It will be MUCH faster.

web_monitoring/html_diff_render.py
```diff
@@ -681,6 +714,21 @@ def fixup_chunks(chunks):
...
+        match_element = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(element)
+        if match_element:
+            element = element[match_element.end():]
+        match_other = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(other)
+        if match_other:
+            other = other[match_other.end():]
+        return str.__eq__(element, other) or str.__eq__(element.lower(), other.lower())
+    except KeyError:
+        raise KeyError("{} is an invalid strict URL rule."
+                       "".format(Strict_url_rule.mode))
```
Mr0grog: You could simplify this with an f-string:

```py
raise KeyError(f'{StrictUrlRule.mode} is an invalid strict URL rule.')
```

web_monitoring/html_diff_render.py
```diff
@@ -536,6 +549,16 @@ def __new__(cls, tag, data, html_repr, pre_tags=None,
         obj.html_repr = html_repr
         return obj
 
+    def __eq__(self, other):
+        # This equality check aims to apply specific rules to the contents
+        # of the tag element solving false positive cases
+        if Strict_url_rule.mode:
+            return clean_resource_url(self, other)
```
Mr0grog: I think this might wind up being way too broad. First, if `other` is a normal word in the page’s text (perhaps a URL that is fully written out instead of an actual element), you’re going to wind up potentially considering this tag equal to it, which clearly isn’t right.

You should probably start by making sure `other` is also an instance of `tag_token`.

Then it might be a good idea to try and pull out all the URLs or URL-containing attributes to compare. For example, an `` tag might have `src`, but it might also have `srcset`.

Then, since this is a generic comparison instead of one specific to `` tags, you should probably also compare all the other attributes if the URL-containing ones are equal. (Alternatively, you could create a new `ImgToken` type that inherits from `tag_token` and is just for `` tags and implement `__eq__` on that instead.)

web_monitoring/html_diff_render.py
```diff
@@ -681,6 +714,21 @@ def fixup_chunks(chunks):
...
+
+def clean_resource_url(element, other):
+    try:
+        match_element = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(element)
+        if match_element:
+            element = element[match_element.end():]
+        match_other = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(other)
+        if match_other:
+            other = other[match_other.end():]
+        return str.__eq__(element, other) or str.__eq__(element.lower(), other.lower())
```
Mr0grog: What if these elements contain multiple things that look like URLs in them, or contain the same URLs in different orders? Consider that an `` elements `alt` attribute could contain a URL, or that it might have multiple URLs in its `srcset`, or that a refactor of the page might have caused `src` and `srcset` attributes to be in a different order!

All these considerations might be different for different elements, so one improvement here might be to make this function take the *URLs* to compare, rather than the elements, and leave extracting the URLs to the token class, which can be customized for different elements or types of content.

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,16 @@
 MAX_SPACERS = 2500
 
 
+class Strict_url_rule:
+    mode = False
+    REGEX_RULES = {'WBM': re.compile(r"/web/\d{14}"),
+                   'jsessionid': re.compile(r";jsessionid[^;=]*=([^;]+)"),
+                   'UKWA': re.compile(r"https://www\.webarchive\.org\.uk/wayback/en/archive/\d{14}mp_/")}
```
Mr0grog: Could you please include comments or docstrings describing what each of these are meant to do? We both know the purpose here, but it might not be clear to someone else that:

- `WBM` stands for Wayback Machine
- The use-case here is in comparing Wayback mementos, where links will have been rewritten to point to another memento in Wayback instead of the original URL.
- etc.

^ Same sort of issues for each rule here, not just `WBM`. I know why we need to consider `jsessionid`, but someone else might not.

web_monitoring/html_diff_render.py
```diff
@@ -681,6 +714,21 @@ def fixup_chunks(chunks):
 
     return result
 
+
+def clean_resource_url(element, other):
+    try:
+        match_element = Strict_url_rule.REGEX_RULES[Strict_url_rule.mode].search(element)
+        if match_element:
+            element = element[match_element.end():]
```
Mr0grog: I am not sure this is quite right. Looking at everything after the match makes sense for Wayback URLs, but for the `jsessionid` rule, where we only want to cut that one bit of text out of the middle of the URL. There is important URL text before it and potentially an important querystring after it.

A better approach would probably be to have different comparison/rule classes (e.g. one for Wayback, one for UK Wayback, one for Java servlet sessions) that each implement their own logic for comparing two URLs. When you set up the diff, you choose which classes to use when comparing. For example:

```py
class WaybackUrlComparator:
    matcher = re.compile(r'/web/\d{14}')

    def compare(self, url_a, url_b):
        match_a = self.matcher.search(url_a)
        match_b = self.matcher.search(url_b)
        if match_a and match_b:
            return url_a[match_a.end():] == url_b[match_b.end():]
        else:
            return url_a == url_b

class WaybackUkUrlComparator(WaybackUrlComparator):
    matcher = re.compile(r'https://www\.webarchive\.org\.uk/wayback/en/archive/\d{14}mp_/')

class ServletSessionUrlComparator:
    matcher = re.compile(r';jsessionid=[^;]+')

    def compare(self, url_a, url_b):
        match_a = self.matcher.sub('', url_a)
        match_b = self.matcher.sub('', url_b)
        return match_a == match_b

File: web_monitoring/html_diff_render.py [hunk 7]
```diff
@@ -536,6 +549,16 @@ def __new__(cls, tag, data, html_repr, pre_tags=None,
...
 
+    def __eq__(self, other):
+        # This equality check aims to apply specific rules to the contents
+        # of the tag element solving false positive cases
+        if Strict_url_rule.mode:
+            return clean_resource_url(self, other)
+        return str.__eq__(self, other) or str.__eq__(self.lower(), other.lower())
+
+    def __hash__(self):
+        return hash(self.lower())
```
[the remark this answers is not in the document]
ftsalamp: To be honest I got carried away from the __hash__ implementation on the Link class in links_diff.py

I've corrected it!

web_monitoring/html_diff_render.py
```diff
@@ -536,6 +549,16 @@ def __new__(cls, tag, data, html_repr, pre_tags=None,
         obj.html_repr = html_repr
         return obj
 
+    def __eq__(self, other):
+        # This equality check aims to apply specific rules to the contents
+        # of the tag element solving false positive cases
+        if Strict_url_rule.mode:
+            return clean_resource_url(self, other)
+        return str.__eq__(self, other) or str.__eq__(self.lower(), other.lower())
```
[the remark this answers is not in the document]
ftsalamp: Same reason as below. Sorry about this!

web_monitoring/html_diff_render.py
```diff
@@ -255,6 +266,8 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         b_headers,
         content_type_options)
 
+    Strict_url_rule.mode = strict_urls
```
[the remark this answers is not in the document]
ftsalamp: I added the parameter as this solution seems ok for now.

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,16 @@
...
+    mode = False
+    REGEX_RULES = {'WBM': re.compile(r"/web/\d{14}"),
+                   'jsessionid': re.compile(r";jsessionid[^;=]*=([^;]+)"),
+                   'UKWA': re.compile(r"https://www\.webarchive\.org\.uk/wayback/en/archive/\d{14}mp_/")}
+
+
 def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
-                     include='combined', content_type_options='normal'):
+                     include='combined', content_type_options='normal',
+                     strict_urls=False):
```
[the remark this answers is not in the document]
ftsalamp: You're right!

web_monitoring/html_diff_render.py
```diff
@@ -241,6 +249,9 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
         - `nocheck` ignores the `Content-Type` header but still sniffs.
         - `nosniff` uses the `Content-Type` header but does not sniff.
         - `ignore` doesn’t do any checking at all.
+    strict_urls : string
+        The value of the URL parameter that indicates if the tag elements and href
+        elements should use special rules when checking equality,
```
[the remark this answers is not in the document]
ftsalamp: Done! :)

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,16 @@
 MAX_SPACERS = 2500
 
 
+class Strict_url_rule:
```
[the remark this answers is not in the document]
ftsalamp: Fixed!

web_monitoring/html_diff_render.py
```diff
@@ -681,6 +714,21 @@ def fixup_chunks(chunks):
 
     return result
 
+
+def clean_resource_url(element, other):
```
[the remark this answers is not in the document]
ftsalamp: Moved

web_monitoring/html_diff_render.py
ftsalamp: This is how a KeyError is raised in diffing_server.py but your suggestion is cleaner.

web_monitoring/html_diff_render.py
ftsalamp: Added

web_monitoring/html_diff_render.py
ftsalamp: Done

web_monitoring/html_diff_render.py
```diff
@@ -544,15 +635,38 @@ def __repr__(self):
...
+        # This equality check aims to apply specific rules to the contents of
+        # the href element solving false positive cases
+        if not isinstance(other, href_token):
+            return False
+        if self.strict_urls:
+            return StrictUrlRule.clean_resource_url(self, other, self.strict_urls)
+        return super().__eq__(other)
+
+    def __hash__(self):
+        return super.__hash__(self)
```
Mr0grog: I think you can just leave this method out; the effect will be the same.

web_monitoring/html_diff_render.py
```diff
@@ -810,17 +926,51 @@ def html(self):
 # I had some weird concern that I needed to make this token a single word with
 # no spaces, but now that I know this differ more deeply, this is pointless.
 class ImgTagToken(tag_token):
-    def __new__(cls, tag, data, html_repr, pre_tags=None,
+
+    regex = re.compile(r"(src|srcset)=('|\")(.*?)('|\")")
```
Mr0grog: I don’t think you actually need this at all (see below), but if you keep it, please use a more descriptive name.

web_monitoring/html_diff_render.py
```diff
@@ -810,17 +926,51 @@ def html(self):
...
+
+    def are_any_urls_equal(self, other):
+        strict_urls = self.strict_urls
+        self_array = self[7:-3]
```
Mr0grog: Once we get to this point, we should hopefully not be doing any more parsing of strings. I think you should probably be modifying `flatten_el`, `fixup_chunks`, and `_customize_token` here. See below for more.

web_monitoring/html_diff_render.py
```diff
@@ -810,17 +926,51 @@ def html(self):
...
+                if strict_urls:
+                    return StrictUrlRule.clean_resource_url(url_a[1:-1], url_b[1:-1], strict_urls)
+                elif url_a == url_b:
+                    return True
+        return False
+
+    def __new__(cls, tag, data, html_repr, strict_urls, pre_tags=None,
                 post_tags=None, trailing_whitespace=""):
-        obj = DiffToken.__new__(cls, "\n\nImg:%s\n\n" % data,
+        found_urls = ImgTagToken.regex.findall(html_repr)
```
Mr0grog: In `flatten_el`, we grab the actual `src` attribute value out of the `img` element: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/e602b3f94cdcbebc54db0861b38aa6f2a665b9bc/web_monitoring/html_diff_render.py#L693-L694

And then in `fixup_chunks`, we create a `tag_token` instance` where the `data` is just the `src` attribute: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/e602b3f94cdcbebc54db0861b38aa6f2a665b9bc/web_monitoring/html_diff_render.py#L636-L641

(So at this point, we’ve already parsed out the src and have it available.)

Finally, in `_customize_token`, we replace the `tag_token` instance with a `ImgTagToken`: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/e602b3f94cdcbebc54db0861b38aa6f2a665b9bc/web_monitoring/html_diff_render.py#L1036-L1044

A few things are important to note here:

- We already have the value of the `src` element in the form of `self.data`, so there’s no need to parse it out again here. We should just compare `self.data` to `other.data` when we compare.

- When parsing in `flatten_el`, we already have meaningful data in the form of `etree`, so we generally try and pull out all the important stuff there instead of doing more string parsing. At current, we get the `src` that way, but we could also get `srcset` if we want to use it in comparisons. That way we don’t have to parse it again here.

- We only do this messy middle step with `tag_token` because of the incremental way this was forked from lxml. Ideally, we’d refactor to get rid of `_customize_token` entirely. If you want to extract `srcset` for use now, you should…

    1. Modify `flatten_el` to pull that info out.
    2. Modify `fixup_chunks` to create a `ImgTagToken` instance with that data instead of making a `tag_token`.
    3. Then you should get rid of the logic in `_customize_token` that makes `ImgTagToken` instances, because it would never run anymore.
    4. Clean up the logic here and in the `are_any_urls_equal` method to use that data instead of doing more string parsing.

web_monitoring/html_diff_render.py
```diff
@@ -810,17 +926,51 @@ def html(self):
...
+        obj.strict_urls = strict_urls
         return obj
 
+    def __eq__(self, other):
+        if isinstance(other, ImgTagToken):
+            return self.are_any_urls_equal(other)
+        return False
+
+    def __hash__(self):
+        return super.__hash__(self)
```
Mr0grog: I think you can just leave this method out; the effect will be the same.

Mr0grog:
@ftsalamp This could also use some tests, please!

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,74 @@
...
+    URL_RULE = re.compile(r'(https?\:\/\/)?(www\.)?')
+    CLASS_RULES = {'WBM': WaybackUrlComparator,
+                   'jsessionid': ServletSessionUrlComparator,
+                   'UKWA': WaybackUkUrlComparator}
+
+    @staticmethod
+    def clean_resource_url(element, other, mode):
+        try:
+            comparator = StrictUrlRule.CLASS_RULES[mode]()
+            return comparator.compare(element, other)
```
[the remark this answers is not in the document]
Mr0grog: > Additionally, you probably want a way to combine the comparators here

To be clear, the *combining* stuff doesn’t need to be in this PR! We can definitely do that later as a separate piece of work.

web_monitoring/html_diff_render.py
ftsalamp: If I don't include this I get a `TypeError: unhashable type: 'ImgTagToken'`

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,74 @@
 MAX_SPACERS = 2500
 
 
+class WaybackUrlComparator:
+    matcher = re.compile(r'web/\d{14}(im_)?/(https?://)?(www.)?')
+    url_schema_matcher = re.compile(r'(https?\:\/\/)?(www\.)?')
```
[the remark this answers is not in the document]
ftsalamp: You are correct! The `url_schema_matcher` is redundant.

web_monitoring/html_diff_render.py
Mr0grog: Ah! I hadn’t realized Python has some special behavior when `__eq__` is defined. From [the docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__):

> A class that overrides `__eq__()` and does not define `__hash__()` will have its `__hash__()` implicitly set to `None`… If a class that overrides `__eq__()` needs to retain the implementation of `__hash__()` from a parent class, the interpreter must be told this explicitly by setting `__hash__ = <ParentClass>.__hash__`.

So I guess you do need to implement `__hash__`, but it might be better to use the suggested for above, or to at least change `super.__hash__(self)` to use the more standard `super().__hash__()` form.

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,74 @@
 MAX_SPACERS = 2500
 
 
+class WaybackUrlComparator:
+    matcher = re.compile(r'web/\d{14}(im_)?/(https?://)?(www.)?')
```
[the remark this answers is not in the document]
ftsalamp: > Finally, in the ending part, is there a case where it would not be an absolute URL? i.e. is (https?://)? really optional? If not, we should probably also remove the (www.)? after it, since that isn’t really doing anything either. Should this also be restricted to http, either? Couldn’t some of these URLs be ftp://, etc?

We could have http://www.anexample......
                        http://anexample......
                        https://www.anexample......
                        https://anexample......
So I believe the regex is correct.
I haven't run into any `ftp://` 's etc so far. @vbanos is this a possibility?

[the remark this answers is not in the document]
ftsalamp: > Based on that, it looks it should probably be (id_|im_|js_|cs_)?.
There can't be _id in a snapshot but the rest should be added.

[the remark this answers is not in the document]
Mr0grog: So, based on that, it looks like the protocol, at least, is not optional, so you should not have `(https?://)?`, you should have `https?://`.

web_monitoring/html_diff_render.py
```diff
@@ -187,8 +187,74 @@
...
+        else:
+            if hasattr(url_a, 'strict_urls'):
+                url_a.strict_urls = None
+            if hasattr(url_b, 'strict_urls'):
+                url_b.strict_urls = None
+            return url_a == url_b
+
+
+class WaybackUkUrlComparator(WaybackUrlComparator):
+    matcher = re.compile(r'https://www\.webarchive\.org\.uk/wayback/en/archive/\d{14}mp_/')
```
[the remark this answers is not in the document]
ftsalamp: > Is mp_ a valid suffix that’s not documented (or documented somewhere I don’t know about)? Is it particular to the UK?

I didn't find any documentation about it but when looking at the HTML code UKWA returns, if a link has the `_mp` attribute, the "raw" snapshot as it was saved is returned. And all their links have it. Otherwise, the source link is rendered in an iframe using JS, and is thus undiffable.

It it specific to the UKWA.

web_monitoring/html_diff_render.py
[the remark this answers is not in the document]
ftsalamp: Sorry, I omitted the examples with the protocol. Again it might or might not be there.

Mr0grog:
@ftsalamp do you think you might be able to rebase this whole branch (it seems like you rebased in the middle somehow) so we get a more readable diff and don’t rewrite a bunch of commits that are already on master?

ftsalamp:
@Mr0grog I've added some tests and fixed a bug they uncovered. If you think we are ready to merge I'll make a new PR from a new branch created from master with a cleaner (only two commits) commit history. Let's discuss the changes here and close the PR when we don't need it anymore.

Mr0grog:
Rebased this branch so I can see what’s actually happening here in order to review.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -175,3 +175,85 @@ def test_html_diff_works_on_documents_with_no_body():
...
+
+    assert results['change_count'] == 0
+
+
+def test_html_diff_works_with_jsessionid():
+    results = html_diff_render(
+        '<a href="https://www.ncdc.noaa.gov/homr/api;jsessionid=A2DECB66D2648B'
+        'FED11FC721FC3043A1"></a>',
+        '<a href="https://www.ncdc.noaa.gov/homr/api;jsessionid=45312D9542FDB0'
+        '15289A1BBD76958F43"></a>',
```
Mr0grog: Same formatting note as above. It’s OK to go over line length if it’s easier to read.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -175,3 +175,85 @@ def test_html_diff_works_on_documents_with_no_body():
...
+    assert results['change_count'] == 0
+
+
+def test_html_diff_works_with_uk_wbm_snapshots():
+    results = html_diff_render(
+        '<a href="https://www.webarchive.org.uk/wayback/en/archive/'
+        '20190525141538/https://www.example.gov/></a>',
+        '<a href="https://www.webarchive.org.uk/wayback/en/archive/'
+        '20181231224558/https://www.example.gov/></a>',
+        include='all', strict_urls='UK_WBM')
```
Mr0grog: Same formatting note as above.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -175,3 +175,85 @@ def test_html_diff_works_on_documents_with_no_body():
 
     assert 'combined' in results
     assert isinstance(results['combined'], str)
+
+
+def test_html_diff_works_wbm_snapshots():
```
Mr0grog: Would you please also add at least one test case for the *non-equal* situation? For example, the exact same test as this, but with `strict_urls=None` and asserting that `results['change_count'] == 1`.

Mr0grog:
Rebased on master again. Working on updating and finalizing this today.

Mr0grog:
OK, going ahead and merging this. Will make separate follow-up issues for comparison vs. normalization and for applying this logic to the links diff.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -175,3 +175,127 @@ def test_html_diff_works_on_documents_with_no_body():
...
+            src="https://upload.wikimedia.org/wikipedia/commons/thumb/0/02/OSIRIS_Mars_true_color.jpg/275px-OSIRIS_Mars_true_color.jpg"
+            srcset="https://upload.wikimedia.org/wikipedia/commons/thumb/0/02/OSIRIS_Mars_true_color.jpg/413px-OSIRIS_Mars_true_color.jpg 1.5x, https://upload.wikimedia.org/wikipedia/commons/thumb/0/02/OSIRIS_Mars_true_color.jpg/550px-OSIRIS_Mars_true_color.jpg 2x"
+            data-file-width="2205"
+            data-file-height="2205"
+            width="275"
+            height="275">
+        ''',
+        include='all')
+
+    assert results['change_count'] == 0
```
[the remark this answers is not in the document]
vbanos: `srcset` can contain various images. The decision on which one to load depends on the current client resolution.
I agree that ALL need to match, not just one.

PR #405, opened 2019-05-15
Use lru_cache to speed up tag_info()

vbanos:
`web_monitoring.html_diff_render.tag_info()` method reads a tag (e.g.
`<span>`) and returns a `TagInfo` named tuple.

Since the range of tags is somehow limited, this method receives the
same tags and repeats its processing over and over again even when
working on a single page. It makes sense to cache its results in memory
using `functools.lru_cache` to speedup processing.

Experiments using `.cache_info()` also showed that the cache is
effective (many hits) and operation is faster.

Note that we moved `tag_info` outside `reconcile_change_groups` method
to be able to reuse the memory cache universally.

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -16,6 +16,7 @@ For now, you can mentally divide this module into two sections:
 """
 from bs4 import BeautifulSoup, Comment
 from collections import Counter, namedtuple
+from functools import lru_cache
 import copy
 import difflib
 from web_monitoring.utils import get_color_palette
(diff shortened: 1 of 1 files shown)


web_monitoring/html_diff_render.py
```diff
@@ -1453,6 +1454,12 @@ def merge_change_groups(change_chunks, doc, tag_type=None):
 
 TagInfo = namedtuple('TagInfo', ('name', 'open', 'source'))
 
+@lru_cache(maxsize=1024)
+def tag_info(token):
```
Mr0grog: If we’re pulling this out of `reconcile_change_groups()`, would you please:

- Add a docstring
- Rename the `token` argument? (Maybe something like `tag` or `tag_source` or `tag_text` or `token_text` or `token_source` or `token_html`.)

Mainly this is because `lru_cache` would be unsafe to use with instances of `DiffToken` (two instances with different properties can still compare the same and wind up using the same slot in the cache’s underlying dictionary—you can test it to see).

I’m 99% certain we never wind up using a `DiffToken` instance as input to this function, but I want to make it extra clear so someone doesn’t accidentally do it. I actually started testing this because *I* forgot this doesn’t take `DiffToken` instances. :)

vbanos: Done!

vbanos:
I'm sorry but I don't have specific numbers. I did test runs comparing the `master` and this branch and it was always faster. Unit tests also run faster.

Issue #206, opened 2018-08-01
Radical suggestion: Deprecate this repo and break up pieces


danielballan:
I am tempted to propose deprecating web-monitoring-processing and creating several separate repos in its place.

* Python RESTer to the Internet Archive CDX API
* web-monitoring-db Python RESTer
* diffing functions
* diffing server, which could gradually grow into the Python implementation of the Ruby server -- see https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/119

Arguments in favor:
* "Many small packages" is good for reuse and shared maintenance. Lots of people might be happy to use and potentially contribute to a modern, well-tested Python RESTer for the Internet Archive CDX API, but be uninterested in the other bits.
* The diffing server and the diffing functions have very different (maybe totally nonintersecting?) dependency sets. It would be nice to attract more contributors to our diffing work, and it might be easier to get up and running if they didn't have to encouter the server code at all.
* It's hard to properly implement semver [and, more generally, to manage release schedules] on a codebase that has so many loosely coupled pieces. This is why I propose separae repos, with one Python package per repo, as opposed to several Python packages that share one repo. It's better  to have one package per repo so that the version on PyPI and the version in the git tag are in sync.
* This would be a graceful way to give up on some of the old code that we have maintained (well, refrained from deleting at least) in the hope that some hackathon contributors might someday return to expand on them. It's been 18 months.

Arguments against:
* There are already a lot of repos! And more repos can mean more baseline maintenance work. Cookiecutters (like [this one](https://nsls-ii.github.io/scientific-python-cookiecutter/) I developed recently) can lower the startup cost, but there is an ongoing maintenance cost to keeping CI up to date, etc.
* Coordinated releases can be tricky. Currently, if we want to change the API between the diffing server and the diffing functions, we can just do it in one granular step. If we split them up, we have to manage coordinated releases. This is probably not a big problem since we are the only user of this code.

Mr0grog:
Note there is another concern not well covered here: scraping (#174) and monitoring (#172). That said, they could maybe be lumped in with a future “Python implementation of the Ruby server.”

My 2¢: I’m not sure I see as much value in splitting it up so granularly. I feel like I see two major areas here…

1. Diffing/Diffing Service (which I kind of want to merge into a Python rewrite of the DB server, except I suspect that would be problematic for Internet Archive if they really want to make something serious out of the diffing work here — can you comment, @vbanos?)

[...]

Issue #138, opened 2017-12-14
Refactor on top of html5-parser


danielballan:
It seems like we should be using https://html5-parser.readthedocs.io/en/latest/ for its robustness and speed.

Once the HTML differ is working well enough and tested, we should consider refactoring on top of this.

## Differs That Need Updating

- [x] side_by_side_text
- [x] links
- [x] links_json
- [x] html_text_dmp
- [x] html_token

Mr0grog:
Question: that parser will build a variety of trees; are you thinking we should switch to an lxml tree or stick with BeautifulSoup? If the latter, I think this: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/05a0b2550ac5a2982487087c29055600cdf8c214/web_monitoring/differs.py#L30-L35

Can just have line 32 replaced with:

```py
soup = html5_parser.parse(html, treebuilder='soup')
```

And everything else stays the same. It also happens to potentially* solve #288.

(*It solves the parsing breakage on my local machine, but not 100% sure that’s the same issue as the server is encountering.)

Mr0grog:
Anyway, we could start with that one and expand to the other differs as we have time to double-check and make sure all the various APIs are compatible.

vbanos:
Fix: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/410

PR #410, opened 2019-05-20
Use html5-parser instead of lxml

vbanos:
Replace all `BeautifulSoup(html, 'lxml')` with
`html5_parser.parse(html, treebuilder='soup', return_root=False)`.

The aim is to improve robustness and speed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ diff_match_patch_python ~=1.0.2
 docopt ~=0.6.2
 git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
+html5-parser ~=0.4.6
 lxml ~=4.3.3
 pandas ~=0.24.2
 sentry-sdk ~=0.8.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ diff_match_patch_python ~=1.0.2
 docopt ~=0.6.2
 git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
-html5-parser ~=0.4.7
+html5-parser ~=0.4.7 --no-binary lxml
 lxml ~=4.3.3
 pandas ~=0.24.2
 sentry-sdk ~=0.8.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ diff_match_patch_python ~=1.0.2
 docopt ~=0.6.2
 git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
-html5-parser ~=0.4.6
+html5-parser ~=0.4.7
 lxml ~=4.3.3
 pandas ~=0.24.2
 sentry-sdk ~=0.8.0
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -287,9 +287,6 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
                 'meta',
                 content=_diff_title(soup_old, soup_new))
             title_meta.attrs['name'] = 'wm-diff-title'
-            if not soup.head:
-                head = soup.new_tag('head')
-                soup.insert(0, head)
             soup.head.append(title_meta)
 
             old_head = soup.new_tag('template', id='wm-diff-old-head')
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -20,6 +20,7 @@ import copy
 import difflib
 from web_monitoring.utils import get_color_palette
 import html
+import html5_parser
 import logging
 import re
 from .content_type import raise_if_not_diffable_html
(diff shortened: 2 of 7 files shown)


vbanos:
I'm concerned about `html5-parser` installation.
You need to install it using `pip install --no-binary lxml html5-parser` because of the following:
 
_It is important that lxml is installed with the --no-binary flag. This is because without it, lxml uses a static copy of libxml2. For html5-parser to work it must use the same libxml2 implementation as lxml. This is only possible if libxml2 is loaded dynamically._ https://html5-parser.readthedocs.io/en/latest/

I couldn't find a way to include `--no-binary` in `setup.py`.

web_monitoring/html_diff_render.py
```diff
@@ -284,6 +287,9 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
                 'meta',
                 content=_diff_title(soup_old, soup_new))
             title_meta.attrs['name'] = 'wm-diff-title'
+            if not soup.head:
+                head = soup.new_tag('head')
+                soup.insert(0, head)
```
Mr0grog: This doesn’t work right. I’m guessing you added it to fix the tests, but it could significantly mangle the output. I took a look and it turns out there’s a bug in html5-parser’s `soup` tree builder: https://github.com/kovidgoyal/html5-parser/issues/18

We can add a workaround for now or we can wait for it to be fixed, but I don’t think we can merge this as-is.

Mr0grog:
Ah, it looks like we need to `apt get install pkg-config libxml2-dev` in the `Dockerfile` now, as well: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/76a3710d4848ea2d97bf45da2688e636dfc11140/Dockerfile#L5-L6

web_monitoring/html_diff_render.py
vbanos: Oh you fixed the `html5-parser` bug, great! I'll remove my workaround.

vbanos:
You also need `libxslt-dev` in addition to `libxml2-dev` I think. https://lxml.de/installation.html

Mr0grog:
Huh, you’re right, that’s really clearly documented. I’m surprised building the Docker image didn’t give me a compiler error on the `pip install` step!

web_monitoring/html_diff_render.py
Mr0grog: Hopefully it gets released soon!

danielballan:
This looks promising

https://github.com/pypa/pip/issues/4307#issuecomment-282816515

though I have never encountered it in the wild myself.

danielballan:
I confirmed that

```bash
python -m venv ~/venv/test-no-binary
echo "lxml --no-binary lxml" > req.txt
pip install -r req.txt 
```

compiles from source, circumventing the binary wheel, per the link above, but this does fall in the category of things that are valid in `requirements.txt` but not valid in the `install_requires` block of `setup.py`.

Mr0grog:
> this does fall in the category of things that are valid in `requirements.txt` but not valid in the `install_requires` block of `setup.py`.

Alright, so the real question is: is that OK? It’s certainly not ideal.

I lean towards “yes”:
- Our local install instructions suggest using `requirements.txt`, so it shouldn’t matter for that case.
- We don’t publish this on PyPI, and I think (?) we plan to do #206 before we do, so the set of things this might affect hopefully isn’t big.
- I *think* we are getting a pretty huge improvement in parsing accuracy and performance here, and I feel like a slightly more complicated install is probably worth that.
    - I’m taking the fact that @vbanos submitted this PR instead of a comment on the issue (#138) as an indicator that Internet Archive sees this similarly. Let me know if I’m misunderstanding.

web_monitoring/html_diff_render.py
vbanos: @Mr0grog still no release for `html-parser` :(

Mr0grog: 0.4.7 is out now!

vbanos: After using `html5-parser 0.4.7` and removing the workaround I see that the tests are fixed.

vbanos:
All issues are addressed now.
If you finally decide that the installation process is cool, that would be great.

requirements.txt
```diff
@@ -5,6 +5,7 @@ diff_match_patch_python ~=1.0.2
 docopt ~=0.6.2
 git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
+html5-parser ~=0.4.7
```
Mr0grog: Per @danielballan’s [comment](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/410#issuecomment-495080497), I think this still needs `--no-binary lxml` on the end.

Mr0grog: @danielballan that’s how this should be set up, right?

Mr0grog:
Noted a couple tiny issues, otherwise this looks good to me. Thanks, @vbanos!

requirements.txt
vbanos: @Mr0grog when I do that, I get an error. Check out:
https://travis-ci.org/edgi-govdata-archiving/web-monitoring-processing/builds/541947661?utm_source=github_status&utm_medium=notification

Mr0grog: Seems like we need to do some better parsing in `setup.py` to support this: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/19d9de2bc558a10851601487bb34c7478cde4747/setup.py#L22-L25

vbanos: @Mr0grog I'm sorry but I'm a bit confused.
What do you suggest to do here? Remove `--no-binary lxml`?

Mr0grog: Well, the [format of lines in a pip file](https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format) is:

```
<requirement specifier> [; markers] [[--option]...]
```

Whereas `install_requires` in `setup.py` can only have `<requirement specifier>`. So it seems like the right move is to remove anything after the specifier.

It’s not perfectly accurate, but the easiest thing might be to just look for ` ;` or ` --` on the line and, if present, remove that string and everything after it. That would probably do the job well enough for our purposes, at least.

vbanos: OK, its done! Thank you for the explanation @Mr0grog . Could you please merge now?

Mr0grog: Looks like you are still some linting errors, but will take a look and see about merging when I am off work tonight or tomorrow.

vbanos: Linting error fixed!

Mr0grog:
👍 Looks great!

PR #411, opened 2019-05-22
Improve encoding detection

vbanos:
Improve `web_monitoring.diffing_server._extract_encoding` in many ways and add more unit tests.

1. Drop regex to detect html & xml encoding from specific tags in their contents and use `cchardet` instead. This has MANY benefits: a) can detect encoding for any type of HTML/XML/other markup without needing to check for specific meta tag. b) previous meta tag extraction couldn't handle border cases with various syntax errors, etc. c) Code is simplified. b) `cchardet` is faster than applying regex to response contents.

2. Handle bad content-type header cases. Previous code couldn't work with them.

3. Use `codecs.lookup` to check if the detected encoding is known to python. If not, fallback to default. Its beter if all encoding related logic is inside `_extract_encoding` and not have some of it in `_decode_body` as it was earlier.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -5,7 +5,7 @@ import hashlib
 import inspect
 import functools
 import os
-import re
+import cchardet
 import sentry_sdk
 import tornado.gen
 import tornado.httpclient
@@ -53,19 +53,6 @@ DIFF_ROUTES = {
     "html_tree_diff": web_monitoring.differs.html_tree_diff,
     "html_differ": web_monitoring.differs.html_differ,
 }
-
-# Matches a <meta> tag in HTML used to specify the character encoding:
-# <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
-# <meta charset="utf-8" />
-META_TAG_PATTERN = re.compile(
-    b'<meta[^>]+charset\\s*=\\s*[\'"]?([^>]*?)[ /;\'">]',
-    re.IGNORECASE)
-
-# Matches an XML prolog that specifies character encoding:
-# <?xml version="1.0" encoding="ISO-8859-1"?>
-XML_PROLOG_PATTERN = re.compile(
-    b'<?xml\\s[^>]*encoding=[\'"]([^\'"]+)[\'"].*\?>',
-    re.IGNORECASE)
 
 client = tornado.httpclient.AsyncHTTPClient()
 
@@ -329,16 +316,19 @@ def _extract_encoding(headers, content):
     content_type = headers.get('Content-Type', '').lower()
     if 'charset=' in content_type:
         encoding = content_type.split('charset=')[-1]
-    if not encoding:
-        meta_tag_match = META_TAG_PATTERN.search(content, endpos=2048)
-        if meta_tag_match:
-            encoding = meta_tag_match.group(1).decode('ascii', errors='ignore')
-    if not encoding:
-        prolog_match = XML_PROLOG_PATTERN.search(content, endpos=2048)
-        if prolog_match:
-            encoding = prolog_match.group(1).decode('ascii', errors='ignore')
-    if encoding:
-        encoding = encoding.strip()
+        if encoding:
+            encoding = encoding.strip()
+    if not encoding and content:
+        # try to identify encoding using cchardet. Use up to 18kb of the
+        # content for detection. Its not necessary to use the full content
+        # as it could be huge. Also, if you use too little, detection is not
+        # accurate.
+        detected = cchardet.detect(content[:18432])
+        if detected:
+            detected_encoding = detected.get('encoding')
+            if detected_encoding:
+                encoding = detected_encoding.lower()
+
     # Handle common mistakes and errors in encoding names
     if encoding == 'iso-8559-1':
         encoding = 'iso-8859-1'
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -326,7 +326,7 @@ class DiffHandler(BaseHandler):
 
 def _extract_encoding(headers, content):
     encoding = None
-    content_type = headers.get('Content-Type', '')
+    content_type = headers.get('Content-Type', '').lower()
     if 'charset=' in content_type:
         encoding = content_type.split('charset=')[-1]
     if not encoding:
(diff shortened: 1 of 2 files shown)


vbanos:
In addition to encoding detection improvements, I noticed something else:

We detect undecodable / binary files by the percentage of a specific char in their content. 
https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/diffing_server.py#L367

I think that there is a better way to do this. Common file formats start with specific values, e.g. a PDF starts with `b'PDF-'` and an octet stream starts with  `b'\x00'`. (There is a list of ~20 checks like that). Since we already have the content of the response in `_decode_body`, we could make a small function to check for these before anything else and raise `UndecodableContentError` very early.

Mr0grog:
> I think that there is a better way to do this. Common file formats start with specific values, e.g. a PDF starts with b'PDF-' and an octet stream starts with b'\x00'.

Yes! We already do *some* of that here: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/e602b3f94cdcbebc54db0861b38aa6f2a665b9bc/web_monitoring/content_type.py#L10-L22

That gets used when calling `content_type.is_not_html` or `content_type.raise_if_not_diffable_html`, which is the first step in some of the diffs.

I don’t remember whether there is a reason I didn’t separate out the binary types so they could be checked *before* calling the diff method (i.e. as part of `diffing_server._decode_body`). It seems like it would make a lot of sense to put them there, so I’ll have to look at the relevant PRs and see if there’s a reason. If you have more stuff worth adding to that logic, though, it would be super helpful! (But probably do it in a different PR.)

Mr0grog:
IIRC, the reason we have both that and the “percentage of NUL terminator characters” check is that we saw a bunch of real-world items that would trip one but not the other. It’s been a while, though, so I don’t remember specific examples.

web_monitoring/diffing_server.py
```diff
@@ -325,35 +313,40 @@ def check_response_for_error(self, response):
...
     if 'charset=' in content_type:
         encoding = content_type.split('charset=')[-1]
-    if not encoding:
-        meta_tag_match = META_TAG_PATTERN.search(content, endpos=2048)
-        if meta_tag_match:
-            encoding = meta_tag_match.group(1).decode('ascii', errors='ignore')
-    if not encoding:
-        prolog_match = XML_PROLOG_PATTERN.search(content, endpos=2048)
-        if prolog_match:
-            encoding = prolog_match.group(1).decode('ascii', errors='ignore')
```
Mr0grog: We felt it was important to behave like a web browser does here, which is to look for these items, in this order, and obey them before trying to detect the encoding. CChardet doesn’t look at any of these, and my thinking was that if we added cChardet, we’d use it *in addition* to this logic for exactly this reason. (See this PR for more on that reasoning: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/161)

Do you have a strong reason for why these should be removed? If not, I would like to make sure we keep them.

web_monitoring/diffing_server.py
```diff
@@ -325,35 +313,40 @@ def check_response_for_error(self, response):
...
+        if encoding:
+            encoding = encoding.strip()
+    if not encoding and content:
+        # try to identify encoding using cchardet. Use up to 18kb of the
+        # content for detection. Its not necessary to use the full content
+        # as it could be huge. Also, if you use too little, detection is not
+        # accurate.
+        detected = cchardet.detect(content[:18432])
```
Mr0grog: Do we need to guard against possible exceptions here?

web_monitoring/diffing_server.py
```diff
@@ -325,35 +313,40 @@ def check_response_for_error(self, response):
...
         encoding = 'iso-8859-1'
     # Windows-1252 is so commonly mislabeled, WHATWG recommends assuming it's a
     # mistake: https://encoding.spec.whatwg.org/#names-and-labels
     if encoding == 'iso-8859-1' and 'html' in content_type:
         encoding = 'windows-1252'
+    # Check if the selected encoding is known. If not, fallback to default.
+    try:
+        codecs.lookup(encoding)
+    except (LookupError, ValueError, TypeError):
+        encoding = 'ascii'
```
Mr0grog: This contradicts and prevents the first line of `_decode_body` from working; we default to UTF-8 encoding and only fall back to ASCII if that doesn’t work: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/e602b3f94cdcbebc54db0861b38aa6f2a665b9bc/web_monitoring/diffing_server.py#L349-L350

Additionally, is there a case where we should expect a `ValueError` from `codecs.lookup`?

web_monitoring/diffing_server.py
vbanos: No, it doesn't throw any exceptions from my experience.

web_monitoring/diffing_server.py
vbanos: I read #161. IMHO you depend too much on website defined encoding and they are wrong some times from my experience. 
I do not disagree strongly with your approach, I just thought correct encoding was covered with `cchardet` and we could also save some CPU cycles by removing the regex.
I will update my PR to keep them. NP.

web_monitoring/diffing_server.py
vbanos: This was a mistake, it should fallback to `utf-8`, I'll fix it.

Its possible to have a `ValueError` if we got a weird `encoding` value from headers or meta tag. I have seen all sorts of things like invalid unicode chars in there. Please lets keep it.

web_monitoring/diffing_server.py
Mr0grog: I hear you on this; it’s just that when I originally tried using cchardet the way you are in #161, I frequently saw *diffs display pages with questionable encoding differently than I would if I just viewed the page in my browser*, and I feel strongly that was a problem.

I’m all for not overspending on CPU cycles, and if you think there’s a more efficient way to match this logic than by using a regex (or if we can better optimize the regex), that’d be great.

web_monitoring/diffing_server.py
Mr0grog: Makes sense!

web_monitoring/diffing_server.py
```diff
@@ -336,24 +338,37 @@ def _extract_encoding(headers, content):
...
 
 def _decode_body(response, name, raise_if_binary=True):
     encoding = _extract_encoding(response.headers, response.body) or 'UTF-8'
-    try:
-        text = response.body.decode(encoding, errors='replace')
-    except LookupError:
-        # If the encoding we found isn't known, fall back to ascii
-        text = response.body.decode('ascii', errors='replace')
-
+    text = response.body.decode(encoding, errors='replace')
```
Mr0grog: We don’t need the `or 'UTF-8'` in the previous line anymore.

vbanos: Done!

Mr0grog:
Oops, not sure what I thought I was waiting for in this PR. Merging now, sorry!

PR #416, opened 2019-06-03
Better binary response detection

vbanos:
Major `web_monitoring.content_type` improvements. We now detect binary types using the first bytes of the raw response and not regex. We do this once in `_decode_body` and not inside each diff method.

Drop redundant `raise_if_not_diffable_html`

Update unit tests. Add unit tests which use file types in `tests/fixtures/` to validate the new method.

Question: What is the best way to pass the `content_type_options` param to `_decode_body` to
pass it to `is_not_html`?


web_monitoring/diffing_server.py
```diff
@@ -367,6 +368,13 @@ def _extract_encoding(headers, content):
 
 
 def _decode_body(response, name, raise_if_binary=True):
+    """Check if the response is binary. If not, continue and detect encoding.
+    Try to define if a significantly large part of the body was undecodable
+    before returning decoded body.
+    """
+    if is_not_html(response.body, response.headers):
```
Mr0grog: I don’t think we should be checking for whether something is HTML here. What if we had a differ for RSS or XML or CSV documents in the future? There are plenty of valid text types that aren’t HTML. This should just be checking for whether what we got is not decodable text. We talked about this on Slack: https://archivers.slack.com/archives/C47DJK8H0/p1559529058016900

I’m also a little worried about not supporting the `content_type_options` argument at this point (its behavior is meant to be analogous to the `X-Content-Type-Options` response header). We felt it was important to have in case we ran into situations were sniffing was likely to produce a false positive. A text document *could* always start with `GIF8` and it seemed important to provide a way to handle that. Do you not think that’s important here?

web_monitoring/html_diff_render.py
```diff
@@ -249,13 +248,6 @@ def html_diff_render(a_text, b_text, a_headers=None, b_headers=None,
     text2 = '<!DOCTYPE html><html><head></head><body><h1>Header</h1></body></html>'
     test_diff_render = html_diff_render(text1,text2)
     """
-    raise_if_not_diffable_html(
-        a_text,
-        b_text,
-        a_headers,
-        b_headers,
-        content_type_options)
-
```
Mr0grog: On Slack, we talked specifically about still needing to do checks here: https://archivers.slack.com/archives/C47DJK8H0/p1559529058016900

At the server level, we can check whether the content is probably a binary type when trying to decode it to text, but we should still be doing additional checks here to determine whether we’re actually dealing with HTML-formatted text here (as opposed to, say, plain text or maybe a CSV or JSON document).

web_monitoring/links_diff.py
```diff
@@ -22,13 +21,6 @@ def links_diff(a_text, b_text, a_headers=None, b_headers=None,
     as an internal link, but not:
         <a href="http://this.domain.com/this/page#anchor-in-this-page">Text</a>
     """
-    raise_if_not_diffable_html(
-        a_text,
-        b_text,
-        a_headers,
-        b_headers,
-        content_type_options)
-
```
Mr0grog: Same here.

web_monitoring/content_type.py
```diff
@@ -42,17 +25,41 @@
...
+    """
+    Determine whether an HTTP response is not HTML by its headers. In general,
+    this errs on the side of leniency; it should have few false positives, but
+    many false negatives.
 
     Parameters
     ----------
-    text : string
-        Potential HTML content string
+    raw_body : raw binary response
```
Mr0grog: Please stick to Numpy docstring style. The *type* should be after the colon and the description should be on the next line and indented.

web_monitoring/content_type.py
```diff
@@ -42,17 +25,41 @@
...
+        if data[1:4] == b'IF8':
+            return "image/gif"
+    elif b0 == b'\x89':
+        if data[1:4] == b'PNG':
+            return "image/png"
+
+
+def is_not_html(raw_body, headers=None, check_options='normal'):
+    """
+    Determine whether an HTTP response is not HTML by its headers. In general,
```
Mr0grog: This sentence is misleading, since this method uses both the headers *and* the `raw_body` bytes.

web_monitoring/tests/test_content_type.py
```diff
@@ -0,0 +1,20 @@
...
+    ('edgi_background.jpg', 'image/jpeg'),
+    ('unknown_encoding.html', None),
+    ('empty.txt', None),
+    ('poorly_encoded_utf8.txt', None)
+    ])
+def test_detect_binary_type(filename, binary_type):
+    """Read file as binary stream. Detect its type
+    """
+    fname = Path(__file__).resolve().parent / 'fixtures' / filename
+    with open(fname, 'rb') as fp:
```
Mr0grog: Please avoid short, abbreviated variable names like `fp`. See the styleguide: https://github.com/edgi-govdata-archiving/web-monitoring/blob/master/CONTRIBUTING.md#naming

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -94,19 +95,21 @@ def test_html_diff_render_should_not_break_with_empty_content():
...
-
+    pdf_raw = Path(pdf_file).read_bytes()
+    a = Mock(body=pdf_raw, headers={})
+    b = Mock(body=b'<p>Just a little HTML</p>')
     with pytest.raises(UndiffableContentError):
-        html_diff_render(
-            '<p>Just a little HTML</p>',
-            pdf_content)
+        html_diff_render(_decode_body(a, name='a'),
+                         _decode_body(b, name='b'))
```
Mr0grog: See my earlier points above about still needing checks in `html_diff_render`.

The way things are written now, the `html_diff_render` call won’t throw an error (or rather, it could throw an error, but it would never be the one we’re testing for). The name of this test is a misnomer and `html_diff_render()` shouldn’t even be involved. If we split up the checks (doing binary decoding checks at the server level, and HTML checks at the method level), we also need to split up the tests—you need unit tests for binary data sent to `_decode_body` and separate unit tests for non-HTML strings/headers sent to `html_diff_render`.

Same comments for all the remaining tests in this PR.

Mr0grog:
So, in principle, I think what we’re trying to do in this PR is still OK (check for binary content when decoding instead of when starting the diff, after decoding), but reviewing it reminded me why I didn’t originally do that:

The sniffing is a *lot* more likely to trigger a false positive if you are decoding any old string of text. When you know you’re expecting HTML, you can be a lot more sure that a bytestring starting with `GIF8` is probably not OK. If you only know you’re looking for text, though (which is all we know at the point where we run `_decode_body`), it’s not nearly as clear that a bytestring starting with `GIF8` is really a problem.

Mr0grog:
@vbanos I just ran across this nice module today that might do a better job detecting a lot more things: https://github.com/h2non/filetype.py

Don’t know if you think it would be useful.

web_monitoring/content_type.py
```diff
@@ -42,17 +25,41 @@
...
+        if data[1:5] == b'PDF-':
+            return "application/pdf"
+        elif data[1:5] == b'!PS-':
+            return "application/postscript"
+    elif b0 == b'G':
+        if data[1:4] == b'IF8':
+            return "image/gif"
+    elif b0 == b'\x89':
+        if data[1:4] == b'PNG':
+            return "image/png"
```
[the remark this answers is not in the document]
vbanos: What I did is based on wayback machine code to detect mime types.
I'm afraid there isn't any public doc available.

vbanos:
About using module: https://github.com/h2non/filetype.py
I'm against adding extra dependencies unless its totally necessary.

vbanos:
After some though, I believe that my PR creates more problems than its solving. @Mr0grog your detailed comments made me understand that. Its better to close it.

Mr0grog:
@vbanos if you measure it and there’s really significant performance gain to be had here, we could still:

- Change from the regex to string comparisons.
- Change `html_diff_render` and `links_diff` to take bytes instead of text and do the decoding there, so we can be a bit more efficient about sniffing the bytes before decoding them. That way there are no worries about false positives in non-HTML text.

PR #418, opened 2019-06-07
Support `different` & changes in DB versions API

Mr0grog:
I needed to run a script and query for versions that were *not* different recently, but the DB API lacks support for that parameter. While I was at it, I added support for `include_change_from_previous` and `include_change_from_earliest` to `list_versions` and `get_version` as well since they were missing but easy to support.

`list_versions` is still missing support for `sort` (that really needs to be implemented on *all* `list_*` methods) and `status` (more complicated because it can be a range and needs custom serialization).


Mr0grog:
Going ahead and merging.

PR #422, opened 2019-06-13
Make a separate `build_docker` job in CircleCI

Mr0grog:
This adds a new CI job called `build_docker` that just builds the Docker image in order to make sure it works.

In #410, I did a bad job testing the Docker build (I tested an earlier version, but forgot to test the final version that we merged, which failed). That led to a hotfix (#419), which is no good. This should help mitigate that.

The first commit here also contains a bogus job called `use_built_docker`, which I’m using to test workspaces so that it’s possible the release workflow wouldn’t need to repeat building. ~Not ready for merging, obviously!~ *(Update: this is all fixed up now.)*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           at: /tmp/workspace
       - run:
           name: Load Image
-          command: docker load /tmp/workspace/docker-image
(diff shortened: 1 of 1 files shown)


Mr0grog:
Alright, this should be ready to roll now.

PR #424, opened 2019-06-19
Clean up iterator and add multiple active elements test

vbanos:
Instead of creating the same `template` tag multiple times inside the
`for` loop, we create it only once and reuse it.

`enumerate` is redundant as we don't use the `index` it creates.

### DIFF
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -341,10 +341,10 @@ def _cleanup_document_structure(soup):
 
 
 def _deactivate_deleted_active_elements(soup):
-    for index, element in enumerate(soup.find_all(ACTIVE_ELEMENTS)):
+    wrapper = soup.new_tag('template')
+    wrapper['class'] = 'wm-diff-deleted-inert'
+    for element in soup.find_all(ACTIVE_ELEMENTS):
         if element.find_parent('del'):
-            wrapper = soup.new_tag('template')
-            wrapper['class'] = 'wm-diff-deleted-inert'
             element.wrap(wrapper)
 
     return soup
--- a/web_monitoring/html_diff_render.py
+++ b/web_monitoring/html_diff_render.py
@@ -341,10 +341,10 @@ def _cleanup_document_structure(soup):
 
 
 def _deactivate_deleted_active_elements(soup):
-    wrapper = soup.new_tag('template')
-    wrapper['class'] = 'wm-diff-deleted-inert'
     for element in soup.find_all(ACTIVE_ELEMENTS):
         if element.find_parent('del'):
+            wrapper = soup.new_tag('template')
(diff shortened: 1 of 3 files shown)


web_monitoring/html_diff_render.py
```diff
@@ -341,10 +341,10 @@ def _cleanup_document_structure(soup):
 
 
 def _deactivate_deleted_active_elements(soup):
-    for index, element in enumerate(soup.find_all(ACTIVE_ELEMENTS)):
+    wrapper = soup.new_tag('template')
+    wrapper['class'] = 'wm-diff-deleted-inert'
```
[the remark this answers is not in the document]
vbanos: @Mr0grog thank you for your very detailed feedback.
I have reverted the problematic change and also added a new unit test for `_deactivate_deleted_active_elements`.

web_monitoring/tests/test_html_diff.py
```diff
@@ -89,6 +90,45 @@ def test_contrived_examples_htmldiffer(fn):
     return d
 
 
+def test_deactivate_deleted_active_elements():
```
Mr0grog: Would you please move this test to `test_html_diff_validity.py`? The tests in this module are meant to give us a set of real-world test cases for determining whether changes to the diff algorithm are better or worse, but not to test correctness (that’s why there are no assertions in this file).

Mr0grog: In fact, there’s already a similar test in that module: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/fc36c7c0a902b71de372cc1bd721ee2b8f06fc30/web_monitoring/tests/test_html_diff_validity.py#L35-L66

So it might be a good template to follow for this.

web_monitoring/tests/test_html_diff_validity.py
```diff
@@ -66,6 +67,31 @@ def test_html_diff_render_doesnt_move_script_content_into_page_text():
...
+    '''
+    a = '''<body>
+           <script src="s1.js"></script>
+           <p>test</p>
+           <script src="s2.js"></script></body>'''
+    b = '''<body><p>test2</p></body>'''
+    result = html_diff_render(a, b)['combined']
+    soup = html5_parser.parse(result, treebuilder='soup', return_root=False)
+    elements = soup.select('del.wm-diff > template.wm-diff-deleted-inert > script')
+    assert len(elements) == 2
```
Mr0grog: This would not have failed for the initial PR you made (you can try it!) because you’re just checking for how many `script` elements are in the result, not whether they got moved. The easiest thing might be to count how many `template.wm-diff-deleted-inert` elements there are.

vbanos: Oh, I didn't realise that :( 
I've updated my PR according to your suggestion.

PR #425, opened 2019-06-21
Fix bug with target urls having unicode chars

vbanos:
When the target URL contains unicode characters like the following
capture from Greek wikipedia:

http://web.archive.org/web/20171227204051/https://el.wikipedia.org/wiki/%CE%86%CF%81%CE%B7%CF%82_(%CF%80%CE%BB%CE%B1%CE%BD%CE%AE%CF%84%CE%B7%CF%82)

`client.fetch` fails and the following result is returned:
```
{"code": 400, "error": "HTTP 400: Bad request"}
```

Using `urllib.parse.quote` we escape the chars that cannot occur raw in
the URL. We do not escape some "safe" characters `;/?:@&=+$,%`
because it will break the URL.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -6,6 +6,7 @@ import inspect
 import functools
 import os
 import re
+from urllib.parse import quote
 import cchardet
 import sentry_sdk
 import tornado.gen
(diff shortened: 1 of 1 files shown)


Mr0grog:
@vbanos this works fine for me without any changes: http://localhost:8888/html_token?a=http%3A%2F%2Fweb.archive.org%2Fweb%2F20171227204051%2Fhttps%3A%2F%2Fel.wikipedia.org%2Fwiki%2F%25CE%2586%25CF%2581%25CE%25B7%25CF%2582_(%25CF%2580%25CE%25BB%25CE%25B1%25CE%25BD%25CE%25AE%25CF%2584%25CE%25B7%25CF%2582)&b=http%3A%2F%2Fweb.archive.org%2Fweb%2F20171227204051%2Fhttps%3A%2F%2Fel.wikipedia.org%2Fwiki%2F%25CE%2586%25CF%2581%25CE%25B7%25CF%2582_(%25CF%2580%25CE%25BB%25CE%25B1%25CE%25BD%25CE%25AE%25CF%2584%25CE%25B7%25CF%2582)

Are you not urlencoding the values in the querystring? In that case, I do get a 400 error, but that seems correct to me. I believe we’ve had some conversations about this before on Slack.

vbanos:
Yes, you are right. URL encoding in our app solves this problem. Thank you.

PR #429, opened 2019-06-25
Use more descriptive errors for upstream failures

Mr0grog:
When a request to another server to get diff content fails, we used to simply return the upstream server's error, which makes it tough to determine which request was problematic (was it the `a` or `b` param?) and can be hard to distinguish from an error inside the diffing server itself.

Instead, we now use 502/504 status codes in these cases (these codes are both for *gateways,* implying that the error happened upstream) and add the URL that failed to the error message. The upstream status code (if any) and message are also included in the full error text. This should help make it much more clear which request is failing and where exactly the failure is coming from.

Hopefully this will make situations like #425 more clear. @vbanos any thoughts on this?

We could also provide more structured data in the response JSON for the upstream URL, status code, and message, but that would require changes to `write_error()` and I wasn’t sure it was actually worthwhile.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -318,6 +318,10 @@ class DiffHandler(BaseHandler):
                     self.send_error(
                         504,
                         reason=f'Timed out while fetching "{url}"')
+                elif isinstance(response.error, ValueError):
+                    self.send_error(
+                        400,
+                        reason=str(response.error))
                 else:
                     self.send_error(
                         502,
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -311,17 +311,17 @@ class DiffHandler(BaseHandler):
             try:
                 response.rethrow()
             except (ValueError, OSError, tornado.httpclient.HTTPError):
-                # Response code == 599 means that
-                # no HTTP response was received.
-                # In this case the error code should
-                # become 400 indicating that the error was
-                # raised because of a bad request parameter.
-                if response.code == 599:
+                url = response.request.url
+                # Timeouts aren't distinguishable from other network failures
+                # by status code, so check the actual error type.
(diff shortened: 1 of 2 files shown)


web_monitoring/diffing_server.py
```diff
@@ -311,17 +311,17 @@ def check_response_for_error(self, response):
...
+                if isinstance(response.error, tornado.simple_httpclient.HTTPTimeoutError):
                     self.send_error(
-                        400, reason=str(response.error))
+                        504,
+                        reason=f'Timed out while fetching "{url}"')
                 else:
                     self.send_error(
-                        response.code, reason=str(response.error))
+                        502,
+                        reason=f'Received a {response.code} status while fetching "{url}": {response.error}')
```
Mr0grog: Important question: if there was a network failure that wasn't a timeout, are we worried about this leaking sensitive information? You might get an error like:

```json
{
  "code": 502,
  "error": "Received a 599 status while fetching \"http://reallynowherethisdoesntexistpleasefail.org/web/20171227204051/https://el.wikipedia.org/wiki/\u0386\u03c1\u03b7\u03c2_(\u03c0\u03bb\u03b1\u03bd\u03ae\u03c4\u03b7\u03c2)\": [Errno 8] nodename nor servname provided, or not known"
}
```

I *think* it’s ok, but we could also have a custom clause for 599 status codes (network failures).

Mr0grog:
The test failures here are CDX API failures on Wayback’s side while building docs. The actual changes here are passing. 🙁

Mr0grog:
Issues resolved and tests passing :)

vbanos:
I think that this improvement will also help us present clearer error messages to the user. Its very welcome.

Mr0grog:
Awesome! I’m going ahead and merging this, then refactoring #431 on top of it.

PR #431, opened 2019-06-25
Reorganize remote content fetching in diff server

### LINKED ISSUE #282: Handle Upcoming Tornado HTTPClient Errors
We’re getting a bunch of warnings from Tornado in our tests that future releases *will* raise some errors even if `HTTPClient.fetch(raise_error=False)`:

```
/home/circleci/web-monitoring-processing/venv/lib/python3.6/site-packages/tornado/httpclient.py:294: DeprecationWarning: raise_error=False will allow 'Unsupported url scheme: example.org' to be raised in the future
  response.error, DeprecationWarning)

/home/circleci/web-monitoring-processing/venv/lib/python3.6/site-packages/tornado/httpclient.py:294: DeprecationWarning: raise_error=False will allow 'Unsupported url scheme: example.org' to be raised in the future
  response.error, DeprecationWarning)

/home/circleci/web-monitoring-processing/venv/lib/python3.6/site-packages/tornado/httpclient.py:294: DeprecationWarning: raise_error=False will allow '[Errno -2] Name or service not known' to be raised in the future
  response.error, DeprecationWarning)

/home/circleci/web-monitoring-processing/venv/lib/python3.6/site-packages/tornado/httpclient.py:294: DeprecationWarning: raise_error=False will allow '[Errno -2] Name or service not known' to be raised in the future
  response.error, DeprecationWarning)
```

Basically, it sounds like it will start only affecting whether an error is raised in response to > 400 status code, and other kinds of errors will always get raised.

Since we don’t wrap that in a `try...except` block, that’s going to cause problems, and we need to fix it.

Mr0grog:
This resolves the deprecations in Tornado's HTTP client by reorganizing the code responsible for fetching and validating the content to diff. We previously had a lot of logic spread around `get()` that was just fetching the remote content. I've placed all of it into a method for fetching a single URL. Hopefully this also makes it a little more readable, though it could probably still be a bit cleaner.

This also adds tests for hash validation, which we apparently did not have before. It also runs hash validation against local files now.

**This will conflict with #429.** Whichever merges second will need updating.

Fixes #282, and should unblock #369.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -174,69 +174,78 @@ class DiffHandler(BaseHandler):
                        'You must provide a URL as the value '
                        'for both `a` and `b` query parameters.')
             return
-        # Special case for local files, for dev/testing.
-        responses = {}
-        for param, url in urls.items():
-            if url.startswith('file://'):
-                if os.environ.get('WEB_MONITORING_APP_ENV') == 'production':
-                    self.send_error(
-                        403, reason=("Local files cannot be used in "
-                                     "production environment."))
-                    return
-                headers = {'Content-Type': 'application/html; charset=UTF-8'}
-                with open(url[7:], 'rb') as f:
-                    body = f.read()
-                    responses[param] = MockResponse(url, body, headers)
-        # Now fetch any nonlocal URLs.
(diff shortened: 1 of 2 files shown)


Mr0grog:
Rebased on top of #429.

PR #453, opened 2019-08-07
Accept responses with HTTP status >= 300 from web archives

vbanos:
We need to accept and process responses from web archives without
requiring a status = 2xx response.
We check for HTTP header `Memento-Datetime` which is used in web
archives to detect that.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -226,14 +226,7 @@ class DiffHandler(BaseHandler):
 
             try:
                 response = yield client.fetch(url, headers=headers,
-                                              validate_cert=VALIDATE_TARGET_CERTIFICATES,
-                                              raise_error=False)
-                # Accept response with HTTP status >= 300 only if its coming
-                # from a web archive. HTTP header `Memento-Datetime` is an
-                # indication about that.
-                if response.code >= 300 and \
-                        response.headers.get('Memento-Datetime') is None:
-                    response.rethrow()
+                                              validate_cert=VALIDATE_TARGET_CERTIFICATES)
             except ValueError as error:
                 self.send_error(400, reason=str(error))
             except OSError as error:
@@ -241,8 +234,16 @@ class DiffHandler(BaseHandler):
             except tornado.simple_httpclient.HTTPTimeoutError:
                 self.send_error(504, reason=f'Timed out while fetching "{url}"')
             except tornado.httpclient.HTTPError as error:
-                self.send_error(502,
-                                reason=f'Received a {error.response.code} status while fetching "{url}": {error}')
+                # If the response is actually coming from a web archive,
+                # allow error codes. The Memento-Datetime header indicates
+                # the response is an archived one, and not an actual failure
+                # to respond with the desired content.
+                if error.response is not None and \
+                        error.response.headers.get('Memento-Datetime') is not None:
+                    response = error.response
+                else:
+                    self.send_error(502,
+                                    reason=f'Received a {error.response.code} status while fetching "{url}": {error}')
 
         if response and expected_hash:
             actual_hash = hashlib.sha256(response.body).hexdigest()
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -226,7 +226,14 @@ class DiffHandler(BaseHandler):
 
             try:
                 response = yield client.fetch(url, headers=headers,
-                                              validate_cert=VALIDATE_TARGET_CERTIFICATES)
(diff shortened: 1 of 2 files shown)


vbanos:
The relevant issue is: https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/404
Another approach to address the same issue is: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/451

vbanos:
When trying to test this new code with a page that returns 404, e.g.
`/html_token?a=http://web.archive.org/web/20101130151428/http://vbanos.gr/&b=http://vbanos.gr/not-found"`
the app runs and returns the expected result but the following new exception is raised:
```
HTTPServerRequest(protocol='http', host='vbanos-dev.us.archive.org:8888', method='GET', uri='/html_token?a=http://web.archive.org/web/20101130151428/http://vbanos.gr/&b=http://vbanos.gr/not-found', version='HTTP/1.1', remote_ip='94.67.65.186')
  File "/opt/venv-3.6.9/lib/python3.6/site-packages/tornado/web.py", line 1699, in _execute
    result = await result
  File "/opt/venv-3.6.9/lib/python3.6/site-packages/tornado/gen.py", line 748, in run
    yielded = self.gen.send(value)
  File "/opt/venv-3.6.9/lib/python3.6/site-packages/web_monitoring-0+untagged.757.g189671b.dirty-py3.6.egg/web_monitoring/diffing_server.py", line 192, in get
    self.write(res)
  File "/opt/venv-3.6.9/lib/python3.6/site-packages/tornado/web.py", line 826, in write
    raise RuntimeError("Cannot write() after finish()")
RuntimeError: Cannot write() after finish()
```
I also tried to use `self.send_error()` instead of `response.rethrow()` to avoid it but I still get the same exception.

web_monitoring/diffing_server.py
```diff
@@ -226,7 +226,14 @@ def fetch_diffable_content(self, url, expected_hash, query_params):
...
                 response = yield client.fetch(url, headers=headers,
-                                              validate_cert=VALIDATE_TARGET_CERTIFICATES)
+                                              validate_cert=VALIDATE_TARGET_CERTIFICATES,
+                                              raise_error=False)
+                # Accept response with HTTP status >= 300 only if its coming
+                # from a web archive. HTTP header `Memento-Datetime` is an
+                # indication about that.
+                if response.code >= 300 and \
+                        response.headers.get('Memento-Datetime') is None:
+                    response.rethrow()
```
Mr0grog: `fetch_diffable_content` returns `None` to signal that it failed. However, it depends on the fact that, if `client.fetch()` raised an error `response` will be `None`:

https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/ea6f992b2a6639461ab45667b177d860724f9ebd/web_monitoring/diffing_server.py#L248

And in the calling function, we do `if not all(content)` to check if either fetch returned `None`:

https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/ea6f992b2a6639461ab45667b177d860724f9ebd/web_monitoring/diffing_server.py#L179-L184

However, now you’re populating `response`, so the calling function doesn’t realize that the request failed and so it keeps trying to do a diff anyway.

So… there are probably a few options here:

- Store `response` in a different variable so you can call `rethrow()` on it:

    ```py
    bad_response = response
    response = None
    bad_response.rethrow()
    ```

- Don’t use `raise_error=False` and instead pull the response object out of the error:

    ```py
    try:
        response = yield client.fetch(url, headers=headers,
                                      validate_cert=VALIDATE_TARGET_CERTIFICATES)
    except SomeOtherErrorType:
        ...
    except tornado.httpclient.HTTPError as error:
        # If the response is actually coming from a web archive,
        # allow error codes. The Memento-Datetime header indicates
        # the response is an archived one, and not an actual failure
        # to respond with the desired content.
        if error.response is not None and error.response.headers.get('Memento-Datetime') is not None:
            response = error.response
        else:
            self.send_error(502,
                            reason=f'Received a {error.response.code} status while fetching "{url}": {error}')
    ```

    Upside: no need to check status codes. Downside: this code is definitely a little less clear.

- Change the error handling in some more significant way. (After re-reading the code here, I realize we need to do something like that anyhow, since both the fetches could have errors, and we have similar issues. BUT we can also leave fixing that for a separate PR.)

vbanos: Thank you for taking the time to explain this so well.

I prefer your 2nd suggestion. It is much more elegant IMHO. I've updated my PR following that approach.

vbanos:
Yes, I agree. Test added.

Issue #231, opened 2018-08-09
Infer local file type from extension


danielballan:
See https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/88#discussion_r208711145

Mr0grog:
To summarize: when loading a `file://` URL in the diffing server, we assume the content is `application/html` (which is not even a correct mime-type; it should be `text/html`) and that it is UTF-8 encoded: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/33e99ee686c87df0254ef071a53d7c74dda1e457/web_monitoring/diffing_server.py#L136-L139

That’s obviously not correct all the time (e.g. we’d like to support PDFs, see #162, edgi-govdata-archiving#issues/36). We should do our best to infer the correct `Content-Type` header based on the file’s extension.

Python has a built-in module for this: https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type

There are also tools on PyPI like [mimelib](https://pypi.org/project/mimelib/) and public databases of extension ↔︎ mime type mappings: https://github.com/jshttp/mime-db

Mr0grog:
Oops, this should have been closed with #457!

PR #457, opened 2019-08-15
Use extension to guess MIME type for local files

mboren:
Previously, we assumed `application/html; charset=UTF-8` for all local files. That's not really a valid MIME type, and we would like to support PDFs, and perhaps other stuff, in the future.

I used the built-in `mimetypes` package to guess values for the `Content-Type` and `Content-Encoding` headers based on the file extension.

This package only suggests `Content-Encodings` for a few archive formats (eg .gz, .bz2), so I'm not sure if it's ever likely to be relevant for our purposes, but, hey, it's free. I'm happy to delete this part if it doesn't seem likely to be useful in the future.

Issue: https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/231

### DIFF
--- a/web_monitoring/diff_server/server.py
+++ b/web_monitoring/diff_server/server.py
@@ -5,6 +5,7 @@ from docopt import docopt
 import hashlib
 import inspect
 import functools
+import mimetypes
 import os
 import re
 import cchardet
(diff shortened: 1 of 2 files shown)


Mr0grog:
Note: reviewed this in person w/ @mboren, who is still working on it!

mboren:
I added tests and moved the logic into the MockResponse class as we discussed. I'm fairly certain that I didn't put the tests in the best place, but I'm not really sure where they should go.

web_monitoring/diffing_server.py
```diff
@@ -78,12 +79,33 @@ def __init__(self, url):
...
     "An HTTPResponse-like object for local file:/// requests."
-    def __init__(self, url, body, headers):
+    def __init__(self, url, body, headers=None):
         self.request = MockRequest(url)
         self.body = body
         self.headers = headers
         self.error = None
 
+        if self.headers is None:
+            self.headers = {}
```
Mr0grog: This check could probably be simplified by just updating the setter for headers above on line 85:

```py
self.headers = headers or {}
```

(It’s not strictly the same, but probably good enough for this case.)

Mr0grog:
It’s been a while, so I went ahead and rebased this, then removed the private API usage in the tests. Going to merge if no issues in CI.

Mr0grog:
Thanks for doing most of the legwork on this, @mboren!

PR #476, opened 2019-09-10
HOTFIX: fix truncated healtcheck script messages

Mr0grog:
Out healthcheck script was unintentionally stuck on a very old version of the Sentry SDK because it had to run on the same codebase as our Wayback imports, which were part of a very long-lived branch. When everything got merged, we wound up getting upgraded to a *much* newer version of the Sentry client package, and it turns out it now truncates strings at 512 characters. Of course, the healthcheck script outputs pretty long messages, so that's not OK.

The max length is hardcoded as a constant in `sentry.utils.MAX_STRING_LENGTH`, and there doesn't appear to be any setting or feature to override it. Instead, there is a comment in the code that says:

    # This is intentionally not just the default such that one can patch `MAX_STRING_LENGTH` and affect `strip_string`.

...so that's what I've gone ahead and done here. It doesn't feel great, so I also commented on a relevant post on Sentry's forums about it: https://forum.sentry.io/t/some-stack-traces-are-truncated/7309/4

### DIFF
--- a/scripts/ia_healthcheck
+++ b/scripts/ia_healthcheck
@@ -12,6 +12,15 @@ import sys
 from web_monitoring import db, internetarchive
 
 
+# The current Sentry client truncates string values at 512 characters. It
+# appears that monkey-patching this module global is only way to change it and
+# that doing so is the intended method:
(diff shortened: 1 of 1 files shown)


Mr0grog:
For what it’s worth: the truncation was introduced in [v0.8.0](https://github.com/getsentry/sentry-python/blob/master/CHANGES.md#080); we jumped from v0.6.2 to v0.11.2.

PR #479, opened 2019-09-23
Remove PageFreezer support

### LINKED ISSUE #379: Remove Pagefreezer support
We have modules for diffing and for pulling data from Pagefreezer, but we haven’t been using them for quite a while—and it’s highly likely they are out of date and have bugs we don’t know about. I’d like to remove them, since they don’t make sense for us to maintain at this point.

Any concerns with this, @danielballan?

Mr0grog:
Remove all our legacy, deprecated code for working with data and diffs from PageFreezer. We haven’t used it in a long time and it’s likely much of this code is buggy or out of date. We no longer maintain our own PageFreezer install anymore either, so it’s unlikely we’ll ever revisit this code in the near-to-medium future.

Fixes #379.


Mr0grog:
I’m going to go ahead and just merge this, since it’s just ripping out code we haven’t touched in multiple years.

PR #487, opened 2019-10-03
Add script to manually import annotations from analyst CSVs

mrotondo:
This is a precursor to any automated importing, and does not include any de-duping or merging, so don't run it on the same csv multiple times!

Known issues (to either be addressed in this PR or subsequent ones, your call):
- doesn't set the annotation author correctly, instead uses the API call initator
- doesn't include any metadata about sheet "schema"
- doesn't access sheets via the google drive API, just parses CSVs
- as noted, no merging or deduping


mrotondo:
I'm putting this up for review so that we can iron out any initial issues and then decide on paths for the merging/deduping and author-setting issues

web_monitoring/annotations_import.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+    arguments = docopt(doc)
+    is_important_changes = arguments['--is_important_changes']
+    csv_path = arguments['<csv_path>']
+
+    # Missing step: Analyze CSV to determine spreadsheet schema version
+    for row in read_csv(csv_path):
+        change_ids = find_change_ids(row)
+        annotation = create_annotation(row, is_important_changes)
+        response = post_annotation(change_ids, annotation)
```
Mr0grog: I’m pretty sure this will raise a `db.WebMonitoringDbError` if there’s a problem with the POST. You might want to consider catching and logging it so the whole job doesn’t crash if there’s one bad row.

web_monitoring/annotations_import.py
```diff
@@ -0,0 +1,116 @@
...
+        print(response)
```
Mr0grog: Logging every line might get a little verbose. You could use a logger and log this at the `DEBUG` level.

web_monitoring/annotations_import.py
```diff
@@ -0,0 +1,116 @@
...
+
+    try:
+        annotation['priority'] = float(csv_row['Priority (algorithm)'])
+    except:
+        annotation['priority'] = 0.0
+
+    return annotation
+
+def post_annotation(change_ids, annotation):
+    cli = Client.from_env()
```
Mr0grog: This is a pretty minor nit, but could you please avoid abbreviating this variable name? See also https://github.com/edgi-govdata-archiving/web-monitoring/blob/master/CONTRIBUTING.md#naming

mrotondo:
Thanks for all the comments! I'll get around to most of this tomorrow, but:

> I think the plan we landed on for the API side of this is that multiple posts by the same user for the same change would just update a single annotation, so that should mitigate this. Of course, I need to check whether that’s what we’re actually doing in practice. :P

I think https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/42bb8c7020e4d139fa278a95e0947cf9c8329659/app/models/change.rb#L82 looks like it is indeed finding any extant annotation for that change by that author, but then overwriting its data.

Mr0grog:
> but then overwriting its data.

Ahhhhh, so it is. I guess you’d need to find it and read it here first, though it might be simpler if posting an annotation took a parameter for update vs. replace. Hmmm.

web_monitoring/annotations_import.py
mrotondo: Done, although from my simulated errors against a local server, it seems like db.py isn't properly catching/surfacing exceptions originating on the server

web_monitoring/annotations_import.py
mrotondo: Done, and including a change to .env.example which actually makes log levels set there stick.

web_monitoring/annotations_import.py
mrotondo: If you want to verify this and make an issue for it, I can look into that as well

Mr0grog: I just tried this and it seems to:

```py
>>> from web_monitoring.db import Client
>>> Client.from_env().add_annotation(annotation={}, page_id='c4328d30-cada-452f-8642-4bff721f5fc2', from_version_id='a8a346fd-e275-498f-9463-11e1b88beb71', to_version_id='77ceecf5-a5a0-4ea6-9a05-afe79f85b356')
  File "<stdin>", line 1, in <module>
  File "/Users/rbrackett/Dev/datarescue/processing/web_monitoring/db.py", line 679, in add_annotation
    _process_errors(res)
  File "/Users/rbrackett/Dev/datarescue/processing/web_monitoring/db.py", line 38, in _process_errors
    raise WebMonitoringDbError(', '.join(map(repr, errors)))
web_monitoring.db.WebMonitoringDbError: {'status': 422, 'title': 'Annotations cannot be empty'}
```

(…although we really should be parsing more useful info out of that…)

Did you have a case where it should have but didn’t?

web_monitoring/annotations_import.py
Mr0grog: Yes! It is a continual process of weeding out ones that were done that way in older code. :\

scripts/annotations_import.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        return None
+
+def create_annotation(csv_row, is_important_changes):
+    # Missing step: capture info from the "Who Found This" column and map it to
+    # the author's user record in the database eventually, but that requires a
+    # change to the API
+
+    bool_keys = [
+        'Language alteration',
```
Mr0grog: This is looking pretty good to me, but the one thing I’m wondering is whether we should attempt to map these names to something a little less prone to change, e.g. “Content change/addition/removal” → “content_changed”.

I’m not sure how often the titles of these columns has been altered without changing their meaning in the past, but I’m fairly sure it’s happened a few times (fixing typos, slight rewordings, etc.). Even though doing a mapping like this means any display needs to also know what text to show, it seems like it would make a more reliable system.

web_monitoring/annotations_import.py
mrotondo: It seems like 404s, e.g. annotations POSTs with page IDs which aren't in the server's DB, get raised as HTTPError (and there's a json decode error as well, for a blank response I presume):

```python
  File "/home/roto/edgi-govdata-archiving/web-monitoring-processing/web_monitoring/db.py", line 34, in _process_errors
    errors = res.json()['errors']
  File "/usr/lib/python3.7/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

  File "scripts/annotations_import.py", line 130, in <module>
    main()
  File "scripts/annotations_import.py", line 123, in main
    response = post_annotation(change_ids, annotation)
  File "scripts/annotations_import.py", line 97, in post_annotation
    from_version_id=change_ids['from_version_id'])
  File "/home/roto/edgi-govdata-archiving/web-monitoring-processing/web_monitoring/db.py", line 679, in add_annotation
    _process_errors(res)
  File "/home/roto/edgi-govdata-archiving/web-monitoring-processing/web_monitoring/db.py", line 36, in _process_errors
    res.raise_for_status()
  File "/usr/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost:3000/api/v0/pages/ca5f40d4-89bf-4d7d-bf1f-e89470f0c1b4/changes/..1fdfcee3-62ea-4bdb-9130-c3b289704dfc/annotations
```

Mr0grog: Oh, that is because you are hitting a server in dev mode instead of production mode. In dev mode, you get a fancy error page instead of JSON: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/e6a98a3f6e0512962d60075d36ce75ee4e728433/app/controllers/api/v0/api_controller.rb#L6

Try a bad production or staging URL, like: https://api-staging.monitoring.envirodatagov.org/api/v0/pages/ca5f40d4-89bf-4d7d-bf1f-e89470f0c1b4/changes/..1fdfcee3-62ea-4bdb-9130-c3b289704dfcd/annotations

On the other hand, it would be nice if:
- The DB server responded with a JSON error if you use the `Accept: application/json` header, regardless of dev vs. production mode.
- `db.py` in this repo actually set that header (which really is an oversight; it should always have been doing that).
- Debatable: should we really fall back to `HTTPError` if the response fails to parse, or still use `WebMonitoringDbError` in that case, just with a special message? It would certainly be more stable if we ever wanted to change HTTP libraries under the hood. (Any thoughts on this?)

But all of those are probably out of scope for this PR.

Mr0grog: Wrote two issues around that: edgi-govdata-archiving/web-monitoring-db#614 and #495.

scripts/annotations_import.py
mrotondo: Done, please take a look to see if the structure looks alright, and if you'd like to change any of the new json keys

scripts/annotations_import.py
```diff
@@ -0,0 +1,174 @@
...
+    def fieldnames(self):
+        return [name.strip() for name in super().fieldnames]
+
+def read_csv(csv_path):
+    with open(csv_path, newline='') as csvfile:
+        reader = DictReaderStrip(csvfile)
+        for row in reader:
+            yield row
+
+DIFF_URL_REGEX = re.compile('^.*\/page\/(.*)/(.*)\.\.(.*)')
```
Mr0grog: Missed this earlier — there’s some unnecessary/malformed escaping here. It works in practice, but could be a little cleaner. Using a raw string might be a good idea, too:

```suggestion
DIFF_URL_REGEX = re.compile(r'^.*/page/(.*)/(.*)\.\.(.*)')
```

scripts/annotations_import.py
```diff
@@ -0,0 +1,174 @@
...
+            'high': 1.0
+        }
+        row_importance = csv_row['Importance?'].lower().strip()
+        significance = importance_significance_mapping.get(row_importance, 0.0)
+    annotation['significance'] = significance
+
+    return annotation
+
+def post_annotation(change_ids, annotation):
+    client = Client.from_env()
```
Mr0grog: Minor nit — might be nice to only create one client and pass it in instead of creating a new one for every row.

Or do we even need a function for this at that point, since the entire function body would just be:

```py
return client.add_annotation(**change_ids, annotation=annotation)
```

scripts/annotations_import.py
mrotondo: Done, thanks for the catch.

scripts/annotations_import.py
mrotondo: :+1:

Mr0grog:
Is this good to merge now?

mrotondo:
This is good to go, thanks!

Mr0grog:
THEN IN IT GOES

PR #488, opened 2019-10-03
Use LABEL instruction instead of MAINTAINER

### LINKED ISSUE #467: Switch MAINTAINER to LABEL directive in Docker
We use the `MAINTAINER` directive in our docker files, but it’s been deprecated:

> The `LABEL` instruction is a much more flexible version of this and you should use it instead, as it enables setting any metadata you require, and can be viewed easily, for example with docker inspect. To set a label corresponding to the `MAINTAINER` field you could use:
> 
> ```
> LABEL maintainer="SvenDowideit@home.org.au"
> ```

- [`MAINTAINER` docs](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)
- [`LABEL` docs](https://docs.docker.com/engine/reference/builder/#label)

rushilgala:
Fixes  #467

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Use an official Python runtime as a parent image
 FROM python:3.7-slim
-MAINTAINER enviroDGI@gmail.com
+LABEL maintainer="enviroDGI@gmail.com"
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


Mr0grog:
Thanks, @rushilgala! Would you like to be added to the [contributors list](https://github.com/edgi-govdata-archiving/web-monitoring-processing/#contributors)?

PR #489, opened 2019-10-03
Add colons between parameter names and types in docs

### LINKED ISSUE #468: Parameter Names and Types are Displayed Wrong in Docs
In a recent Sphinx or Numpydoc update, our docs stopped differentiating between a method’s parameters and their types, so they get joined up with no spaces in the display:

Screen Shot 2019-08-29 at 8 47 28 AM

They used to display like this:

Screen Shot 2019-10-04 at 12 35 42 PM

*(Updated 2019-10-04 to cross out the incorrect analysis below. See the first comment (https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/468#issuecomment-538532324) for correct info.)*

---

**Old, incorrect analysis:**

~We think there was an update that switched the parameter documentation style from:~

```
Parameters
----------
page_id : string
    Page to which the Version is associated
```

~To:~

```
Parameters
----------
page_id: string
    Page to which the Version is associated
```

~Note the missing space between the parameter name and the colon. So we probably just need to update all the docstrings throughout the codebase to remove those spaces.~

mrotondo:
This addresses the change in the doc generation parser which was swallowing the colon. Fixes #468


mrotondo:
Thanks for digging into this! I'll stick with the issue, the actual fix will be more satisfying than this commit, which definitely felt arbitrary!

mrotondo:
This PR now adds the custom styling which properly causes API parameters to have ": " inserted before their type. In #468, you also mention adding "nicer styling for field lists", is there a particular aspect of the current (definitely very plain) list formatting that you think should be improved?

Mr0grog:
I don’t think I had any particularly strong ideas here. The biggest thing for me was perhaps to visually distinguish the headings (e.g. “Parameters,” “Returns,” etc.) in a clearer way from the actual list items, since they both have the same styling right now.

You could also try matching it up with the old styling (where those headings were to the left of the list of parameters/return values/etc. CSS grid would probably be the easiest way to do that (just spitballing here, not really well tested):

```css
.rst-content .field-list.simple {
  display: grid;
  grid-template-columns: 10em 1fr;
}

.rst-content .field-list.simple > dt {
  background: transparent;
  border: none;
  display: block;
}
```

But you could also just leave it for now, and I or someone else with stronger layout ideas could come back around later to improve.

mrotondo:
I'm gonna leave this alone for now, since it's more legible than it was and the class constructor field lists don't have quite the same css structure as the method field lists. This should be good to go

Mr0grog:
Great, merging this, then. 😄

PR #490, opened 2019-10-04
Use async/await syntax in diff server

### LINKED ISSUE #432: Use Python native async syntax in diff server
In the diffing server, we use Tornado, which existed before asyncio and has specialized decorators for creating coroutines. Since version 5 (which we are on, and we are ready to upgrade to 6), it’s not only had support for the new native Python async syntax, but Tornado devs encourage its use and are deprecating some of the async features Tornado duplicates.

This is great! But we definitely should upgrade our code to use the new native syntax now that it’s well supported.

I’m reasonably certain these changes are confined to:
- `web_monitoring/diffing_server.py`
- `web_monitoring/tests/test_diffing_server_exc_handling.py`

And involve replacing code like:

```python
# Old-style, non-native:
@tornado.gen.coroutine
def some_function(x, y, z):
    ...
    result = yield some_async_function(x)
    ...
    raise tornado.gen.Return(return_value)

# New, native async style:
async some_function(x, y, z):
    ...
    result = await some_async_function(x)
    ...
    return return_value
```

mrotondo:
…more modern asyncio syntax. Resolves #432

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -174,10 +174,11 @@ class DiffHandler(BaseHandler):
             return
 
         # TODO: Add caching of fetched URIs.
-        content = [await self.fetch_diffable_content(url,
-                                                     query_params.pop(f'{param}_hash', None),
-                                                     query_params)
-                         for param, url in urls.items()]
+        requests = [self.fetch_diffable_content(url,
+                                                query_params.pop(f'{param}_hash', None),
+                                                query_params)
+                    for param, url in urls.items()]
+        content = await asyncio.gather(*requests)
         if not all(content):
             return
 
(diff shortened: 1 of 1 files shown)


web_monitoring/diffing_server.py
```diff
@@ -252,19 +250,20 @@ def fetch_diffable_content(self, url, expected_hash, query_params):
...
         """
         Actually do a diff between two pieces of content, optionally retrying
         if the process pool that executes the diff breaks.
         """
         executor = self.get_diff_executor()
         for attempt in range(tries):
             try:
-                result = yield executor.submit(caller, func, a, b, **params)
-                raise tornado.gen.Return(result)
+                loop = asyncio.get_running_loop()
```
Mr0grog: You could probably pull this out of the for loop and do it at the same time we first get `executor`.

web_monitoring/diffing_server.py
```diff
@@ -252,19 +250,20 @@ def fetch_diffable_content(self, url, expected_hash, query_params):
...
+                result = await loop.run_in_executor(
+                    executor, functools.partial(caller, func, a, b, **params))
+                return result
```
Mr0grog: I think you can just `return await whatever` directly:

```suggestion
                return await loop.run_in_executor(
                    executor, functools.partial(caller, func, a, b, **params))
```

mrotondo:
Thanks for the pointer about gather, my intuition about gen.coroutine yielding a list (that they would execute async-but-sequentially) was wrong.

Mr0grog:
Yeah, it’s one of the few bits of syntax that was much more nicely done in Tornado’s custom coroutine setup than in the newer native Python stuff.

Is this good to go now?

PR #497, opened 2019-10-16
Use 502 status code for hash mismatches in diff server

### LINKED ISSUE #491: If fetched content does not match hash in diffing server, don’t return 500 error
In the diffing server, you can provide expected SHA-256 hashes for the two pages you are diffing, e.g:

```
/html_token?a=https://example.com/v1&a_hash=abc&b=https://example.com/v2&b_hash=xyz
```

If the hash doesn’t match the content we got when fetching `a` or `b`, we don’t run the diff and instead return an error: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/d8481952c85fd338b4912e98046e4eb4fd50041e/web_monitoring/diffing_server.py#L247-L253

That validation is super useful, but the error we return uses the `500` status code, which means something went wrong on our end. Something didn’t go wrong on our end, though! The problem was either upstream (the content at URL `a` or `b` was wrong) or on the user’s end (they asked for the wrong URLs, or mistyped the hash, etc).

So, we should return a different status code. Maybe one of:

- `409` (“conflict”) since what you asked for is in conflict with the actual current state of the data.
- `422` (“unprocessable entity”) since you provided parameters that can’t be used.
- `502` (“bad gateway”) since we got an upstream response that was no good.

I lean towards `502`, since the problem may or may not be something the user can actually address. The downside is that we also use this if the upstream server returned a non-200 response code, so it requires some human inspection to differentiate between the two `502` error cases.

Mr0grog:
When using the diff server, you can provide a hash to verify that the two pieces of content you are diffing were correctly received. When that hash doesn’t match, we used to respond with a 500 error, but we now use a 502 error to clarify that the problem is related to the upstream server, not us.

Since that can be a little ambiguous (502 = hash mismatch, but also 502 = upstream server error), this also adds a mechanism for sending extra metadata with the error. Specifically, you can now check `error.type` in the response for a more detailed error code, like `HASH_MISMATCH` or `UPSTREAM_ERROR`.

Fixes #491.

### DIFF
--- a/web_monitoring/diffing_server.py
+++ b/web_monitoring/diffing_server.py
@@ -241,7 +241,12 @@ class DiffHandler(BaseHandler):
                     response = error.response
                 else:
                     self.send_error(502,
-                                    reason=f'Received a {error.response.code} status while fetching "{url}": {error}')
+                                    reason=f'Received a {error.response.code} '
+                                           f'status while fetching "{url}": '
+                                           f'{error}',
+                                    extra={'type': 'UPSTREAM_ERROR',
+                                           'url': url,
+                                           'upstream_code': error.response.code})
 
         if response and expected_hash:
             actual_hash = hashlib.sha256(response.body).hexdigest()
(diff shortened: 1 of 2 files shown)


Mr0grog:
Metadata for a hash mismatch looks like:

```json
{
    "code": 502,
    "error": "Fetched content at \"https://example.com\" does not match hash \"xyz\".",
    "type": "HASH_MISMATCH",
    "url": "https://example.com",
    "expected_hash": "xyz",
    "actual_hash": "3587cb776ce0e4e8237f215800b7dffba0f25865cb84550e87ea8bbac838c423"
}
```

Issue #486, opened 2019-10-02
`db.Client` should include timeouts for all HTTP requests


Mr0grog:
The [`Client` class in `db.py`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/803af9662c061045c2ebe7e0a82e314ca9b98249/web_monitoring/db.py#L118) should include optional timeouts on all HTTP calls.

- The simplest method might be to accept an optional `timeout` keyword argument to the constructor (and the `from_env()` class method) to set a timeout for all calls that instance makes.

- Even fancier would be to support `timeout` as a keyword argument on all method calls that make HTTP requests that overrides the class level timeout. Or maybe an `http_options` or `request_options` dict that takes any special arguments to the actual `request` call?

jatinAroraGit:
Hey @Mr0grog , 
Can I have a shot at this. I am very new to Python so wanna try to crack this one to sharpen up.

julianmclain:
@Mr0grog Is this still relevant?

I like the idea of your second bullet point. I'm looking at the `http_client` in [twilio-python](https://github.com/twilio/twilio-python/blob/master/twilio/http/http_client.py) for inspiration. They use a class level timeout and optional `timeout=None` keyword arg for each method call.

I'm also curious about what prompted this issue. Is the primary goal to add a default timeout, or will modules using the DB client immediately start customizing the timeout?

(also what should be the default timeout?)

Mr0grog:
> Is this still relevant?

Yes, definitely!

> I'm looking at the `http_client` in [twilio-python](https://github.com/twilio/twilio-python/blob/master/twilio/http/http_client.py) for inspiration. They use a class level timeout and optional `timeout=None` keyword arg for each method call.

👍 

> I'm also curious about what prompted this issue. Is the primary goal to add a default timeout, or will modules using the DB client immediately start customizing the timeout?

The primary goal is definitely a default. IIRC, this was prompted by me occasionally issuing poor-performing queries that would hang for a while. But as a more general concern: it’s kind of dangerous to send HTTP requests without timeouts, since it can make your program hang. It’s a bad pattern that we didn’t set timeouts in the DB client to begin with. 🙁

> (also what should be the default timeout?)

I think 30 seconds might be reasonable, but if I’m honest, that’s a totally arbitrary choice! 😜

julianmclain:
Ok great - thanks for the background. I'll take a look at it this week.

[...]

PR #502, opened 2019-10-31
Unify db client get/post request creation

mrotondo:
In preparation for tasks such as adding timeouts or accept headers to all requests, centralize API request creation

### DIFF
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -163,11 +163,26 @@ variables:
 Alternatively, you can instaniate Client(user, password) directly.""")
         return cls(email=email, password=password, url=url)
 
-    def get_request(self, url, params=None):
+    def get(self, url, params=None):
         return requests.get(url,
                             params=params,
                             auth=self._auth,
                             headers={'accept': 'application/json'})
+
+    def post(self, url, data, params=None):
+        return requests.post(url,
+                             data=json.dumps(data),
+                             params=params,
+                             auth=self._auth,
+                             headers={'Content-Type': 'application/json'})
+
+    def post_multi(self, url, data, params=None):
+        return requests.post(url,
+                             data='\n'.join(map(json.dumps, data)),
+                             params=params,
+                             auth=self._auth,
+                             headers={'Content-Type': 'application/x-json-stream'})
+
 
     ### PAGES ###
 
@@ -228,7 +243,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
                   'active': active,
                   'include_total': include_total or None}
         url = f'{self._api_url}/pages'
-        res = self.get_request(url, params=params)
+        res = self.get(url, params=params)
         _process_errors(res)
         result = res.json()
         data = result['data']
@@ -270,7 +285,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         response : dict
         """
         url = f'{self._api_url}/pages/{page_id}'
-        res = self.get_request(url)
+        res = self.get(url)
         _process_errors(res)
         result = res.json()
         # In place, replace datetime strings with datetime objects.
@@ -355,7 +370,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
             url = f'{self._api_url}/versions'
         else:
             url = f'{self._api_url}/pages/{page_id}/versions'
-        res = self.get_request(url, params=params)
+        res = self.get(url, params=params)
         _process_errors(res)
         result = res.json()
         # In place, replace datetime strings with datetime objects.
@@ -389,7 +404,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         url = f'{self._api_url}/versions/{version_id}'
         params = {'include_change_from_previous': include_change_from_previous,
                   'include_change_from_earliest': include_change_from_earliest}
-        res = self.get_request(url, params=params)
+        res = self.get(url, params=params)
         _process_errors(res)
         result = res.json()
         data = result['data']
@@ -438,9 +453,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
             title=title,
             source_metadata=source_metadata)
         url = f'{self._api_url}/pages/{page_id}/versions'
-        res = requests.post(url, auth=self._auth,
-                            headers={'Content-Type': 'application/json'},
-                            data=json.dumps(version))
+        res = self.post(url, version)
         _process_errors(res)
         return res.json()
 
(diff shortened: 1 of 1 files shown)


web_monitoring/db.py
```diff
@@ -163,6 +163,27 @@ def from_env(cls):
...
+                             params=params,
+                             auth=self._auth,
+                             headers={'Content-Type': 'application/json'})
+
+    def post_multi(self, url, data, params=None):
+        return requests.post(url,
+                             data='\n'.join(map(json.dumps, data)),
+                             params=params,
+                             auth=self._auth,
+                             headers={'Content-Type': 'application/x-json-stream'})
```
Mr0grog: Under the hood, `request.get/post/etc()` just call `requests.request('<method>', *args, **kwargs)`, so I wonder if we might not just be better off having a single method here, e.g:

```py
def request(self, method, url, **kwargs):
    kwargs.setdefault('auth', self._auth)
    kwargs.setdefault('headers', {})
    kwargs['headers'].setdefault('accept', 'application/json')

    if 'data' in kwargs:
        # Serialize data based on whether its a dict or sequence
        # and set content-type header

    return requests.request(method, url, **kwargs)
```

Or at least have each of these call into a common method like the above that just adds the commone bits (e.g. auth, the `accept` header).

That way we don’t have any repeated logic that could get out of sync (e.g. your `post` and `post_multi` methods are missing the `accept` header).

Mr0grog: Actually, I also forgot `requests.request` and `requests.post` can take a `json` parameter instead of `data` that will automatically get serialized as JSON and set the right header! Not sure why we weren’t using that before.

Mr0grog: Actually, another good option here might be to construct a [`requests.Session`](https://requests.kennethreitz.org/en/master/api/#request-sessions) instance for the `db.Client` and use that.

We do something similar for `WaybackClient` (except we have a special session class there, which we *probably* don’t need here): https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/0c91dca087f62eeb5330d9c05de6e0062f30dad3/web_monitoring/internetarchive.py#L360-L361

web_monitoring/db.py
```diff
@@ -725,7 +734,7 @@ def get_version_content(self, version_id):
         """
         db_result = self.get_version(version_id)
         content_uri = db_result['data']['uri']
-        res = requests.get(content_uri)
+        res = self.get(content_uri)
         _process_errors(res)
         if res.headers.get('Content-Type', '').startswith('text/'):
```
mrotondo: Now that we're setting the accept header to json on all requests, will this branch ever be hit?

web_monitoring/db.py
mrotondo: Thanks for the tip on Session, I'm adopting that (and will bring in the timeout setting from #486 if combining two issues into one PR is OK).

For your comment about the json parameter, do you see any documentation indicating that it'll properly newline-separate multiple json objects and set content-type=x-json-stream? I don't see any, so I'm going to continue to use data parameter for now, but happy to drop the conditional if possible!

mrotondo:
I'm going to leave the timeout out of this PR in case @jatinAroraGit wants to take care of it

web_monitoring/db.py
Mr0grog: Ohhhhh, thanks for catching this! We actually shouldn’t be setting the `accepts` header for this (though I’m not sure it will make a difference in most cases) and yes, this branch should still be here.

This method is sort of an odd one out — it’s getting the raw response body from an archived request rather than data (i.e. it returns the response body associated with a particular Version), so we could get back all kinds of stuff. The branch here is meant to be a useful shortcut (since `.content` is bytes and `.text` is a decoded string).

mrotondo: Posted a change that restores this special case and blanks out the accept header (as per the "Remove a value from a from a dict parameter" note [here](https://requests.kennethreitz.org/en/master/user/advanced/#session-objects). How's that look to you?

Mr0grog: Makes a lot of sense! The one thing that feels a bit iffy about it to me is the `raw_response` parameter, which changes the return type. Somebody once suggested I should avoid writing functions where an argument switches the return types or shape, which I’ve generally found to be smart advice. (So I might just use the session directly in the one exceptional situation, or have a `request` and a `request_json` method that adds the parsing to `request` or something.)

That said, I don’t think this is an especially big deal in this one case. Happy to merge as is.

web_monitoring/db.py
```diff
@@ -163,6 +168,22 @@ def from_env(cls):
 Alternatively, you can instaniate Client(user, password) directly.""")
         return cls(email=email, password=password, url=url)
 
+    def request(self, method, url, data=None, raw_response=False, **kwargs):
+        if data is not None:
+            headers = kwargs.setdefault('headers', {})
+            if isinstance(data, collections.Sequence):
```
Mr0grog: Can you please use `collections.abc.Sequence` instead? `collections.Sequence` is deprecated and fires off warnings in tests. It’s going to be removed in Python 3.9.

web_monitoring/db.py
```diff
@@ -163,6 +168,22 @@ def from_env(cls):
 Alternatively, you can instaniate Client(user, password) directly.""")
         return cls(email=email, password=password, url=url)
 
+    def request(self, method, url, data=None, raw_response=False, **kwargs):
```
Mr0grog: Reading through this, I just noticed that it might be nice to handle `self._api_url` here, too, e.g:

```py
if not url.startswith('http://') and not url.startswith('https://'):
    url = f'{self._api_url}{url}'
```

web_monitoring/db.py
mrotondo: No, you're right - I (and Python) let myself do the easy thing. I split the core logic into a request method and added a request_json wrapper that all but this instance use.

web_monitoring/db.py
mrotondo: Done

web_monitoring/db.py
mrotondo: Done

web_monitoring/db.py
Mr0grog: Oops, didn’t realize I never replied to this! No, it won’t automatically do streaming/newline-delimited JSON, so you’d still have to do the extra work you are doing for the `POST /imports` endpoint. So keeping it the way you have it is all good — I guess it simplifies the choice of which parameter to use. :)

PR #509, opened 2019-11-02
Use dateutil for date parsing, drop pandas

### LINKED ISSUE #360: Use `dateutil` for Date Parsing Everywhere
In some parts of the codebase, we are using the `dateutil` package, while in others we are using `pandas` (!) and in some places, I think we might still be calling `strptime()`. (Noticed this issue while looking at #356.) We should make audit all our files and make sure to use `dateutil` everywhere.

Additionally, in places where we know a date is an ISO 8601 date (like parsing responses in `web_monitoring.db`), we should use `dateutil.parser.isoparse()` to make sure there’s never any ambiguity.

This requires a little spelunking through the codebase (note: we don’t need to worry about the `.ipynb` [Jupyter/iPython Notebook] files for this), but should *mostly* be a straightforward search for date parsing locations and replace with calls to `dateutil.parser.parse()`.

Mr0grog:
I felt spurred on to just do this by #507 today.

The only actual remaining use of pandas in our codebase was for parsing dates, which is a little overkill! We also already depended on dateutil, so we should actually have used it in all the places we are parsing dates. This does so, and since there are no remaining uses of pandas, drops it from the dependencies.

A fix to how we handled dates in CDX searches came along for the ride, since I realized we were kind of failing to respect time zone info if it was present on a search date. 😓

(^^^ cc @danielballan since you are splitting this code out into a separate package.)

Fixes #360.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,7 +4,6 @@ dependencies:
 - python-dateutil
 - docopt
 - lxml
-- pandas
 - requests
 - toolz
 - tornado
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,6 @@ git+https://github.com/anastasia/htmldiffer@develop
 git+https://github.com/danielballan/htmltreediff@customize
 html5-parser ~=0.4.8 --no-binary lxml
 lxml ~=4.4.1
-pandas ~=0.25.1
 sentry-sdk ~=0.13.1
 requests ~=2.22.0
 toolz ~=0.10.0
--- a/web_monitoring/cli.py
+++ b/web_monitoring/cli.py
@@ -39,11 +39,11 @@ and results between them.
 
 from collections import defaultdict
 from datetime import datetime, timedelta
+import dateutil.parser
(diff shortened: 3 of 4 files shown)


danielballan:
Thanks. I think (as much as I can recall) that I reached for pandas because I know its time-parsing idiosyncracies well. But dateutil is a much better choice: a lighter dependency and the _de facto_ standard choice for this nowadays, even more so since its lead maintainer was made a CPython core dev.

Mr0grog:
Actually, I think *I* was the perpetrator of this (at least in this particular spot). I remember doing it — not sure if we weren’t yet using `dateutil` or if I just hadn’t seen it, but I remember thinking Pandas was workable since we were already depending on it.

Mr0grog:
Yep, my fault :P https://github.com/edgi-govdata-archiving/web-monitoring-processing/commit/e0622fc7a9f2b6b50b7072ceab1124e0bb36dcfa

danielballan:
Ha.

PR #511, opened 2019-11-18
Use Wayback package and remove internetarchive module

Mr0grog:
Now that @danielballan has published an alpha version of `wayback`, this PR makes use of it. :)

Still to do:

- [x] Update README
- [x] Update all usage in our scripts
- [x] Add tests for new functionality in scripts that is untested
- [x] Remove internetarchive tests
- [x] Remove internetarchive docs
- [ ] Determine whether we can handle `original_url_for_memento` better (depends on edgi-govdata-archiving/wayback#6 )


web_monitoring/cli.py
```diff
@@ -53,7 +53,9 @@
 from tqdm import tqdm
 from urllib.parse import urlparse
 from web_monitoring import db
-from web_monitoring import internetarchive as ia
+import wayback
+# FIXME: importing a private API like this is not great :(
+from wayback._client import original_url_for_memento
```
Mr0grog: @danielballan any thoughts on how to best handle this?

- It might be nice to expose as a helper, but
- We could also solve the use-case here by adding the data `original_url_for_memento` gets onto the response objects returned by `get_memento`.

Mr0grog:
OK, pending test success and a plan for `original_url_for_memento`, this should be good to go.

web_monitoring/cli.py
danielballan: Utility functions are inherently more reusable than utility methods. I like to consider the question, "Does this method need access to any internal state?" If no, lean toward function. Exposed publicly in https://github.com/edgi-govdata-archiving/wayback/pull/6

Mr0grog: That works!

FWIW, I was thinking of it more like attributes rather than exposing a method:

```py
memento = client.get_memento('http://web.archive.org/web/20170813195036/https://arpa-e.energy.gov/?q=engage/events-workshops')
assert memento.archived_url == 'https://arpa-e.energy.gov/?q=engage/events-workshops'
assert memento.date == datetime.datetime(2017, 8, 13, 19, 50, 36)
```

Mr0grog: (Obviously that still doesn’t need state, but makes a cleaner API since it’s common/relevant info a consumer might want, and definitely the sort of data we’d want to expose in edgi-govdata-archiving/wayback#2.)

danielballan: That's a solid point. Might be worth revisiting https://github.com/edgi-govdata-archiving/wayback/pull/6 again.

danielballan:
This needs to be updated to use `memento_url_data`, which is is available from the just-released https://pypi.org/project/wayback/0.2.0a2/.

danielballan:
I suggest this order of operations:

- [ ] Update this PR to use v0.2.0rc1 (published to PyPI just now) and confirm that it passes.
- [ ] Tag v0.2.0 final.
- [ ] Update thie PR again to remove the dependency on an alpha.
- [ ] Merge this.

Mr0grog:
@danielballan Updated to v0.2.0rc1. Feel free to proceed with the rest if/when CI passes.

Mr0grog:
Deployed to our ETL server; I’ll check tonight to make sure the nightly Internet Archive import went fine.

Mr0grog:
This PR should not have been merged and has been reverted in 5338ef135d376615c6bba1296e5379a4237ece9d. I’ll open a new PR with fixed code.

PR #512, opened 2019-12-01
Use `wayback` package, remove `internetarchive` module, take 2

Mr0grog:
This is a second take on #511, which was incomplete and had to be reverted. What’s different here?

- Fixed references to error types. These require importing from a private part of `wayback`, so we might want to merge and release edgi-govdata-archiving/wayback#7, which provides public access to those, first.
- Fixed some broken method calls.
- Fix the fact that `date` got renamed to `timestamp` on `CdxRecord` right before the 0.2.0 release.
- Add an end-to-end test that runs the whole import process, and covers some of the error cases we encounter from the various clients. It’s not perfect and doesn’t cover all errors, but was a lot quicker to get done than a completely custom test with mock client instances that purposely trigger all the various possible errors.


Mr0grog:
Going to add a custom user agent to all our sessions so edgi-govdata-archiving/wayback#18 is safe, then will merge.

PR #524, opened 2019-12-04
Reset connections in some cases for CDX failures

Mr0grog:
This is a better solution than retrying a bagillion times like we previously did. This fix has actually been patched onto our production code for a few weeks, and I somehow failed to backport it to the actual main codebase!

We also do the same technique in `WaybackRecordsWorker`. It feels imperfect, but I’m probably going to go ahead and merge for now just so production is no longer a special snowflake.


Mr0grog:
After some discussions with @danielballan, we have *strong* suspicions some explicit `requests.Response.close()` might solve the underlying issue. Merging this first, but then we’ll explore that fix after.

PR #528, opened 2019-12-05
Reorganize content of web_monitoring into directories

Mr0grog:
This reorganizes the content of `web_monitoring` into a hierarchy of modules with several directories based on this discussion: https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/206#issuecomment-561975711

```
web_monitoring/
    tests/
        [same as today]
    diff/
        content_type.py
        differs.py
        diff_errors.py
        html_diff_render.py
        links_diff.py
    diff_server/
        server.py
    cli/                     # Directory to hold CLI command script stuff
        cli.py               # Entrypoint -- parses args and calls other modules
        ia_healthcheck.py    # Currently in scripts folder
        ia_import.py         # Most of what's currently in cli.py
    __init__.py
    _version.py
    utils.py
    db.py
```

- [x] `diff`
- [x] `diff_server` (I renamed the module itself to `diff_server/server.py`)
- [x] `cli`
    - [x] Main `cli.py` file
    - [x] `ia_healthcheck` script
    - [x] `annotations_import` script

As part of moving the scripts, I’d like to reconfigure our CLI setup so we only have two scripts: `wm` and `wm-diff-server`. The `wm` command currently has `wm import ia` and `wm import ia-known-pages` subcommands, and I’m thinking I’d add `wm ia-healthcheck` and `wm annotations-import`. I think there might be room to rethink the command names and structure more, but that’s probably out of scope for this PR (making this script change is already questionably big).

Out of scope, but probably things we might want to do after this PR:

- Split up the contents of `diff_server/server.py`.
- More relative imports everywhere.
- Reorganize/rename some of the stuff in `diff`.
- Break apart the insanity in `cli/cli.py` a bit.
- Consider whether any of these new directories should have `__init__.py` files for easier access to subpackage contents.


Mr0grog:
@danielballan Would it be a good idea to make the tests directory parallel the structure of these new subpackages?

Mr0grog:
Actually, nevermind, I don’t think I want to merge the healthcheck and annotation import scripts into the main `wm` script in this PR.

I still think that’s worth doing, but I realized I probably want to drop docopt and go back to argparse (unless somebody has a strong recommendation for click or cliff or whatever else). Better to keep this PR simple and quick — get things mostly in the right place *on disk* now so it’s minimally disruptive, *then* do changes to the code itself in separate PR.

PR #531, opened 2019-12-26
Change Python version from 3.6 to 3.7 in docs

vbanos:


Mr0grog:
Have we added anything that actually requires 3.7? I see we are testing in 3.7, but `setup.py` still notes 3.6.

vbanos:
Yes, I can mentioned at least one thing. 
We use `queue.SimpleQueue` https://docs.python.org/3/library/queue.html#queue.SimpleQueue
which was introduced in Python 3.7.
This is where its used: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/utils.py#L107

vbanos:
Done!

PR #550, opened 2020-02-06
HOTFIX: Don't use domain prefix queries for CDX

Mr0grog:
When we have lots of pages at a given domain, we used to issue a single CDX query for everything at that domain, rather than a separate query for each URL we are tracking. However, it turns out this doesn't actually work:
- Some indexes are excluded from prefix queries in CDX, so you won't get URLs included in those indexes.
- Wayback folks think there might also be a bug causing some pages to not get included in prefix queries when they should.

So in order to make sure we are actually getting data about *all* the pages we track, we can't use the old optimization of querying by domain. For now, just set `MAX_QUERY_URLS_PER_DOMAIN` to an absurdly high number that we'll never reach.


Mr0grog:
For relevant discussion with Wayback team, see https://internetarchive.slack.com/archives/C92PYHLCE/p1580161598004300

Issue #525, opened 2019-12-04
Handle too many connections to Wayback automatically


Mr0grog:
Long ago, we worked around an issue where we were getting lots of connection failures from Wayback with a dirty hack: if we ran out of retries but still had a failure to establish a new connection, we’d try resetting the whole session (only once!) and start over: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/74b31b361c1f05836adbf7a25006c7417c453c01/web_monitoring/cli.py#L264-L273

At the time, I knew this was a kind of ugly hack, and probably masking some underlying issues. After some discussion with @danielballan and some spelunking through the source of `requests` and `urllib3` today, we realized there were two problems:

1. We might be holding open connections for an unnecessarily long time. This can be fixed by explicitly closing our response objects.

    - [x] @danielballan did this for `WaybackClient.search` in edgi-govdata-archiving/wayback#19

    - [x] I need to do this when we load Mementos here in `cli.py`: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/74b31b361c1f05836adbf7a25006c7417c453c01/web_monitoring/cli.py#L260-L263

        *(Update: we reliably read the whole response, which closes the connection, so we are actually OK here.)*

2. But that doesn’t cover everything. Ultimately, `requests` buries some features around connection pooling from `urllib3` in a way that means each `WaybackSession` (we have one per thread, for 36 in production right now) could possibly create infinity connections and then hold onto up to 10. (so we could wind up attempting to hold 360 open connections to Wayback!) Some things we can do to be smarter about this:

    - [x] Just reduce the number of threads in production! *Already on this, but it doesn’t make the problem obvious or clear if you scale up too far. We should ultimately do better.*

    - [x] When `get_memento()` redirects, make sure we read the body and close the response before moving onto the next in the redirect chain. (edgi-govdata-archiving/wayback#20)

    - [ ] Automatically step down the number of memento threads (to a point) if one gets too many connections refused. (How Many is Too Many? Two? Five? Probably > 1 to differentiate between something spurious and Wayback actually telling us we have too many connections open. Then again, maybe the retry configuration can be partially responsible for this? Much nuance here.) We have to make sure we re-queue the `CdxRecord` that was being loaded in this case. The error we’d see to trigger this would include `Failed to establish a new connection: [Errno 111] Connection refused`

[...]

PR #551, opened 2020-02-06
Use a shared HTTPAdapter across all threads

Mr0grog:
I spent an absurd amount of time over the last few days trying different approaches to better managing connections and errors with Wayback. Part of that was prompted by #525 and #507 

This basically rigs up a situation where we create a single `HTTPAdapter` for use across all our sessions (one session per thread). We then set the maximum number of connections in the adapter to the number of threads we are using, and tell it to block (if you don’t tell it to block, it’ll just go ahead and make more connections than the max anyway, which, ??? is kinda weird but hey). This has a huge impact on error rate and performance:

**Before:**
- Takes roughly 13-16 minutes to download 500 mementos from Wayback.
- Has an error rate of 20-30% (we see lower error rates at the end of the day because we go back and retry everything several times).

**After:**
- Takes roughly 8-9 minutes to download 500 mementos from Wayback
- Has an error rate of 0.0-1.2%

(Tested on my local machine at the office with a pretty speedy connection, alternating between different approaches to hopefully ensure they all got tested under similar load conditions at the Internet Archive.)

Why not just share a single Session or Client? Requests is apparently not thread-safe, and it seems like its authors have been struggling with small, niggling issues around thread-safety for a while. It’s not great to be sharing an [`HTTPAdapter`](https://github.com/psf/requests/blob/master/requests/adapters.py) across threads, but it’s a small interface and doesn’t interact much with the rest of requests. This *seems* OK, and I haven’t run into any issues yet. Going any lower means wrapping parts of urllib3, which makes me feel like “why are we even using requests?” So I figured this is the sweet spot for now.

Along the way, I also implemented an AIMD (Additive Increase/Multiplicative Decrease) sizing algorithm for pools (this is the algorithm typically used for [managing TCP window sizes](https://en.wikipedia.org/wiki/TCP_congestion_control) to optimize network congestion). It turns out that it didn’t seem to perform a whole lot better than what I ultimately did here, and it’s much more complicated to implement, so I left it out for now. Maybe it’ll come back someday! ¯\\\_(ツ)\_/¯

**There’s still a little cleanup to do to make this final, so this is not yet ready to merge.** (See all the commented out bits.)

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -233,7 +233,6 @@ class WaybackRecordsWorker(threading.Thread):
         self.adapter = adapter
         session_options = session_options or dict(retries=3, backoff=2,
                                                   timeout=(30.5, 2))
-        # session = wayback.WaybackSession(user_agent=USER_AGENT, **session_options)
         session = CustomAdapterSession(adapter=adapter,
                                        user_agent=USER_AGENT,
                                        **session_options)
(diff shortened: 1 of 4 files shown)


Mr0grog:
Oh, also of note here: after *lots* of digging around in requests and urllib3 source code, I really don’t think our `reset()` method was doing anything useful other than paving over our connection management problems. That’s why it’s commented out here. HOWEVER, that makes this PR dependent on all the recent updates we’ve made in the [`wayback` package](https://github.com/edgi-govdata-archiving/wayback) to fix places we held onto open connections.

web_monitoring/cli/cli.py
```diff
@@ -204,7 +236,7 @@ def run(self):
 
             self.handle_record(record, retry_connection_failures=True)
 
-        self.wayback.close()
+        # self.wayback.close()
```
danielballan: It makes sense that if an `adapter` is passed in to `__init__` it's the user's job to close it (as written now) because that enables sharing it between instances for parallel use. But if `adapter=None` and we create our own for this instance, should we close it here?

Mr0grog: Ah yeah, that’s a good point!

Mr0grog:
Had to add rate limiting here because of the current situation with the Internet Archive. Like the rest of this stuff, we should eventually find a way to stuff that functionality down into the Wayback package itself.

Mr0grog:
Rebased on top of #553.

Mr0grog:
This has been running well for the past week or so. Going to remove the commented out bits and merge tomorrow, ugly as this may be.

Mr0grog:
Took me a little longer than a day to get back to this. But it’s merged now!

PR #552, opened 2020-02-13
Add libxml2 & libxslt to install instructions

Mr0grog:
Add libxml2 and libxslt to the installation instructions. @Frijol was trying to get this project set up and and got cryptic errors because these weren’t installed. We should have remembered to mention them.


Mr0grog:
Turns out this was not @Frijol’s main issue (still investigating that, but clearly some linking/include path issues of some sort), but this still seems like something we should have in the docs. Will merge soon if no other feedback.

PR #553, opened 2020-02-14
Track import errors by import ID

Mr0grog:
When the import script breaks things up into multiple individual imports, it's hard to figure out which imports triggered which error messages (if there were errors), and therefore hard to remediate those errors later. Instead of returning a list of all the errors from all the imports that we are monitoring, this now returns a dict mapping import IDs to errors, and displays it with one import job's error list per line.

It looks like:

```sh
$ scripts/wm import ia-known-pages --from 24 --pattern '*ncddc.noaa.gov/*'
importing: 21 versions [00:04,  2.04 versions/s]
Loaded 22 CDX records:
      21 successes (95.45%),
       1 could not be played back (4.55%),
       0 had no actual memento (0.00%),
       0 unknown errors (0.00%).
importing: 21 versions [00:04,  5.05 versions/s]
Import jobs IDs: (30307,30308)
Polling web-monitoring-db until import jobs are finished...
Database Import Processing Errors:
  30307: none
  30308: ["Row 8: Response body for 'http://web.archive.org/web/20200213233820id_/https://ncddc.noaa.gov/subjects/climatechange/sciencevideos.htm' did not match expected hash (29fae4211c797fb5acdd27bb0eb83cbf3dcd1035f2ee5ac73ea5c4233b482f8c)"]
```

### DIFF
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -544,8 +544,10 @@ Alternatively, you can instaniate Client(user, password) directly.""")
                         continue
                     data = result['data']
                     if data['status'] == 'complete':
-                        errors[import_id] = data['processing_errors']
                         import_ids.remove(import_id)
+                        job_errors = data['processing_errors']
+                        if job_errors:
+                            errors[import_id] = job_errors
                 time.sleep(1)
         except KeyboardInterrupt:
             ...
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -515,21 +515,21 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         """
         Poll status of Version import jobs until all complete.
 
-        Use Ctrl+C to exit early. A list of the errors (so far) will be
-        returned.
+        Use Ctrl+C to exit early. A dict mapping the import IDs to any errors
+        from those imports (so far) will be returned.
 
         Parameters
         ----------
-        import_ids : collection
+        import_ids : iterable of (str or int)
         stop : threading.Event, optional
             A threading.Event to monitor in order to determine whether to stop
             monitoring before all imports are complete.
 
         Returns
         -------
-        errors : tuple
+        errors : dict of {str or int : list}
         """
-        errors = []
+        errors = {}
         import_ids = list(import_ids)  # to ensure mutable collection
         try:
             while import_ids and (stop is None or not stop.is_set()):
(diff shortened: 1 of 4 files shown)


web_monitoring/db.py
```diff
@@ -515,21 +515,21 @@ def monitor_import_statuses(self, import_ids, stop=None):
...
 
-        Use Ctrl+C to exit early. A list of the errors (so far) will be
-        returned.
+        Use Ctrl+C to exit early. A dict mapping the import IDs to any errors
+        from those imports (so far) will be returned.
 
         Parameters
         ----------
-        import_ids : collection
+        import_ids : iterable of (str or int)
```
danielballan: This is perfectly readable, and I'm not aware of a convention for this. There is a gradual migration to using the language of type hints here, which would express this as `Union[Iterable[str], Iterable[int]]`. I have mixed feelings about that movement; is gains something in specificity but loses something in readability to novices.

Mr0grog: Ah, ok. I’ll stick With this, then, especially since we aren’t actually using type annotations right now.

Mr0grog:
Slightly tweaked the output:
- Use the same language to talk about IDs and errors (“import job”).
- Only list jobs that errored (I tried this in an actual run and all the `<job_id>: None` lines were just waaaay too much noise).
- Show count of errors for each job before listing the errors.
- Show a total count of errors at the end.

```sh
$ scripts/wm import ia-known-pages --from 24 --pattern '*ncddc.noaa.gov/*'
importing: 21 versions [00:04,  2.04 versions/s]
Loaded 22 CDX records:
      21 successes (95.45%),
       1 could not be played back (4.55%),
       0 had no actual memento (0.00%),
       0 unknown errors (0.00%).
importing: 21 versions [00:04,  5.05 versions/s]
Import job IDs: (30307,30308)
Polling web-monitoring-db until import jobs are finished...
Import job errors:
  30308:   1 errors ["Row 8: Response body for 'http://web.archive.org/web/20200213233820id_/https://ncddc.noaa.gov/subjects/climatechange/sciencevideos.htm' did not match expected hash (29fae4211c797fb5acdd27bb0eb83cbf3dcd1035f2ee5ac73ea5c4233b482f8c)"]
  Total:   1 errors
```

I think this is ready to go.

Issue #563, opened 2020-03-18
Imports from Wayback list the final redirect URL twice


Mr0grog:
When we import versions from Wayback to web-monitoring-db using the script in [`web_monitoring/cli/cli.py`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/cli/cli.py), we are listing the final URL in a redirect chain twice. See this version for an example: https://api-staging.monitoring.envirodatagov.org/api/v0/versions/47951a31-ff31-44b2-b910-3c41278a9399

This is probably a bug in `format_memento()`: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/3f731b6a7662e58e0ded222b65646302778e28ef/web_monitoring/cli/cli.py#L290-L339

PR #564, opened 2020-03-20
Validate URLs in CLI importer

### LINKED ISSUE #543: `wm import ia` should validate URL before trying to import
I accidentally just swapped around the date and URL arguments and got some funky failed importing. It would be better if the command validated that the URL it was trying to load looks like at actual URL (e.g. starts with `https?://`) and stopped early with a nice error if not.

julianmclain:
This looked like a good first issue, so I took a shot at it. Let me know what you think.

In terms of validation, I wasn't sure exactly what you need. For example, are you only looking at specific TLDs? Is protocol required? etc... 

I decided to use `urlparse` and just confirm that the url has a valid protocol and network location (which apparently doesn't require a TLD being present as I learned).

### DIFF
--- a/web_monitoring/tests/test_cli.py
+++ b/web_monitoring/tests/test_cli.py
@@ -6,7 +6,7 @@ import vcr
 from wayback import WaybackClient
 from web_monitoring.cli.cli import (_filter_unchanged_versions,
                                     WaybackRecordsWorker, import_ia_db_urls,
-                                    _all_urls_valid)
+                                    _is_valid)
 
 
 # The only matters when re-recording the tests for vcr.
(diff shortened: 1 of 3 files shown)


julianmclain:
Thanks for the review! Ok, I like the idea of getting rid of `_all_urls_valid()` too.

web_monitoring/tests/test_cli.py
```diff
@@ -48,6 +49,20 @@ def test_filter_unchanged_versions():
...
+    valid_urls = ['https://example.com', 'http://example.com/interesting/path', 'https://subdomain.example.com']
+    valid_url_missing_domain = ['https://google']
+    invalid_single_url = ['asldkfje']
+    invalid_and_valid_urls = ['https://exmaple.com', 'http:badprotocol.com']
+
+    assert all(_is_valid(url) for url in valid_single_url)
+    assert all(_is_valid(url) for url in valid_urls)
+    assert all(_is_valid(url) for url in valid_url_missing_domain)
+    assert not all(_is_valid(url) for url in invalid_single_url)
+    assert not all(_is_valid(url) for url in invalid_and_valid_urls)
```
Mr0grog: Now that this function isn’t working on a list, I’m not sure it makes sense to be testing lists of things here. It’s fine as-is, but would be nicer if these were just:

```py
assert _is_valid(some_valid_url)
assert not _is_valid(some_invalid_url)
# etc.
```

julianmclain: Ah yeah, you're right. That's a quick fix, let me tidy up before we merge.

julianmclain:
Ok - fixed the tests. 

I'd love to be listed in the contributors section! I'd also be curious which issue you think would be good to pick up next. `Imports from Wayback list the final redirect URL twice #563` from a few days ago also looked like a good candidate. I'll shoot you a DM in the Archivers Slack.

Mr0grog:
Ugh, sorry it took me a few days to get back around to this. Looks great — I’ll merge and add you to the contributor list.

#563 would be a great one to fix, and should be pretty straightforward. :)

PR #565, opened 2020-03-27
Remove duplicate final redirect url from redirects

### LINKED ISSUE #563: Imports from Wayback list the final redirect URL twice

julianmclain:
Your intuition was right - in `format_memento()` the final url was being added to the list of redirects even though it's already included as the last entry in memento's `history`.

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -354,7 +354,6 @@ class WaybackRecordsWorker(threading.Thread):
                 lambda response: wayback.memento_url_data(response.url)[0],
                 memento.history))
             redirected_url = wayback.memento_url_data(memento.url)[0]
-            redirects.append(redirected_url)
             metadata['redirected_url'] = redirected_url
             metadata['redirects'] = redirects
 
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -354,8 +354,10 @@ class WaybackRecordsWorker(threading.Thread):
                 lambda response: wayback.memento_url_data(response.url)[0],
                 memento.history))
             redirected_url = wayback.memento_url_data(memento.url)[0]
+            redirects.append(redirected_url)
             metadata['redirected_url'] = redirected_url
-            metadata['redirects'] = redirects
+            metadata['redirects'] = [url for i, url in enumerate(redirects)
+                                     if url not in redirects[:i]]
 
         return dict(
             # Page-level info
--- a/web_monitoring/tests/test_cli.py
+++ b/web_monitoring/tests/test_cli.py
@@ -109,9 +109,9 @@ def test_format_memento_handles_redirects():
 
         assert isinstance(version, dict)
         assert version['source_metadata']['redirected_url'] == final_url
-        assert len(version['source_metadata']['redirects']) == 3
+        assert len(version['source_metadata']['redirects']) == 2
         assert version['source_metadata']['redirects'][0] == url
-        assert version['source_metadata']['redirects'][2] == final_url
+        assert version['source_metadata']['redirects'][1] == final_url
 
 
 # TODO: this test covers some of the various error cases, but probably not all


web_monitoring/cli/cli.py
```diff
@@ -354,7 +354,6 @@ def format_memento(self, memento, cdx_record, maintainers, tags):
                 lambda response: wayback.memento_url_data(response.url)[0],
                 memento.history))
             redirected_url = wayback.memento_url_data(memento.url)[0]
-            redirects.append(redirected_url)
```
Mr0grog: Something seemed kind of fishy about this, so I dug in a bit more, and it turns out this is not quite right.

First, I ran a quick script to grab some examples:

```py
from web_monitoring import db
from itertools import islice

def versions_with_redirects():
    client = db.Client.from_env()
    chunk = 1
    while chunk > 0:
        versions = client.list_versions(sort=['capture_time:desc'], chunk=chunk)
        for version in versions['data']:
            if 'redirects' in version['source_metadata']:
                yield version
        chunk = 1 if versions['links']['next'] else -1

# Look at the most recent 10
for version in islice(versions_with_redirects(), 10):
    meta = version['source_metadata']
    print(f'{version["uuid"]}')
    print(f'  Actual Wayback URL: {meta["original_url"]}')
    print('  Redirects:')
    for url in meta['redirects']:
        print(f'    {url}')
```

Which printed:

```
7eff0f06-123e-4f03-9f03-0851de7a0b2b
  Actual Wayback URL: http://web.archive.org/web/20200327033521id_/https://www.bia.gov/WhoWeAre/BIA/OTS/NaturalResources/FishWildlifeRec/index.htm
  Redirects:
    https://www.bia.gov/WhoWeAre/BIA/OTS/NaturalResources/FishWildlifeRec/index.htm
    https://www.bia.gov/bia/ots/division-natural-resources/branch-fish-wildlife-recreation
    https://www.bia.gov/bia/ots/division-natural-resources/branch-fish-wildlife-recreation
78431de2-d4d5-4410-9b87-7d4acbef2b1f
  Actual Wayback URL: http://web.archive.org/web/20200327031824id_/https://science.energy.gov/sbir/reporting-fraud/
  Redirects:
    https://science.energy.gov/sbir/reporting-fraud/
    https://www.energy.gov/science/office-science
    https://www.energy.gov/science/office-science
68e2f13a-0618-40e1-a36d-5b48362b6f40
  Actual Wayback URL: http://web.archive.org/web/20200327020641id_/https://science.energy.gov/news/highlights/?perf=University&p=1
  Redirects:
    https://science.energy.gov/news/highlights/?perf=University&p=1
    https://www.energy.gov/science/listings/science-highlights?perf=University&p=1
# ...etc...
```

The first two have the problematic duplication we're trying to fix, but the third doesn’t! So it can’t quite be as simple as just not appending the last item.

So I grabbed the first one and ran `get_memento()` on it to see what the actual data we would have sent to `format_memento()` would be…

```py
from wayback import WaybackClient
wb = WaybackClient()
memento = wb.get_memento('http://web.archive.org/web/20200327033521id_/https://www.bia.gov/WhoWeAre/BIA/OTS/NaturalResources/FishWildlifeRec/index.htm')
for item in memento.history:
    print(item.url)
print(memento.url)
>>> http://web.archive.org/web/20200327033521id_/https://www.bia.gov/WhoWeAre/BIA/OTS/NaturalResources/FishWildlifeRec/index.htm
>>> http://web.archive.org/web/20200327033521id_/https://www.bia.gov/bia/ots/division-natural-resources/branch-fish-wildlife-recreation
>>> http://web.archive.org/web/20200318032153id_/https://www.bia.gov/bia/ots/division-natural-resources/branch-fish-wildlife-recreation
```

It turns out the last two are actually different URLs, not duplicates! We’re grabbing just the URL that the Wayback machine snapshotted and throwing out the date, but forgetting that might result in seeing the same URL twice in a row. (For more on why that happens, check out these docs in the Wayback package: https://github.com/edgi-govdata-archiving/wayback/blob/5301daf0f3ed8676418977f80790aa059abb3ae0/wayback/_client.py#L675-L699)

Anyway, I think there are two things to be done here:

1. The root problem is that `memento.history` probably should not have included the second of the three URLs (because it’s not actually a memento of the redirect, it’s just an artifact of how the Wayback Machine does lookups). That should be fixed in the [Wayback package](https://github.com/edgi-govdata-archiving/wayback) itself.

2. But this package should probably be resilient against the bug by just not adding duplicate entries to `metadata['redirects']`. This might be one kind of ugly way to do it:

    ```py
    metadata['redirects'] = [url for i, url in enumerate(redirects)
                             if redirects[i - 1] != url]
    ```

Mr0grog: I made an issue over on `wayback` for (1): https://github.com/edgi-govdata-archiving/wayback/issues/36

julianmclain: I should have known that was too easy to be true!

I'm still trying to wrap my head around the logic for when a memento is a redirect. The documentation in the Wayback package has been super helpful. Also, thanks for including your script to fetch versions with redirects. I'll step through `get_memento()` with a few examples which should make it more concrete.

I don't think removing duplicates gets much better than your 1-liner. With a small tweak we could make it a more general sanity check that removes duplicates at any position in the list:
```
metadata['redirects'] = [url for i, url in enumerate(redirects)
						 if url not in redirects[:i]]
```

Mr0grog: Yeah, dealing with redirects in Wayback in kinda 🤯

PR #575, opened 2020-04-02
Immediately validate DB credentials in wm CLI

julianmclain:
👋 It sounded like you just need something quick and dirty for [#58](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/58), so I took a stab at it. Wanted to get it in front of you before I polish it up / add a test. LMK what you think.

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -828,18 +828,27 @@ def _is_valid(url):
 
 
 def validate_db_credentials():
-    """Validate DB credentials by creating a client and hitting the /users/sessions route"""
+    """
+    Validate Web Monitoring DB credentials by creating a temporary client and
+    attempting to get the current user session.
+
+    Raises
+    ------
+    InvalidCredentials
+        If the credentials are not authorized for the provided host.
+    """
     test_client = db.Client.from_env()
     try:
         test_client.validate_credentials()
-    except db.InvalidCredentials as exc:
-        raise Exception("""
-            401 Unauthorized for the WEB_MONITORING_DB_URL provided, check the following environment
-            variables:
+    except db.UnauthorizedCredentials:
+        raise Exception(f"""
+            Unauthorized credentials for {test_client._base_url}.
+            Check the following environment variables:
 
                 WEB_MONITORING_DB_URL
-                WEB_MONITORING_DB_EMAIL
-                WEB_MONITORING_DB_PASSWORD""")
+                WEB_MONITORING_DB_EMAIL {os.environ['WEB_MONITORING_DB_EMAIL']}
+                WEB_MONITORING_DB_PASSWORD
+                """)
 
 
 def main():
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -829,8 +829,17 @@ def _is_valid(url):
 
 def validate_db_credentials():
     """Validate DB credentials by creating a client and hitting the /users/sessions route"""
-    credentails_test_client = db.Client.from_env()
-    credentails_test_client.validate_db_credentials()
+    test_client = db.Client.from_env()
+    try:
+        test_client.validate_credentials()
+    except db.InvalidCredentials as exc:
+        raise Exception("""
+            401 Unauthorized for the WEB_MONITORING_DB_URL provided, check the following environment
+            variables:
+
+                WEB_MONITORING_DB_URL
+                WEB_MONITORING_DB_EMAIL
+                WEB_MONITORING_DB_PASSWORD""")
 
 
 def main():
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -827,6 +827,12 @@ def _is_valid(url):
         return False
 
 
+def validate_db_credentials():
+    """Validate DB credentials by creating a client and hitting the /users/sessions route"""
+    credentails_test_client = db.Client.from_env()
+    credentails_test_client.validate_db_credentials()
+
+
 def main():
     doc = f"""Command Line Interface to the web_monitoring Python package
 
@@ -865,6 +871,7 @@ Options:
                   'or `resolved-response`')
             return
 
+        validate_db_credentials()
         if arguments['ia']:
             import_ia_urls(
                 urls=[arguments['<url>']],
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -834,14 +834,14 @@ def validate_db_credentials():
 
     Raises
     ------
-    InvalidCredentials
+    UnauthorizedCredentials
         If the credentials are not authorized for the provided host.
     """
     test_client = db.Client.from_env()
     try:
         test_client.validate_credentials()
     except db.UnauthorizedCredentials:
-        raise Exception(f"""
+        raise db.UnauthorizedCredentials(f"""
             Unauthorized credentials for {test_client._base_url}.
             Check the following environment variables:
 
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -118,6 +118,10 @@ class MissingCredentials(RuntimeError):
     ...
 
 
+class InvalidCredentials(RuntimeError):
+    ...
+
+
 class Client:
     """
     Communicate with web-monitoring-db via its REST API.
@@ -706,6 +710,23 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         data['updated_at'] = parse_timestamp(data['updated_at'])
         return result
 
+    ### USERS ###
+
+    def get_user_session(self):
+        """
+        Get the current user session
+
+        Parameters
+        ----------
+        none
+
+        Returns
+        -------
+        response : dict
+        """
+        user_session_url = f'{self._api_url[:-7]}/users/session'
+        return self.request_json(GET, user_session_url)
+
     ### CONVENIENCE METHODS ###
 
     def get_version_content(self, version_id):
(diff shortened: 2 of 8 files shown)


web_monitoring/db.py
```diff
@@ -706,6 +710,23 @@ def get_annotation(self, *, annotation_id, page_id, to_version_id,
...
+
+        Parameters
+        ----------
+        none
+
+        Returns
+        -------
+        response : dict
+        """
+        user_session_url = f'{self._api_url[:-7]}/users/session'
```
Mr0grog, line 727: I think it’d probably be better to save the `url` that was passed into the constructor separately from `_api_url` and use that here, rather than hacking off an magic number of characters.

web_monitoring/db.py
```diff
@@ -760,3 +781,25 @@ def get_version_by_versionista_id(self, versionista_id):
...
+
+        Returns
+        -------
+        none
+        """
+        try:
+            response = self.get_user_session()
+        except requests.exceptions.HTTPError as exc:
+            if exc.response.status_code == 401:
+                raise MissingCredentials(f'\n\n\n\nCheck your DB credentials - 401 Unauthorized for url: {exc.request.url}\n\n\n\n')
```
Mr0grog, line 801: This logic is probably better handled in `_process_errors`, since it really applies to all requests. It also might be better if this error were a little less “presentational” — without all the line breaks (and if we really want them, use a multi-line string in stead of `\n`). `cli.py` is a better place to get fancy with formatting.

This should probably also be `raise InvalidCredentials(...)`, right?

web_monitoring/db.py
```diff
@@ -760,3 +781,25 @@ def get_version_by_versionista_id(self, versionista_id):
...
+            else:
+                # Client didn't receive a 401, but hit another 4XX or 5XX error
+                raise RuntimeError(f'\n\n\n\nOops: an unexpected error occurred while checking DB credentials...\n\n{exc}\n\n\n\n')
```
Mr0grog, line 804: Similarly, I think it’s OK to just let other exceptions raise here without special handling. Providing nicer messaging should go in `validate_db_credentials()` in `cli.py`.

I think the ideal implementation for this whole method might just be:

```py
def validate_credentials(self):
    self.get_user_session()
```

And all the error handling should move to different places.

web_monitoring/cli/cli.py
```diff
@@ -827,6 +827,12 @@ def _is_valid(url):
         return False
 
 
+def validate_db_credentials():
+    """Validate DB credentials by creating a client and hitting the /users/sessions route"""
+    credentails_test_client = db.Client.from_env()
+    credentails_test_client.validate_db_credentials()
```
Mr0grog, line 833: Typo:

```suggestion
    credentials_test_client = db.Client.from_env()
    credentials_test_client.validate_db_credentials()
```

web_monitoring/db.py
```diff
@@ -760,3 +781,25 @@ def get_version_by_versionista_id(self, versionista_id):
         # Make result look like the result of `get_version` rather than the
         # result of `list_versions`.
         return {'data': result['data'][0]}
+
+    def validate_db_credentials(self):
```
Mr0grog, line 785: Given this is a method on the DB client, we can probably drop `_db_` from the name here.

julianmclain:
Thanks for the feedback! For some reason I was looking at making this "minimally invasive" vs. putting things where they probably belong best. I'll take a look at a 2nd draft over the weekend.

web_monitoring/db.py
julianmclain: Makes sense. What's your general thought on getting fancy with the formatting in `cli.py`? 👍 or 👎?

Mr0grog: 👍 Nice, explanatory, user-centric error messages are always good when surfaced from the script/CLI level! (Also good lower down; they should just be more focused more on developer rather than user ergonomics.)

Mr0grog:
> I was looking at making this "minimally invasive" vs. putting things where they probably belong best.

No worries. I’d rather get this more “right” than quick; the issue’s been sitting for quite a while so clearly doesn’t need a rush job. :)

web_monitoring/db.py
julianmclain: Good call, removed the magic

web_monitoring/db.py
julianmclain: Makes sense - moved the 'presentational' error handling to the cli and used a docstring instead of line breaks.

web_monitoring/db.py
julianmclain: Good call

julianmclain:
OK, I feel like we got a vast improvement with take 2. Let me know what you think!

web_monitoring/db.py
```diff
@@ -706,6 +714,23 @@ def get_annotation(self, *, annotation_id, page_id, to_version_id,
...
 
+    ### USERS ###
+
+    def get_user_session(self):
+        """
+        Get the current user session
+
+        Parameters
+        ----------
+        none
```
Mr0grog, line 725: No need to include this section if there’s nothing to call out.

web_monitoring/db.py
```diff
@@ -760,3 +785,17 @@ def get_version_by_versionista_id(self, versionista_id):
...
+        """
+        Validate that the DB Client is authorized for the provided host
+
+        Parameters
+        ----------
+        none
+
+        Returns
+        -------
+        none
```
Mr0grog, line 799: Same here — no need to include these sections if there’s nothing to put in them.

web_monitoring/db.py
```diff
@@ -706,6 +714,23 @@ def get_annotation(self, *, annotation_id, page_id, to_version_id,
         data['updated_at'] = parse_timestamp(data['updated_at'])
         return result
 
+    ### USERS ###
+
+    def get_user_session(self):
+        """
+        Get the current user session
```
Mr0grog, line 721: Please use full sentences with periods in the docstring.

web_monitoring/db.py
```diff
@@ -760,3 +785,17 @@ def get_version_by_versionista_id(self, versionista_id):
         # Make result look like the result of `get_version` rather than the
         # result of `list_versions`.
         return {'data': result['data'][0]}
+
+    def validate_credentials(self):
+        """
+        Validate that the DB Client is authorized for the provided host
```
Mr0grog, line 791: Same as above, full sentences. Also, I think the important bit to call out here is that this will raise an error in the case of invalid credentials, so it should be used kind of like `assert`.

You can also uses a `Raises` section to call out the specific error type if you want. (See the Numpy docstring styleguide: https://numpydoc.readthedocs.io/en/latest/format.html#sections)

julianmclain: The Numpy reference was super helpful, thanks for the link.

web_monitoring/db.py
```diff
@@ -36,7 +40,10 @@ def _process_errors(res):
     try:
         errors = res.json()['errors']
     except Exception:
-        res.raise_for_status()
+        if res.status_code == 401:
+            raise InvalidCredentials()
```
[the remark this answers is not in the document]
julianmclain: Done - I also renamed `InvalidCredentials` to `UnauthorizedCredentials`. Probably insignificant, but I think it's a better description

web_monitoring/cli/cli.py
```diff
@@ -827,6 +827,21 @@ def _is_valid(url):
         return False
 
 
+def validate_db_credentials():
+    """Validate DB credentials by creating a client and hitting the /users/sessions route"""
```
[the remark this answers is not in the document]
julianmclain: Noted here and in the `validate_credentials` method on the DB client.

julianmclain:
All good points on style - this one should look better. I also wanted to get your opinion on writing a test for the client and cli. Do you think they're worth adding? I haven't used vcrpy before, so I figured it's a good excuse to do some reading.

Mr0grog:
> I also wanted to get your opinion on writing a test for the client and cli. Do you think they're worth adding?

Definitely! I’m more concerned about a unit test for the client than a test focused on the CLI, but both would be nice.

> I haven't used vcrpy before, so I figured it's a good excuse to do some reading.

👍 VCR libraries are super useful, especially for complex request/response cycles, and it gets you closer to testing with “real” data than most mocks. (On the other hand, I am often on the fence for simpler cases, where manual mocks help clarify what’s happening/what the actual conditions being tested are. VCR can be very opaque in terms of what the meaningful parts of a server response are.)

web_monitoring/cli/cli.py
```diff
@@ -827,6 +827,30 @@ def _is_valid(url):
...
 
 
+def validate_db_credentials():
+    """
+    Validate Web Monitoring DB credentials by creating a temporary client and
+    attempting to get the current user session.
+
+    Raises
+    ------
+    InvalidCredentials
```
Mr0grog, line 837: This doesn’t match what you’re actually raising. ;)

This description should either change to `Exception` or it *and* the code should change to `UnauthorizedCredentials`.

julianmclain: Doh 🤦, will fix in the next commit.

julianmclain:
I could see that. I'm used to caveman copy-pasting to create mocks, which sucks but at least you understand how each request / response is interacting. Looking forward to checking it out, and I'll get a test together for both modules.

julianmclain:
Test are in, and fixed the mislabelled exception in the CLI.

julianmclain:
It's ready to go!

Mr0grog:
And so it is merged! :shipit: ⚓

PR #579, opened 2020-04-17
Use cURL-based HTTP client in diff server

Mr0grog:
This is effectively a revival of #293 (from @vbanos). The cURL HTTP client in Tornado is both faster and supports more funny business various HTTP servers do. I'm hoping this might also improve some of the weird memory & SSL errors we run into in production every so often.

Last time around, we had some concerns about whether `CURLOPT_MAXFILESIZE` would be good enough to restrict the size of responses and mitigate memory issues: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/293#issuecomment-431418167 Happily, we now have tests around this and it does (even though I did prototype a more invasive fix).

What still needs doing here:

- [x] Docker support
- [x] TravisCI support
- [x] Installation/setup docs (since this adds new system dependencies that can’t be installed with Pip)
- [x] Error handling (the cURL client raises different errors than the “simple” client, and all our handling is currently tailored for the “simple” client) (clearly we also need better tests around these since they should be failing!)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,6 +29,10 @@ export ACCESS_CONTROL_ALLOW_ORIGIN_HEADER="*"
 # Maximum diffable body size, in bytes.
 export DIFFER_MAX_BODY_SIZE='10485760' # 10 MB
 
+# Use Tornado's "simple" HTTP client to get diffable content. By default, the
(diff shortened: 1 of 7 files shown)


Mr0grog:
Assuming this passes tests, it ought to be good to go.

PR #580, opened 2020-04-21
Shut down the process pool safely

Mr0grog:
We've been seeing `OSError` getting raised from inside the differ process pool for a while, but it's always been really unclear what's actually going on. (It seemed to occur in conjunction with SSL errors, so for a while I had thought some really strange voodoo was going on.) It turns out I had naively trusted the documentation for `executor.shutdown()`, which is either unclear or incorrect. It implies `executor.shutdown(wait=False)` returns immediately, but doesn't tear down the executor until after all pending tasks are complete, just like `executor.shutdown(wait=True)`. This is not true! Setting `wait=False` will cause internal polling on the pending tasks to raise the `OSError`.

To work around this, I've added a thin shim that schedules the shutdown and waits on a thread (using the default executor). The result is awaitable, but not a coroutine, so you can trigger it and move on *or* `await` it if you need to know when it's done.

This should definitely resolve these two super common exceptions:
- https://sentry.io/organizations/environmental-data-governance-/issues/1084734796/?project=1317561&query=is%3Aunresolved
- https://sentry.io/organizations/environmental-data-governance-/issues/1512363487/?project=1317561&query=is%3Aunresolved

And I suspect those might be indirectly causing this one, so it should get resolved, too 🤞:
- https://sentry.io/organizations/environmental-data-governance-/issues/1132991547/?project=1317561&query=is%3Aunresolved&statsPeriod=14d
    (But if not, #579 should also address this one.)

### DIFF
--- a/web_monitoring/diff_server/server.py
+++ b/web_monitoring/diff_server/server.py
@@ -20,7 +20,7 @@ import traceback
 import web_monitoring
 from ..diff import differs, html_diff_render, links_diff
 from ..diff.diff_errors import UndiffableContentError, UndecodableContentError
-from ..utils import Signal
+from ..utils import shutdown_executor_in_loop, Signal
 
 logger = logging.getLogger(__name__)
 
@@ -200,10 +200,7 @@ class DiffServer(tornado.web.Application):
                 for child in differs._processes.values():
                     child.kill()
             else:
-                await tornado.ioloop.IOLoop.current().run_in_executor(
-                    None,
-                    lambda: differs.shutdown(wait=True)
-                )
+                await shutdown_executor_in_loop(differs)
 
     def handle_signal(self, signal_type, frame):
         """Handle a signal by shutting down the application and IO loop."""
@@ -428,7 +425,9 @@ class DiffHandler(BaseHandler):
         if reset or not executor:
             if executor:
                 try:
-                    executor.shutdown(wait=False)
+                    # NOTE: we don't need await this; we just want to make sure
+                    # the old executor gets cleaned up.
+                    shutdown_executor_in_loop(executor)
                 except Exception:
                     pass
             executor = concurrent.futures.ProcessPoolExecutor(
--- a/web_monitoring/utils.py
+++ b/web_monitoring/utils.py
@@ -1,3 +1,4 @@
+import asyncio
 import hashlib
 import io
 import logging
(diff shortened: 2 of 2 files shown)


Mr0grog:
For later reference, I used this script to verify the issue:

```py
from concurrent.futures import ProcessPoolExecutor
import time
import asyncio


def do_a_thing():
    time.sleep(1)
    finished_at = time.time()
    print(f'Done at: {finished_at}')
    return finished_at


async def stop_executor(executor):
    await asyncio.sleep(1.1)
    print('Shutting down...')
    try:
        # Change to to wait=True and everything is hunky-dory
        executor.shutdown(wait=False)
    except OSError:
        # This won't happen, since the error happens later, out of our control.
        print('OSError handled!')


async def main():
    executor = ProcessPoolExecutor(2)
    loop = asyncio.get_running_loop()
    print('Starting')

    # Start several slow tasks in a process pool
    tasks = [loop.run_in_executor(executor, do_a_thing) for x in range(10)]
    # Schedule a shutdown partway through the tasks
    results = await asyncio.gather(stop_executor(executor), *tasks)

    for index, result in enumerate(results):
        print(f'{index}: {result}')

    print('All done!')


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

Mr0grog:
For later reference, I used this script to verify the issue:

```py
from concurrent.futures import ProcessPoolExecutor
import time
import asyncio


def do_a_thing():
    time.sleep(1)
    finished_at = time.time()
    print(f'Done at: {finished_at}')
    return finished_at


async def stop_executor(executor):
    await asyncio.sleep(1.1)
    print('Shutting down...')
    try:
        # Change to to wait=True and everything is hunky-dory
        executor.shutdown(wait=False)
    except OSError:
        # This won't happen, since the error happens later, out of our control.
        print('OSError handled!')


async def main():
    executor = ProcessPoolExecutor(2)
    loop = asyncio.get_running_loop()
    print('Starting')

    # Start several slow tasks in a process pool
    tasks = [loop.run_in_executor(executor, do_a_thing) for x in range(10)]
    # Schedule a shutdown partway through the tasks
    results = await asyncio.gather(stop_executor(executor), *tasks)

    for index, result in enumerate(results):
        print(f'{index}: {result}')

    print('All done!')


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

Mr0grog:
We've been seeing `OSError` getting raised from inside the differ process pool for a while, but it's always been really unclear what's actually going on. (It seemed to occur in conjunction with SSL errors, so for a while I had thought some really strange voodoo was going on.) It turns out I had naively trusted the documentation for `executor.shutdown()`, which is either unclear or incorrect. It implies `executor.shutdown(wait=False)` returns immediately, but doesn't tear down the executor until after all pending tasks are complete, just like `executor.shutdown(wait=True)`. This is not true! Setting `wait=False` will cause internal polling on the pending tasks to raise the `OSError`.

To work around this, I've added a thin shim that schedules the shutdown and waits on a thread (using the default executor). The result is awaitable, but not a coroutine, so you can trigger it and move on *or* `await` it if you need to know when it's done.

This should definitely resolve these two super common exceptions:
- https://sentry.io/organizations/environmental-data-governance-/issues/1084734796/?project=1317561&query=is%3Aunresolved
- https://sentry.io/organizations/environmental-data-governance-/issues/1512363487/?project=1317561&query=is%3Aunresolved

And I suspect those might be indirectly causing this one, so it should get resolved, too 🤞:
- https://sentry.io/organizations/environmental-data-governance-/issues/1132991547/?project=1317561&query=is%3Aunresolved&statsPeriod=14d
    (But if not, #579 should also address this one.)

PR #581, opened 2020-04-28
Add default timeout to db client

julianmclain:
I didn't add `timeout` as an optional parameter to any of the client methods because I wasn't sure if it's needed anywhere yet. Let me know if it is, and I can add it wherever. This one looked like it could be a candidate: [monitor_import_statuses](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/a05fd623f5b983db84d4f2370197a61bd4276153/web_monitoring/db.py#L522)

### DIFF
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -12,6 +12,7 @@ import warnings
 
 
 DEFAULT_URL = 'https://api.monitoring.envirodatagov.org'
+DEFAULT_TIMEOUT = 30.5
 GET = 'GET'
 POST = 'POST'
 
@@ -148,12 +149,16 @@ class Client:
     url : string, optional
         Default is ``https://api.monitoring.envirodatagov.org``.
     """
-    def __init__(self, email, password, url=DEFAULT_URL):
+    def __init__(self, email, password, url=DEFAULT_URL, timeout=DEFAULT_TIMEOUT):
         self._api_url = f'{url}/api/v0'
         self._base_url = url
         self._session = requests.Session()
         self._session.auth = (email, password)
         self._session.headers.update({'accept': 'application/json'})
+
+        if timeout is not None and timeout <= 0:
+            raise ValueError(timeout)
+        self._timeout = timeout
 
     @classmethod
     def from_env(cls):
@@ -182,9 +187,12 @@ variables:
 Alternatively, you can instaniate Client(user, password) directly.""")
         return cls(email=email, password=password, url=url)
 
-    def request(self, method, url, data=None, **kwargs):
+    def request(self, method, url, data=None, timeout=None, **kwargs):
         if not url.startswith('http://') and not url.startswith('https://'):
             url = f'{self._api_url}{url}'
+
+        if timeout is not None and timeout <= 0:
+            raise ValueError(timeout)
 
         if data is not None:
             headers = kwargs.setdefault('headers', {})
@@ -194,12 +202,15 @@ Alternatively, you can instaniate Client(user, password) directly.""")
             else:
                 headers.update({'Content-Type': 'application/json'})
                 kwargs['data'] = json.dumps(data)
-        response = self._session.request(method=method, url=url, **kwargs)
+        response = self._session.request(method=method, 
+                                         url=url,
+                                         timeout=timeout if timeout is not None else self._timeout,
+                                         **kwargs)
         _process_errors(response)
         return response
 
-    def request_json(self, method, url, data=None, **kwargs):
-        response = self.request(method, url, data, **kwargs)
+    def request_json(self, method, url, data=None, timeout=None, **kwargs):
+        response = self.request(method, url, data, timeout, **kwargs)
         return response.json()
 
     ### PAGES ###
@@ -495,7 +506,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
             If true, don't import versions of a page that have the same hash as
             the version captured immediately before them.
         batch_size : integer, optional
-            Default batch size is 50000 Versions.
+            Default batch size is 1000 Versions.
 
         Returns
         -------
@@ -548,7 +559,7 @@ Alternatively, you can instaniate Client(user, password) directly.""")
         try:
             while import_ids and (stop is None or not stop.is_set()):
                 for import_id in tuple(import_ids):
-                    # We are mainly interrested in processing errors. We don't
+                    # We are mainly interested in processing errors. We don't
                     # expect HTTPErrors, so we'll just warn and hope that
                     # everything works in the second pass.
                     try:
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -160,7 +160,7 @@ class Client:
     url : string, optional
         Default is ``https://api.monitoring.envirodatagov.org``.
     timeout: float, optional
-        A connection timeout in seconds to be used for all requests. 0 indcates 
+        A connection timeout in seconds to be used for all requests. `0` indcates 
         no timeout. Individual method calls may override this value if `timeout`
         can be provided as an argument. The default value is 30.5 seconds.
     """
--- a/web_monitoring/db.py
+++ b/web_monitoring/db.py
@@ -160,9 +160,9 @@ class Client:
     url : string, optional
         Default is ``https://api.monitoring.envirodatagov.org``.
     timeout: float, optional
-        A connection timeout in seconds to be used for all requests. `0` indcates 
-        no timeout. Individual method calls may override this value if `timeout`
-        can be provided as an argument. The default value is 30.5 seconds.
+        A connection timeout in seconds to be used for all requests. `0` indicates 
+        no timeout should be used. Individual method calls may override this value
+        if `timeout` can be provided as an argument. The default value is 30.5 seconds.
     """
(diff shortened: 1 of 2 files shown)


web_monitoring/db.py
```diff
@@ -12,6 +12,7 @@
 
 
 DEFAULT_URL = 'https://api.monitoring.envirodatagov.org'
+DEFAULT_TIMEOUT = 30.5
```
julianmclain, line 15: I saw you used 30.5 in the [WaybackRecordsWorker](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/a05fd623f5b983db84d4f2370197a61bd4276153/web_monitoring/cli/cli.py#L235), so I went with that (guessing you chose it based on the recommendation to use something slightly bigger than a multiple of 3 over in [requests docs](https://requests.readthedocs.io/en/master/user/advanced/#timeouts)).

web_monitoring/db.py
```diff
@@ -489,7 +500,7 @@ def add_versions(self, versions, *, update='skip', create_pages=None,
             If true, don't import versions of a page that have the same hash as
             the version captured immediately before them.
         batch_size : integer, optional
-            Default batch size is 50000 Versions.
+            Default batch size is 1000 Versions.
```
julianmclain, line 503: This looked out of date based on the signature above:
```
    def add_versions(self, versions, *, update='skip', create_pages=None,
                     skip_unchanged_versions=None, batch_size=1000):
```

Mr0grog:
Hey @julianmclain, sorry I’ve been super slow to get to this. Our whole infrastructure has been in somewhat dire straits, and I’ve been focusing on that at the moment. I’ll try to make sure I get to this tomorrow at the latest. 🙏

julianmclain:
@Mr0grog Seriously no worries, that definitely sounds more important. This PR isn't going anywhere.

web_monitoring/db.py
Mr0grog: Ahhh, we use a super long timeout for Wayback because it is *very* slow for pages that aren’t accessed frequently. This is probably fine, though.

web_monitoring/db.py
```diff
@@ -142,13 +143,17 @@ class method :meth:`Client.from_env`.
     url : string, optional
         Default is ``https://api.monitoring.envirodatagov.org``.
     """
-    def __init__(self, email, password, url=DEFAULT_URL):
+    def __init__(self, email, password, url=DEFAULT_URL, timeout=DEFAULT_TIMEOUT):
```
Mr0grog, line 146: Make sure to add this to the class docstring above, too!

web_monitoring/db.py
```diff
@@ -142,13 +143,17 @@ class method :meth:`Client.from_env`.
...
         self._api_url = f'{url}/api/v0'
         self._base_url = url
         self._session = requests.Session()
         self._session.auth = (email, password)
         self._session.headers.update({'accept': 'application/json'})
 
+        if timeout is not None and timeout <= 0:
+            raise ValueError(timeout)
```
Mr0grog, line 154: This should have a more descriptive message. Also, I’m thinking `0` should probably be OK — it should imply “no timeout,” i.e. same as `None`. That is a pretty typical interpretation of `0` in a lot of timeout-related systems.

web_monitoring/db.py
```diff
@@ -176,10 +181,13 @@ def from_env(cls):
 Alternatively, you can instaniate Client(user, password) directly.""")
         return cls(email=email, password=password, url=url)
 
-    def request(self, method, url, data=None, **kwargs):
+    def request(self, method, url, data=None, timeout=None, **kwargs):
         if not url.startswith('http://') and not url.startswith('https://'):
             url = f'{self._api_url}{url}'
 
+        if timeout is not None and timeout <= 0:
+            raise ValueError(timeout)
```
Mr0grog, line 189: Same here for more informative messaging and handling `0`.

web_monitoring/db.py
Mr0grog: Ah, thanks for catching this. 50,000 is even above the [maximum size](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/d20d4ed9be6e7214a49895582c17ff5de05b3296/app/controllers/concerns/paging_concern.rb#L4-L5)!

web_monitoring/db.py
julianmclain: Good catch

web_monitoring/db.py
```diff
@@ -172,14 +181,31 @@ def from_env(cls):
...
         if not url.startswith('http://') and not url.startswith('https://'):
             url = f'{self._api_url}{url}'
 
+        if timeout is not None and timeout < 0:
+            raise ValueError(f'Timeout must be non-negative. "{timeout}" was provided.')
+        elif timeout == 0:
+            timeout = None
+        elif timeout is None:
+            timeout = self._timeout
+
```
julianmclain, line 208: I don't love this block of code but couldn't think of anything better.

web_monitoring/tests/test_db.py
```diff
@@ -292,3 +293,41 @@ def test_validate_credentials_should_raise():
...
+        method='GET', url=f'{AUTH["url"]}/users/session', timeout=7.5)
+
+
+@patch.dict('os.environ', clear=True)
+@patch('web_monitoring.db.requests.Session')
+def test_from_env_with_no_timeout(mock_session):
+    os.environ.update({'WEB_MONITORING_DB_URL': AUTH['url'],
+                       'WEB_MONITORING_DB_EMAIL': AUTH['email'],
+                       'WEB_MONITORING_DB_PASSWORD': AUTH['password'],
+                       'WEB_MONITORING_DB_TIMEOUT': '0'})
```
julianmclain, line 329: Just realizing how repetitive this is. Worth creating a re-usable dictionary?

Mr0grog: You can simplify these by adding keyword args to `update`, e.g:

```py
# dict.update(other_dict, additional='key' value='pairs')
os.environ.update(AUTH, WEB_MONITORING_DB_TIMEOUT=0)
```

julianmclain: Oh, nice! Learned something new.

julianmclain: On second glance, I think we'd still need to create a new dict or do a fancy transformation on `AUTH`. It's setup to be spread over the constructor args, i.e. `Client(**AUTH)`. 

Now that I can delete all that code for parsing a timeout ENV variable, I'll just go back to creating a `Client` with the constructor.

web_monitoring/db.py
julianmclain: I love the idea of abstracting it out into a separate function.

If we want to support 'no timeout' as a state, I think 0 is a good value for it. The `NO_VALUE = object()` reads really weirdly to me (although I'm not an expert). I'll let you decide. Same for whether we validate, urllib3 actually has pretty good error messages:
```
ValueError: Attempted to set connect timeout to -1, but the timeout cannot be set to a value less than or equal to 0
```

Mr0grog: > The `NO_VALUE = object()` reads really weirdly to me

Oh, for sure. You won’t see this too often, and it’s generally never meant to be a user-visible thing, but it’s a relatively common practice if you need to be able to distinguish between something being set to `None` and not being set at all. `object()` is the simplest way to get a unique value that will never be equal to anything else in Python.

We do something similar for our `FiniteQueue` implementation, but use a class so that we get nice names when debugging (because, unlike `NO_VALUE`, that one is likely to be more visible to users of the code): https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/93e9e7cb80ef85e973717df46f270c72cbcf1022/web_monitoring/utils.py#L167-L169

**Anyway! I am cool if you want `0` to represent “no timeout” and `None` to represent “I didn’t specify” so you don’t need to do the above. :)**

julianmclain: Ok that actually makes a lot of sense, thanks for the example. Looking at it kind of reminded me of how Scala uses a `Nil` singleton to represent empty lists. Here you're just defining your own singleton to represent `a value wasn't set`.

If you're still happy with `0`, I'm still happy with `0` :)

Mr0grog: Yep, all good. 👍

web_monitoring/db.py
```diff
@@ -121,6 +122,17 @@ def _build_importable_version(*, page_url, uuid=None, capture_time, uri,
     return version
 
 
+def _validate_timeout(timeout, default=None):
+    if timeout is None:
+        return default
+    elif timeout is not None and timeout < 0:
+        raise ValueError(f'Timeout must be non-negative. "{timeout}" was provided.')
```
Mr0grog, line 129: Since this is an `elif` branch, you don’t need to check `timeout is not None` here.

julianmclain: Good catch

Mr0grog (approved):
Looking back through this, I’m really not sure what we were waiting for to get finished, even though I remember there being some last thing that needed correction. ¯\\\_(ツ)\_/¯

I rebased this and added a few minor formatting tweaks, and one very minor functional change to correct a loophole I noticed in the constructor arguments.

PR #621, opened 2020-08-28
Set an import's media_type instead of metadata

Mr0grog:
We used to use `source_metadata` to set the media type and encoding information about a version, but web-monitoring-db has long since added top-level support for media type information. This updates the import code to use that, and also makes sure not to set `media_type` to an empty string, which is invalid (we used to do this, and imports would be rejected in some rare cases as a result).

Since we are processing a memento's `Content-Type` header in a more nuanced way now, this also means we might start storing extra media type parameters that we were leaving out before (we previously only captured the `charset` parameter).

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -338,13 +338,6 @@ class WaybackRecordsWorker(threading.Thread):
             if k.startswith(prefix)
         }
 
-        # Parse media type information.
-        media_type, *media_params = memento.headers.get('content-type', '').split(';')
-        # Clean up whitespace, remove empty parameters, etc.
-        stripped_params = (param.strip() for param in media_params)
-        media_params = [param for param in stripped_params if param]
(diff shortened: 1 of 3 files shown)


web_monitoring/cli/cli.py
```diff
@@ -338,9 +338,14 @@ def format_memento(self, memento, cdx_record, maintainers, tags):
             if k.startswith(prefix)
         }
 
+        # Parse media type information.
+        media_type, *media_params = memento.headers.get('content-type', '').split(';')
+        # Clean up whitespace, remove empty parameters, etc.
+        stripped_params = (param.strip() for param in media_params)
+        media_params = [param for param in stripped_params if param]
+        media_type_parameters = '; '.join(media_params)
```
Mr0grog, line 346: Future note: I think we may want to push this parsing down into the Wayback package. Then we might have something like:

```python
media_type = memento.media.type
media_type_parameters = str(memento.media.parameters)
```

PR #640, opened 2020-10-24
Replace diff-related modules with stubs

### LINKED ISSUE #638: Extract diffs and diff server into separate repo & Python package
Following on the heels of #206 and #477, I believe it’s time to create a separate repo and Python package for the diffs and the diff server (i.e. for `web_monitoring.diff`, `web_monitoring.diff_server`, and the `wm-diffing-server` command-line tool).

## Why?

This is a big change, but it’s important and worthwhile because:

- We’ve proven through time and through third-party usage that the diff tools here are reasonably abstract from EDGI’s Web Monitoring work and useful to others. (For example, the Internet Archive now runs the diff server as part of the Wayback Machine.)

- Having it mixed in with other tools here means the dependencies are less clear, and the architectural and release needs of different parts of this project sometimes collide and get in each others’ way.

- Outside users (read: the Internet Archive) have repeatedly asked for clearer release management and versioning, which has been tough to do well with all the cross-concerns in this repo (see above point).

- EDGI’s Web Monitoring team is moving towards releasing more *data* publicly, and I’d like to be able to tell a clearer story about about how to analyze that data, reproduce our diffs for analysis, etc. That’s much easier to do when there is a focused Python package people can install and use.

## The New Repo/Package

The new repo and package won’t be prefixed with “web-monitoring-” because one of the motivating factors here is that we are abstracting these tools from the Web Monitoring project. (This is also why the [*Wayback*](https://github.com/edgi-govdata-archiving/wayback) package is not “web-monitoring-wayback.”)

However, the new name is still up in the air. Possible candidates (please suggest others in the comments if you have thoughts):

- webdiffer
- web-diff
- webdiffs
- webdiff-suite
- diffweb

**I would really love some suggestions here; “webdiff” is an existing project on PyPI, and I’m somewhat worried all these names are too close to that.** (“Web” is in most of these to try and describe a little more clearly the area our diff algorithms are focused on, but also that they are more broad than just HTML. There’s probably a better choice out there that @danielballan and I haven’t thought of yet.)

Other important features of the new project:

- The server should be an optional addition. It’s in the codebase, but you’d run:

    ```sh
    $ pip install webdiffer[server]
    ```

    To install it instead of:

    ```sh
    $ pip install webdiffer
    ```

- Some of the experimental diffs that we don’t actively use should also be optional dependencies. Specifically: https://github.com/anastasia/htmldiffer and https://github.com/danielballan/htmltreediff. They haven’t been especially active projects, and I think there are some dependency conflicts.

- Should handle docs like the [Wayback](https://github.com/edgi-govdata-archiving/wayback) package does rather than how they are handled here.

- Should be released as a package on PyPI, with clear versioning.

- For now, this project should probably depend on the new one and re-export it, so things continue to work while external users update. After the Internet Archive and EDGI’s deployments have all migrated, we’ll remove it from here.

## Timeframe

I’m looking to do this next week! I know that’s soon, but I’d like to do this before the Web Monitoring team does a public data release, and hopefully have time to iron out any issues.

## To-Do

- [x] Create new repo
- [x] Filter history of this repo to just the relevant bits
- [x] Rename package and internal references
- [x] Upload to new repo
- [x] Separate diff & server requirements
- [x] Rewrite README
- [x] Write minimal docs
- [x] Set up readthedocs.org
- [x] Get tests running in Circle
- [x] Get docker builds running/publishing in Circle
- [x] Move all relevant issues to the new repo
- [x] Publish v0.1.0 to PyPI
- [ ] Deprecate diffs and diff server in this package

Mr0grog:
**Work-in-progress! This still needs [edgi-govdata-archiving/web-monitoring-diff][diff-repo] to be ready before merging.**

All the diff and diff_server related code is being moved out into a new `web_monitoring_diff` package ([edgi-govdata-archiving/web-monitoring-diff][diff-repo]). This removes the parts that are completely removable from here, and replaces other parts with stubs that just import the relevant implementations from web_monitoring_diff (so other code that uses this repo can keep working for a while before transitioning).

Fixes #638.

[diff-repo]: https://github.com/edgi-govdata-archiving/web-monitoring-diff


Mr0grog:
I’m closing this in favor of #642. It makes more sense to leave the current implementations intact and deprecated here, rather than replace them with pointers to the new package (because that prevents us from making many kinds of changes to the new package).

PR #652, opened 2020-11-01
Fix unplaybackable cache and retries

### LINKED ISSUE #582: Provide clearer messaging when zero records imported from IA
When using `wm import ia`, it's hard to differentiate when 0 records were imported vs when records were actually imported.

To be able to tell, the user has to catch that there was no import job ID when there are 0 records.

We should have a different message for this case and a non-zero exit code.

Mr0grog:
I recently noticed that the unplaybackable cache didn't seem to be used, and I'm not sure when it broke. While doing that, I also noticed that the secondary retry functionality was also broken (!) so I fixed that too.

Finally, I went ahead and added better logging for the case where we didn't actually import anything. Fixes #582.

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -173,6 +173,9 @@ def _add_and_monitor(versions, create_pages=True, skip_unchanged_versions=True, 
     versions = _get_progress_meter(versions)
     import_ids = cli.add_versions(versions, create_pages=create_pages,
                                   skip_unchanged_versions=skip_unchanged_versions)
+    if len(import_ids) == 0:
+        return
+
     print('Import job IDs: {}'.format(import_ids))
     print('Polling web-monitoring-db until import jobs are finished...')
     errors = cli.monitor_import_statuses(import_ids, stop_event)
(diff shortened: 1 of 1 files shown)


Mr0grog:
Part of this is that I was thinking about whether the unplaybackable cache should be extended to cover successful loads, too, since we now have a *huge* amount of overlap in the timeframes of our normal runs. That’s done to account for Wayback’s frequent indexing issues, but means we wind up pulling down the same mementos a lot, and I’m not sure there’s any reason no to avoid doing so by sticking them in the cache. That’s out of scope for this PR, though.

web_monitoring/cli/cli.py
```diff
@@ -605,10 +613,9 @@ def import_ia_urls(urls, *, from_date=None, to_date=None,
             maintainers,
             tags,
             stop_event,
-            unplaybackable,
+            unplaybackable=unplaybackable,
             tries=(None,
-                   dict(retries=3, backoff=4, timeout=(30.5, 2)),
-                   dict(retries=7, backoff=4, timeout=60.5))))
+                   dict(retries=3, backoff=4, timeout=(30.5, 2)))))
         memento_thread.start()
```
Mr0grog, line 619: I simplified this since the second and third retries haven’t actually been running for a while anyhow.

PR #680, opened 2020-12-18
Show every CDX record in import progress meter

Mr0grog:
One long-standing issue with the import script is that, because of the overly-complex retry queue logic, it only shows progress in terms of new versions loaded, and does not include skipped versions and errors in the count, leading to a lack of clarity in what's going on. The real question the progress bar should be answering is "how many CDX records have been processed?" Now that we've stopped using the retry queue, this is a start towards doing that.

The code here is *super* messy. There's a lot of crazy logic to unwind, and this script can be futzy to work with, so cleaning up the logic may take a little while. But this first commit *does* work, even if it's pretty ugly.


Mr0grog:
(This is driven by me trying to debug a lot of issues with the Wayback import script this week. Not being clear on whether the script is stuck because of something with Wayback servers or because it's just chewing on a block of errors or skips is a pain point.)

Mr0grog:
This needs more cleanup, but it also needs to get in so it’s not blocking other needed import fixes. :\

PR #686, opened 2021-01-06
Drop Travis; switch to RTD to docs publishing.

danielballan:
We currently use Travis-CI for docs publishing _only_. This PR uses RTD instead, and removes the Travis-CI configuration and associated encrypted deploy key. This change is motivated by pricing changes in Travis-CI.

<hr />

@Mr0grog When we last chatted about this I had in mind moving to GitHub Actions (which is where I have migrated work projects). But since (1) we were only using Travis-CI to build docs and (2) we already use RTD for the `wayback` repository, this seemed like a better approach here.


danielballan:
Power-cycled in an attempt to trigger the RTD webhoook

Mr0grog (approved):
Sorry I didn’t get to this yesterday! Looks reasonable to me. I’m going to go ahead and merge it in now, we’ll see if it actually runs correctly.

As a side note, can you add me as a maintainer to the RTD project?

PR #688, opened 2021-01-08
Fix up Read The Docs build

Mr0grog:
I certainly screwed up reviewing #686! This fixes a syntax error with a comment in the RTD requirements file, and also tweaks the language slightly just because.


Mr0grog:
RTD does not seem to be picking this up, so I going to merge and see if that does it.

PR #763, opened 2022-02-01
Upgrade to new CircleCI convenience image

Mr0grog:
The old images have been deprecated, so this really needs to get updated. Nothing seems to have *broken* yet, though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,9 +17,11 @@ jobs:
           command: |
             echo 'Creating virtualenv...'
             echo `which python`
-            python -m venv venv
+            # python -m venv venv
+            pyenv virtualenv venv
             echo 'Created virtualenv'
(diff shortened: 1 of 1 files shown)


Mr0grog:
`venv` seems to have some issues on the new `cimg/python` images, as does `pyenv-virtualenv` (it’s installed, but doesn’t seem to be configured correctly). There’s a discussion about this on their forum: https://discuss.circleci.com/t/upgrading-to-the-convenience-image-breaks-everything/42746

So this may just be a non-starter for now. We can leave as-is until it breaks, or switch to GitHub Actions (which I’ve used quite a bit on other projects at this point and is great).

Mr0grog:
D'oh, this was just a matter of the cache having a virtualenv that was not compatible with these new images, which really shouldn’t have been that surprising. 🤦

PR #819, opened 2022-10-05
Handle PDFs with no metadata

Mr0grog:
We were previously raising exceptions for PDFs with no metadata at all (I think `.metadata` may have previously always been an object even if empty, but not sure... in any case, it no longer is).


Mr0grog:
A decent real-world use case for this is:

```sh
scripts/wm import ia 'https://arpa-e.energy.gov/sites/default/files/nepa/Updated-CX%20Determination-PARC.pdf' --from 2022-07-24 --to 2022-07-25 --dry-run
```

PR #854, opened 2025-01-23
Update runtime dependencies

Mr0grog:
Reviving this project a bit since we need web monitoring infra active again. This updates just the runtime dependencies to their latest. It does not touch dev/test/docs dependencies.

Other future notes:
- I have been enjoying uv; we may want to start using that.
- We are locked into Python 3.10.x for now by cchardet. web-monitoring-diff also has this issue, and fixing it is a *huge* project (that I need to do) since there are no real good solutions (the tradeoffs here are surprisingly complicated; I did a lot of benchmarking and testing a year ago that I should have written up and need to return to).


Mr0grog:
Well, I wound up having to update some (but not all) dev deps anyway. Also had to specify which urllib3 we are using because VCR is not cross-compatible (and then had to regenerate our VCR cassettes). An alternative is adding a hack like we use in `wayback`, but that doesn’t seem worthwhile for this project: https://github.com/edgi-govdata-archiving/wayback/blob/47a436c27b83b3cb40bf18afcc12d3f3bc7b14a3/src/wayback/tests/support.py

Mr0grog:
…and now that means dropping Python 3.9 support, since requiring urllib3 v2 makes for conflicting requirements on Pythons older than 3.10. sigh.

PR #855, opened 2025-01-23
Add `--archive-s3 <bucket>` option

### LINKED ISSUE #663: Import script should upload bodies directly to S3
The Internet Archive import script(s) (`wm import ia` and `wm import ia-known-pages`) should have an option that causes them to upload Mementos to S3:

```sh
$ wm import ia 'http://www.epa.gov/' --from <time> --s3 <bucket_name_or_s3://_uri>
```

S3 credentials should be read from the standard AWS environment variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).

Setting this should cause the importer to save memento bodies in S3 before sending import metadata to web-monitoring-db. The metadata's `uri` property should be rewritten to point to the uploaded location in S3. The objects in S3 should be named by their SHA-256 base-16-encoded hash and their `Content-Type` header should be set appropriately, as web-monitoring-db currently does: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/46561ae6eb52b0d923f7832100d161fc98667d0c/lib/archiver/archiver.rb#L36-L66

The value of the `--s3` should be one of:
- A bucket name, e.g. `edgi-wm-archive`
- A bucket name and path prefix, e.g. `edgi-wm-archive/wayback/`
- An S3 URI, e.g. `s3://edgi-wm-archive`
- An S3 URI with a path prefix, e.g. `s3://edgi-wm-archive/wayback/`

## Why?

When we use the scripts here to import data from the Wayback Machine, we process the mementos from Wayback and then send the *metadata* to the DB’s import API. The API doesn’t actually accept the raw data of the response bodies (it’s complicated to do in a safe and effective way, and, although there is [a placeholder for it in the code](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/46561ae6eb52b0d923f7832100d161fc98667d0c/app/jobs/import_versions_job.rb#L99-L102), we never made it happen).

The flow is something like this:

```
                    ┌─────────────────────┐
                    │                     │
                ┌──▶│   Wayback Machine   │◀─────────────────┐
                │   │                     │                  │
                │   └─────────────────────┘                  │
                │           ▲                                │
                └─────┐     │                                │
 ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─│─ ┐  │       ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐  │
   Import Script      │     │         Web-Monitoring-DB      │
 │                    │  │  │       │                     │  │
     ┌──────────────┐ │     │          ┌───────────────┐     │
 │   │              │ │  │  │       │  │    Process    │  │  │
     │  Search CDX  │─┘     │   ╔═════▶│   Metadata    │     │
 │   │              │    │  │   ║   │  │               │  │  │
     └──────────────┘       │   ║      └───────────────┘     │
 │           ║           │  │   ║   │          ║          │  │
             ▼              │   ║              ▼             │
 │   ┌───────────────┐   │  │   ║   │  ┌───────────────┐  │  │
     │               │      │   ║      │ Load Memento  │     │
 │   │ Load Mementos │───┼──┘   ║   │  │     Data      │──┼──┘          ┌───────────────┐
     │               │          ║      │               │       Metadata │               │
 │   └───────────────┘   │      ║   │  └───────────────┘  │    ╔═══════▶│  PostgreSQL   │
             ║                  ║              ║               ║        │               │
 │           ▼           │      ║   │          ▼          │    ║        └───────────────┘
     ┌───────────────┐          ║      ┌───────────────┐       ║
 │   │   Format as   │   │      ║   │  │               │  │    ║        ┌───────────────┐
     │  Importable   │══════════╝      │     Save      │═══════╣        │               │
 │   │   Metadata    │   │          │  │               │  │    ╚═══════▶│      S3       │
     └───────────────┘                 └───────────────┘       Memento  │               │
 └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘          └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘      Data   └───────────────┘
```

You can see here that *both* the import script and and web-monitoring-db have to get data from Wayback. The problem that happens here is when the Wayback Machine loads slowly (either because it’s under heavy load or we’re getting a memento that is rarely accessed), it might fail to load on the web-monitoring-db side, and therefore that record fails to save. This typically happens in about 1-2 of every 2000 mementos. Besides the occasional failure, this double-loading is also a waste of bandwidth and resources for both us and the archive!

*(Note: The failures aren’t a serious problem because we typically grab overlapping sets of data from the Wayback Machine each time we run the script. The likelihood of a memento failing this way across multiple imports is pretty low. We do the overlap to work around the fact that Wayback has frequent indexing issues that sometimes cause mementos to be unfindable until several days after they were archived. The overlap period is longer than such outages typically last.)*

We can work around this by doing what the old Versionista import script used to do: upload to S3 ourselves, before sending the metadata to web-monitoring-db. Web-monitoring-db will automatically skip loading mementos if the location it’s given is in an S3 bucket it already knows is OK.

Basically, we want the workflow to be more like:

```
                    ┌─────────────────────┐
                    │                     │
                ┌──▶│   Wayback Machine   │
                │   │                     │
                │   └─────────────────────┘
                │           ▲
                └─────┐     │
 ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─│─ ┐  │       ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
  Import Script       │     │        Web-Monitoring-DB
 │                    │  │  │       │                     │
     ┌──────────────┐ │     │          ┌───────────────┐
 │   │              │ │  │  │       │  │    Process    │  │
     │  Search CDX  │─┘     │   ╔═════▶│   Metadata    │
 │   │              │    │  │   ║   │  │               │  │
     └──────────────┘       │   ║      └───────────────┘
 │           ║           │  │   ║   │          ║          │
             ▼              │   ║              ║
 │   ┌───────────────┐   │  │   ║   │          ║          │
     │               │      │   ║              ║
 │   │ Load Mementos │───┼──┘   ║   │          ║          │
     │               │══════╗   ║              ║
 │   └───────────────┘   │  ║   ║   │          ║          │
             ║              ║   ║              ║
 │           ▼           │  ║   ║   │          ▼          │
     ┌───────────────┐      ║   ║      ┌───────────────┐
 │   │   Format as   │   │  ║   ║   │  │               │  │
     │  Importable   │══════╬═══╝      │     Save      │════╗
 │   │   Metadata    │   │  ║       │  │               │  │ ║
     └───────────────┘      ║          └───────────────┘    ║
 └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘  ║       └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘ ║
                            ║                               ║
                            ▼                               ║
                    ┌───────────────┐  ┌───────────────┐    ║
                    │               │  │               │    ║
                    │      S3       │  │  PostgreSQL   │◀═══╝
                    │               │  │               │
                    └───────────────┘  └───────────────┘
```

Mr0grog:
This is pretty ugly, but it's a working first cut at the problem and fixes #663. This is suddenly more important since IA seems to be under a lot of load with peopole checking up on Trump-related website changes, and I'm seeing a lot more import-related failures on things that work fine on the processing side and fail to re-download in the DB importer. This will short-circuit that issue.

This ideally would go in a separate worker component, but we just don’t have the right data available from outside `WaybackRecordsWorker` right now. This would also ideally be tested more, but 🤷

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -286,6 +286,11 @@ class RequestStatistics:
 # Ideally this is incorporated into our shared HTTP adapter/session.
 MEMENTO_STATISTICS = RequestStatistics(leaderboard_size=10, leaderboard_min=10)
 
+# TODO: wrap this up in something nicer so we don't need to manage the lock
+# in the middle of business logic.
+STORED_CONTENT_HASHES = set()
+STORED_CONTENT_HASHES_LOCK = threading.Lock()
+
 
 class WaybackRecordsWorker(threading.Thread):
     """
@@ -382,10 +387,18 @@ class WaybackRecordsWorker(threading.Thread):
             version = self.format_memento(memento, record, self.maintainers,
                                           self.tags)
             if self.archive and version['version_hash']:
-                path = self.archive / version['version_hash']
-                if not path.exists():
-                    logger.info(f'Uploading to S3: {record.raw_url}')
+                hash = version['version_hash']
+                upload = True
+                with STORED_CONTENT_HASHES_LOCK:
+                    upload = hash not in STORED_CONTENT_HASHES
+                    if upload:
+                        STORED_CONTENT_HASHES.add(hash)
+
+                path = self.archive / hash
+                if upload and not path.exists():
+                    logger.info(f'Uploading to S3 (hash={hash}, url="{record.raw_url}")')
                     path.write_bytes(memento.content)
+
                 version['uri'] = path.as_url()
 
             return version
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -320,14 +320,7 @@ class WaybackRecordsWorker(threading.Thread):
                                        user_agent=USER_AGENT,
                                        **session_options)
         self.wayback = wayback.WaybackClient(session=session)
-        self.archive = None
-        if archive_bucket:
-            self.archive = S3Path(f's3://{archive_bucket}', client=S3Client(extra_args={
-                'ACL': 'public-read',
-                # Ideally, we'd gzip stuff, but the DB needs to learn to correctly
-                # read gzipped items first.
-                # 'ContentEncoding': 'gzip'
-            }))
+        self.archive_bucket = archive_bucket
 
     def is_active(self):
         return not self.cancel.is_set()
@@ -386,7 +379,7 @@ class WaybackRecordsWorker(threading.Thread):
         with memento:
             version = self.format_memento(memento, record, self.maintainers,
                                           self.tags)
-            if self.archive and version['version_hash']:
+            if self.archive_bucket and version['version_hash']:
                 hash = version['version_hash']
                 upload = True
                 with STORED_CONTENT_HASHES_LOCK:
@@ -394,7 +387,14 @@ class WaybackRecordsWorker(threading.Thread):
                     if upload:
                         STORED_CONTENT_HASHES.add(hash)
 
-                path = self.archive / hash
+                archive = S3Path(f's3://{self.archive_bucket}', client=S3Client(extra_args={
+                    'ACL': 'public-read',
+                    'ContentType': version['media_type'] or 'application/octet-stream',
+                    # Ideally, we'd gzip stuff, but the DB needs to learn to correctly
+                    # read gzipped items first.
+                    # 'ContentEncoding': 'gzip'
+                }))
+                path = archive / hash
                 if upload and not path.exists():
                     logger.info(f'Uploading to S3 (hash={hash}, url="{record.raw_url}")')
                     path.write_bytes(memento.content)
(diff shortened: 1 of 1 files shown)


Mr0grog:
Turns out this is uploading data with an invalid `binary/octet-stream` media type. We should use the `Content-Type` header from the memento or, if not present, the sniffed media type (or we should just use the sniffed type?).

A quick look at Cloudpathlib suggests this might be hard. We could configure gzip encoding and a standard ACL for *all* uploads, but we’d have to construct a custom client for each upload now (I don’t *think* this is high overhead, but… I dunno?). Maybe need to drop down to boto3 and drop cloudpathlib (for this use case at least).

Also while I’m here, we should keep a list of seen hashes in memory so we don’t waste effort checking S3 if we encounter something a second time.

PR #858, opened 2025-02-04
Warcs have useful content too

Mr0grog:
We now have some of our own crawls, and need to import data from the resulting WARCs. This scripts runs through a WARC file and uploads the bodies to our hash storage and the creates an import job with the resulting "versions".

This is a little biased towards WARCs output by Browsertrix.crawler, because that's what we're using right now. It will probably need some adapting to work with other WARCs, if and when we encounter them (hopefully from IA soon!).

While I was at it, I updated the import code to use the “new” import format (from several years ago!) that we never updated to before. It makes things much clearer.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ python-dateutil ~=2.9.0
 docopt ~=0.6.2
 lxml ~=5.3.0
 pypdf[crypto] ~=5.1.0
+PyYAML ~=6.0.2
 sentry-sdk ~=1.14.0
 requests ~=2.32.3
 # We need to specify urllib3 to make sure everyone is using the same version;
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,5 @@ requests ~=2.32.3
 urllib3 ~=2.3.0
 toolz ~=1.0.0
 tqdm ~=4.67.1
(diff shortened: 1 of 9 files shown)


Mr0grog:
Some pages don’t have `<title>` elements in their raw HTML, but do have them added later by scripts. Browsertrix successfully grabs these titles in its `pages.jsonl` file, so we should probably use that when available. For example, here’s a line from that file where the title is not in the raw HTML:

```js
{"id":"154802d9-b976-4a26-a038-fe9f38df0fa0","url":"https://noaa.maps.arcgis.com/home/item.html?id=f66c1e33f91d480db7d1b1c1336223c3","title":"NMFS ESA Critical Habitat gdb - Overview","loadState":4,"ts":"2025-02-28T03:38:33.463Z","status":304,"seed":true,"depth":0,"favIconUrl":"https://noaa.maps.arcgis.com/favicon.ico"}
```

Unfortunately there doesn’t seem to be anything for this directly in the WARC file.

Mr0grog:
This has been sitting separate for too long. Need to clean up stuff that’s duplicative with the IA import tool and land this. I think the title stuff mentioned above is skippable for now. Otherwise it mostly seems to be working well with current Browsertrix data.

PR #860, opened 2025-02-14
Add `--from` and `--sample` opts to healthcheck

Mr0grog:
I frequently find myself making little tweaks to the constants in the healthcheck script when running it. This adds proper CLI options to let you specify a different sample size and timeframe so you don't have to change the code.


Mr0grog:
Reminder to myself: make sure to push a new release with this, and update the [nightly import script](https://github.com/edgi-govdata-archiving/web-monitoring-ops/blob/717536b0fef85734177c3317f4ef94f15de28621/kubernetes/production/ia-import-job.yaml#L26) to use the new, nicer time syntax (so we can say `--from 7d` instead of `--from 168  # 1 week ago`.

PR #861, opened 2025-02-17
Update import options for versions

Mr0grog:
Many years ago, we made a bunch of changes to the versions table in the DB and the corresponding format of import records in order to make them clearer and more definitely resemble a record of an HTTP response (since that is effectively what they wound up being). We never updated the corresponding import code here, though! That means we've had backwards-compatibility code sitting around in the DB import script for years. This finally updates things, and after shipping, we can also clean up the DB.

This is extracted from #858, which I thought would land sooner.

### DIFF
--- a/web_monitoring/cli/cli.py
+++ b/web_monitoring/cli/cli.py
@@ -413,13 +413,13 @@ class WaybackRecordsWorker(threading.Thread):
         with memento:
             version = self.format_memento(memento, record, self.maintainers,
                                           self.tags)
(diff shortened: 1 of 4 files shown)


Mr0grog:
Need to double-check this running locally, but either web.archive.org is down right now or I am blocked for too many requests. Dunno, gonna give it an hour or so.

Mr0grog:
Archive.org is back operating well today, and this does work fine.

PR #870, opened 2025-09-30
Handle PDF parsing errors more comprehensively

Mr0grog:
We currently handle `PyPdfError` exceptions when parsing PDFs. However, I ran into a malformed PDF today that totally broke our crawl handling because it raises a `ValueError`. Even though this is handled better in newer versions of pypdf (which I’ve also upgraded to here), it still seems like this kind of thing is always reasonably possible, especially if we happen to get some garbled data. This addresses the issue by handle all `Exception` subclasses (but we log a full exception trace if it’s not a `PyPdfError` since we still don't really *expect* those).


Mr0grog:
NOTE TO SELF: need to remember to upgrade the crawler's dependency on this package after merging.

PR #913, opened 2026-02-27
Fix redundant IA import errors in Sentry

### LINKED ISSUE #899: Improve error tracking for IA import CDX failures
After upgrading to Sentry SDK v2 and Internet Archive starting to make progress on its indexing backlog, we are seeing lots of redundant errors from IA imports in Sentry:

Image

All these “Error getting CDX data for {url}” errors should probably be grouped.

It looks like the issue is that Sentry is now picking up on our logger here: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/127e3c679282a48584d938185ecf413a1a5a80fa/web_monitoring/cli/cli.py#L795-L815

We should probably have it ignore the main CLI logger, and add explicit `sentry.exception()` calls at those error handling sites.

kevo-1:
### Summary 

This PR resolves the issue where Sentry was incorrectly grouping these errors together by picking up on the main CLI module logger.

### Changes

- Ignored the main  `cli.py` `logger (__name__)` during Sentry SDK initialization so that these logged errors aren't automatically captured generically.

- Added explicit `sentry_sdk.capture_exception(error)` calls into the appropriate `except blocks`  to ensure errors are still tracked but with their proper, distinct context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ This project wouldn’t exist without a lot of amazing people’s help. Thanks t
 | [💻](# "Code") [⚠️](# "Tests") [🚇](# "Infrastructure") [📖](# "Documentation") [💬](# "Answering Questions") [👀](# "Reviewer") | [Dan Allan](https://github.com/danielballan) |
 | [💻](# "Code") | [Vangelis Banos](https://github.com/vbanos) |
 | [💻](# "Code") [📖](# "Documentation") | [Chaitanya Prakash Bapat](https://github.com/ChaiBapchya) |
+| [💻](# "Code") | [Kevin (Hany) Bastawrous](https://github.com/kevo-1) |
 | [💻](# "Code") [⚠️](# "Tests") [🚇](# "Infrastructure") [📖](# "Documentation") [💬](# "Answering Questions") [👀](# "Reviewer") | [Rob Brackett](https://github.com/Mr0grog) |
 | [💻](# "Code") | [Stephen Buckley](https://github.com/StephenAlanBuckley) |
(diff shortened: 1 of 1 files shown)


web_monitoring/cli/cli.py
```diff
@@ -331,6 +331,7 @@ def handle_record(self, record):
                 self.unplaybackable[record.raw_url] = datetime.utcnow()
             self.results_queue.put([record, None, error])
         except Exception as error:
+            sentry_sdk.capture_exception(error)
             self.results_queue.put([record, None, error])
```
Mr0grog: You should not need to capture here, since the error is getting put on the results queue, which then gets handled by `_filter_and_summarize_mementos()` below.

web_monitoring/cli/cli.py
```diff
@@ -988,7 +994,18 @@ def validate_db_credentials():
...
+    sentry_logging = LoggingIntegration(
+        level=logging.INFO,        # Capture info and above as breadcrumbs
+        event_level=logging.ERROR  # Send errors as events
+    )
+    sentry_sdk.init(
+        integrations=[sentry_logging]
+    )
+    # Ignore the main logger so we can manually send exceptions with URLs grouped
+    from sentry_sdk.integrations.logging import ignore_logger
+    ignore_logger(__name__)
```
Mr0grog: I think all this `LoggingIntegration` configuration is going a little overboard, and as far as I can tell here, it’s also just reproducing the default settings. I think this can and should just be pared back to calling `ignore_logger(__name__)` after initializing with no special options (i.e. same `init()` call as before).

Mr0grog (changes requested):

